PR #28, opened 2016-09-05
Fix example in README

GeertJohan:
Percentages in a string should be either escaped `"%%"` or placed in a raw string. ``` `%` ```


aarondl:
Is this relating to PostgreSQL? Because in Go it doesn't matter which quotes you use outside of a printf statement for the % character. Will merge anyway because it seems to be tricking out the syntax highlighter in the README. Thanks :)

GeertJohan:
Ah sorry, yes I made a thinking error. Github showed strange syntax but the percentage is indeed Printf only.

PR #37, opened 2016-09-16
Add SQL snippet for MySQL

nguyendangminh:


nullbio:
Thanks for the PR. This is not something we're interested in adding, due to it being so similar to the Postgres version, we'd rather leave this out to reduce clutter.

PR #43, opened 2016-09-21
use supplied conflictColumns in Upsert

ryanbarry:
using upsert today and found it generated an empty conflict target when given a non-empty `conflictColumns` argument. i'm not sure if there should be some additional work done on the passed in value before just sticking it in the query builder... but this is what i did to get it working for me now.


nullbio:
I think you're right Ryan. Good catch. We're just making the final adjustments for the release now, so should be able to get this out in an hour or so hopefully. Will merge this as well after I take a closer look at it. Thanks.

aarondl:
Hey @ryanbarry, thanks for the code. We've implemented this fix in an ever so slightly different way that avoids the else statement. Thanks for finding this bug!

PR #48, opened 2016-09-24
Run gofmt on generated output files

nubunto:
This removes the necessity of having the user running gofmt on generated
output every time after making a database change.
This does not break existing tests.

### DIFF
--- a/sqlboiler.go
+++ b/sqlboiler.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"go/build"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"regexp"
 	"strings"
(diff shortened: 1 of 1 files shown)


aarondl:
In the past we've run gofmt without shelling out to the binary and I wanted to keep it that way, but we had only done it on individual templates. I've refactored the output.go file significantly (simplified a lot of stuff). And now run the gofmt piece on the final output as well. Should have also improved performance. Closing this in favor of that :) 9e4b5b750c6966f2d70bfede73c62e8890abfdcd

Currently this is in the dev branch which is a little unstable - CI is failing because one to one relationships are in flux. We'll be fixing this tomorrow and hopefully releasing 2.0.2 to solve the two open issues, as well as the things you've submitted.

Thanks for the PR regardless! Even if we didn't accept it this time, the code got better because of this.

PR #49, opened 2016-09-26
Fix cross schema constraints issue

### LINKED ISSUE #46: Generating models with self-reference fails
Possibly related to #45, but I see a different behavior, so I wanted to file a separate ticket.

Here is the SQL this time:

```
CREATE TABLE industries (
    id integer NOT NULL,
    name text,
    parent_id integer
);

ALTER TABLE ONLY industries
    ADD CONSTRAINT industries_pkey PRIMARY KEY (id);


ALTER TABLE ONLY industries
    ADD CONSTRAINT industries_parent_id_fkey FOREIGN KEY (parent_id) REFERENCES industries(id);
```

So, parent_id references another industry.

The models build works, but generates invalid code:

```
type Industry struct {
	ID       int         `boil:"id" json:"id" toml:"id" yaml:"id"`
	Name     null.String `boil:"name" json:"name,omitempty" toml:"name" yaml:"name,omitempty"`
	ParentID null.Int    `boil:"parent_id" json:"parent_id,omitempty" toml:"parent_id" yaml:"parent_id,omitempty"`

	R *industryR `boil:"-" json:"-" toml:"-" yaml:"-"`
	L industryL  `boil:"-" json:"-" toml:"-" yaml:"-"`
}

// industryR is where relationships are stored.
type industryR struct {
	Parent           *Industry
	Parent           *Industry
	Parent           *Industry
	Parent           *Industry
	ParentIndustries IndustrySlice
	ParentIndustries IndustrySlice
	ParentIndustries IndustrySlice
	ParentIndustries IndustrySlice
}
...
```

It generates 4 copies of Parent, and 4 copies of ParentIndustries.  There should only be a single instance of each.  It also appears that the code for Adding/Replacing on the relationship does not compile.

jseriff:
Query was not properly joining on both name and schema.  This will fix the issue where a constraint exists across multiple schemas and causes duplicate relationships to be created.

Fixes #46

### DIFF
--- a/bdb/drivers/postgres.go
+++ b/bdb/drivers/postgres.go
@@ -242,8 +242,8 @@ func (p *PostgresDriver) ForeignKeyInfo(schema, tableName string) ([]bdb.Foreign
 		ccu.table_name as dest_table,
 		ccu.column_name as dest_column
 	from information_schema.table_constraints as tc
-		inner join information_schema.key_column_usage as kcu ON tc.constraint_name = kcu.constraint_name
-		inner join information_schema.constraint_column_usage as ccu ON tc.constraint_name = ccu.constraint_name
+		inner join information_schema.key_column_usage as kcu ON tc.constraint_name = kcu.constraint_name and tc.constraint_schema = kcu.constraint_schema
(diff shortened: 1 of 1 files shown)


aarondl:
Merged into dev, will be creating a release based on multiple bug fixes as usual. Thanks for the code. Starting to look at other issues now.

PR #56, opened 2016-10-11
mysql driver: query only base tables

SlinSo:
hi @aarondl,

as long as it is not decided if views should be supported we have to ignore those when generating the model. Blacklisting would have worked, but I think this query fix is more appropriate.


aarondl:
https://github.com/vattle/sqlboiler/releases/

Released :)

PR #57, opened 2016-10-18
Add new line after warning message

yogipatel:
When multiple incompatible data type warnings are emitted, it's hard to read the type name since there's no white space between the errors. This adds a new line to the end of each warning.


nullbio:
Hi @yogipatel -- I added this commit already when I received no response from you on your fork. It should be good to go now. Just pull in the latest version.

PR #65, opened 2016-11-12
Fix reading basedir from config and cli flag

### LINKED ISSUE #64: BaseDir option not respected
Neither setting it in the config nor passing in as cli flag works.

pritambaral:
Fixes #64


aarondl:
Merged here. I rebased your commit on to dev so we could properly release/tag.
https://github.com/vattle/sqlboiler/commit/d2eccc98adfeb23868f21d677963e7a2d3e095ae

Thank you for your contribution :)

PR #96, opened 2017-01-24
Add option to change output file extension

bjyoungblood:
Within a package, I like to clearly separate generated code from non-generated code. Being able to customize the file extension is an easy way to draw a line between the two.

For example, `users.gen.go` would contain the generated user model, then `users.go` could contain additional methods (`VerifyPassword`, for example).

This would also make it easier to target generated files in scripts (such as a CI script to flag changes to generated files).

I set the default extension to `.go` in order to maintain backwards compatibility.

### DIFF
--- a/boilingcore/boilingcore.go
+++ b/boilingcore/boilingcore.go
@@ -47,6 +47,10 @@ type State struct {
 
 // New creates a new state based off of the config
 func New(config *Config) (*State, error) {
+	if config.FileExtension == "" {
+		config.FileExtension = ".go"
+	}
+
 	s := &State{
 		Config: config,
 	}
--- a/README.md
+++ b/README.md
@@ -292,6 +292,7 @@ Flags:
       --basedir string          The base directory has the templates and templates_test folders
   -b, --blacklist stringSlice   Do not include these tables in your generated package
   -d, --debug                   Debug mode prints stack traces on error
+  -e, --extension string        The extension to use for output files (default ".go")
       --no-auto-timestamps      Disable automatic timestamps for created_at/updated_at
       --no-hooks                Disable hooks feature for your models
       --no-tests                Disable generated go test files
--- a/boilingcore/config.go
+++ b/boilingcore/config.go
@@ -15,6 +15,7 @@ type Config struct {
 	NoTests          bool
 	NoHooks          bool
 	NoAutoTimestamps bool
+	FileExtension    string
 
 	Postgres PostgresConfig
 	MySQL    MySQLConfig
--- a/boilingcore/output.go
+++ b/boilingcore/output.go
@@ -34,7 +34,7 @@ func generateOutput(state *State, data *templateData) error {
 		templates:            state.Templates,
 		importSet:            defaultTemplateImports,
 		combineImportsOnType: true,
-		fileSuffix:           ".go",
+		fileSuffix:           state.Config.FileExtension,
 	})
 }
 
@@ -46,7 +46,7 @@ func generateTestOutput(state *State, data *templateData) error {
 		templates:            state.TestTemplates,
 		importSet:            defaultTestTemplateImports,
 		combineImportsOnType: false,
-		fileSuffix:           "_test.go",
+		fileSuffix:           fmt.Sprintf("_test%s", state.Config.FileExtension),
 	})
 }
 
@@ -58,7 +58,7 @@ func generateSingletonOutput(state *State, data *templateData) error {
 		data:           data,
 		templates:      state.SingletonTemplates,
 		importNamedSet: defaultSingletonTemplateImports,
-		fileSuffix:     ".go",
+		fileSuffix:     state.Config.FileExtension,
 	})
 }
 
@@ -70,7 +70,7 @@ func generateSingletonTestOutput(state *State, data *templateData) error {
 		data:           data,
 		templates:      state.SingletonTestTemplates,
 		importNamedSet: defaultSingletonTestTemplateImports,
-		fileSuffix:     ".go",
+		fileSuffix:     state.Config.FileExtension,
 	})
 }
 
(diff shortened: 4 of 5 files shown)


boilingcore/boilingcore.go
```diff
@@ -47,6 +47,10 @@ type State struct {
 
 // New creates a new state based off of the config
 func New(config *Config) (*State, error) {
+	if config.FileExtension == "" {
```
bjyoungblood: Note: this preserves backwards compatibility for anyone who is importing this package as a library instead of using the CLI directly.

nullbio:
Hi @bjyoungblood. Thanks for the PR.  In regards to your use case, I don't think it's a good idea to be storing your own custom code inside the generated package. Anything custom should go in its own seperate package. A couple of reasons for this:

* We've tried to export everything that you may need to use, so there's no reason that I can see to include it in the same package. Correct me if I'm wrong though.
* If we ever make changes to SQLBoiler in the future, and choose a filename you've chosen in your models folder as a custom file, it will overwrite it. 
* If we decide to rm the models folder for some reason in the future (perhaps we're introducing new files or filename changes and we need to do cleanup), it will delete your custom code.

The bottom two aren't explicitly problems now, but it may lead to unexpected behavior in the future depending on the direction SQLBoiler takes, and so I think it would be promoting an ill advised practice.

As for the CI flagging thing, can you not flag a specific folder?

We also give you the ability to choose your own output folder name and package name in case you'd like to use "models" for something else (perhaps the stuff you're referring to right now).

@aarondl What are your thoughts?

bjyoungblood:
Your reasoning makes sense. I also go back and forth on whether it's a good idea to include custom and generated code in the same package, but there are some situations where I think it can make sense.

The primary use case I had in mind was to be able to attach extra methods to the generated models, primarily to keep things more organized.

Being able to have
`func (u User) VerifyPassword(verify string) (ok bool, err error)`
seems cleaner to me than to have a separate package with:
`func VerifyUserPassword(user models.User, verify string) (ok bool, err error)`

There might be a better way than either of those though.

aarondl:
My thoughts on code generation (and therefore sqlboiler's output design) is such that the folder is meant to be delete-able. At any point in your app's lifecycle you should be able to delete and regenerate your models from scratch. And therefore it's sort of "taboo" if you will to modify anything within that folder because it's meant to disappear at any time.

It's a fairly common scenario to regenerate your models. Say you delete a table and regenerate your models, sqlboiler will not delete that old model file, so you'll have extraneous code hanging around unless you manually audit the directory. This is a simple scenario where keeping portions of generated code can fail, and therefore we recommend (and should probably document) that the best way to regenerate models is to delete the directory.

Most code generation tools behave in this way and they will usually jam a bunch of comments at the top of each file saying: Do not edit this.

This is the header from go-bindata:
```
// Code generated by go-bindata.
// sources:
// ... elided
// DO NOT EDIT!
```

It's nice to be able to enhance the SQLBoiler models but it should be done on the outside in a separate package otherwise you may have trouble upgrading sqlboiler and regenerating, or schema changes and regenerating etc. I'd actually try to make the claim that this is better organization than your suggestion since differentiating between user code and generated code when you're reading your regular app code is key, consider the difference between the following:

```go
// Is this our code, or sqlboiler code? The only way to find out is to look
// at the filename that contains this code
user.VerifyUserPassword("verify")

// Immediately recognizable as our code
modext.VerifyUserPassword(user, "verify")
```

If you -must- have methods attached to the object you could embed the original model in one of your own types but this would make things quite inconsistent and I'm kind of sorry to have even mentioned the possibility.

I'd suggest that the proper solution here is to agree that outside packages are a safer way to extend what's here as well as perhaps we need a "do not edit" preamble in each of our output files to ensure people understand that they're specifically not to be edited. As well as some documentation in the readme on both "how to regenerate" / "how to extend". Thoughts?

bjyoungblood:
> It's a fairly common scenario to regenerate your models. Say you delete a table and regenerate your models, sqlboiler will not delete that old model file, so you'll have extraneous code hanging around unless you manually audit the directory.

This was my reasoning behind being able to customize the file extension. Any file generated by SQLBoiler would have an identifiable extension (possibly `.boil.go`), so rather than deleting the entire directory, just delete `models/*.boil.go`. Of course, it would then be up to the user to take care of any custom code related to deleted/renamed tables.

> I'd actually try to make the claim that this is better organization than your suggestion since differentiating between user code and generated code when you're reading your regular app code is key

I can definitely see the argument there, and I don't disagree. I think there are different tradeoffs either way. It's better organization from the perspective that you know whether it's generated or not, but it's worse organization when it comes to colocating related code.

I think the question of how to organize extra/custom code could use some further discussion. Off the top of my head, I see two options:

1. Create a `modext` package where you keep everything, but this has the downside of having a lot of package-level functions all in one place, which seems like it could get a bit disorganized:
   ```go
   GetPostURL(p models.Post) *url.URL
   GetSignedJWTForUser(u models.User) (string, err)
   FormatUserProfileName(u models.User) string
   UserCanViewPost(u models.User, p models.Post) (bool, err)
   UserCanEditPost(u models.User, p models.Post) (bool , err)
   VerifyUserPassword(u models.User, verify string) (bool, err)
   ```
   I suppose you could organize this a bit better by attaching the methods to empty structs, but I'm not sure how I feel about that.

2. Create separate packages for each model. This seems like it could also get unwieldy and could create some naming problems (if you have a `users` package, it's going to be hard not to shadow the package with a variable name)

> you could embed the original model in one of your own types

I considered that and agree that it's not really the right way.

> perhaps we need a "do not edit" preamble in each of our output files

IMO, this is a good practice regardless of the greater discussion of mixing generated/non-generated code. I don't think I've encountered a code generator that didn't do this.

> as well as some documentation in the readme on both "how to regenerate" / "how to extend".

I think this would be great -- I'm just getting started with SQLBoiler, and this was my first big question.

aarondl:
This was a great discussion. I'm going to go ahead with the following action items in the very near future (hoping to get to all of them except public slack tomorrow):
- Document extending the models packages
- Document the best practice for regenerating models and adding a `--wipe` flag to help with this
- Add a no-edit warning to the top of the files
- Get a public slack so these discussions can be easier

Though I don't think this particular implementation is well suited for the masses and so I will close this pull request, thanks for helping discuss alternatives and again for offering to take this particular implementation to build scripts.

Issue #100, opened 2017-01-30
Custom type mappings


bjyoungblood:
My main project is an application that uses a lot of financial data. We store all of our financial data in our database as `decimal` fields (as a side note, this type is currently unmapped in the MySQL driver, so our decimals end up as strings -- but this actually turns out to be better than floats).

Ultimately, what we need is to be able to tell SQLBoiler to use an an arbitrary-precision type for any floating-point value (float / double / real / decimal / numeric). This was [very easy to achieve](https://github.com/bjyoungblood/sqlboiler/commit/2100d49196d69c14c4ff76defeec72904351b4e4) by changing a few lines.

One way to control this would be to add a `--float-type` flag (similar to the existing `tinyint-as-bool` option). However, I thought it would be worth discussing whether it would make sense (or even be possible) to give the user a way to override any type mapping.

A few things I'm not quite sure about yet:

- How to define mappings. Based on `DbType` or `FullDbType` -- both seem like they would have limitations (I'm thinking it would be nice to implement `tinyint-as-bool` this way)
- How to deal with Postgres arrays (I haven't had a chance to use SQLBoiler with Postgres yet)
- It could be hard to maintain backwards compatibility, so this might need to be a v3 feature

If this sounds like a feature you'd be interested in, I'd be happy to put together a PR.

aarondl:
This is definitely something we've thought about in the past. It's a complexity trade-off we had written off in favor of mappings that would mostly work for everyone. I wanted to avoid something that would inevitably end up as complicated or more than things like https://github.com/knq/xo#command-line

We actually had a lot of discussion about how to deal with decimal. I'm kinda surprised to see the postgres driver turns them into float64 because as you say - it's basically strictly worse than a string for the use case. That'll have to be a v3 breakage.

One of the challenges with the custom mappings is custom imports as well. As you saw you had to add one to use the decimal library you wanted to do, so it's something to keep in mind. I think the key is to have three different fields that can be used to construct an "and" condition that would check DBType, FullDBType, and the PG-specific (arrays) ArrType.

Another problem is that the tests we generate use a randomization function that knows all the types in use and how to deal with them. I'm not sure that this is something we can overcome without a complete redo of that system, and even then I'm not sure what that system would have to become to make it work.

[...]

PR #102, opened 2017-02-01
[WIP] Driver plugin implementation

bjyoungblood:
Pursuant to discussion in #100, I put together an idea of what driver plugins might look like. As you can see, the implementation was fairly easy and didn't even require any breaking changes.

To use a plugin instead of a built-in driver, it's just `sqlboiler /path/to/plugin.so`.

A few items for consideration and/or discussion:

### Binary/plugin package versioning
The SQLBoiler binary and any plugins must be built with the exact same version of any packages imported by both.

I can see a few strategies that would help plugins to avoid these issues:
  - Plugins can vendor their dependencies.
    - This can potentially cause problems. If SQLBoiler imports the mysql driver, and a plugin imports its vendored mysql driver, the program will panic when the second driver is registered.
    - I haven't tested how well this would work with viper, it would probably be preferable for the plugin to use viper to define ifs own config directly instead of the injection strategy I used in e439620d6b1dcfca752cfd0b85f0d72744736d9c.

  - Plugins should avoid any dependencies on SQLBoiler packages that may change often.
    - The following changes would help with this:
      - Move `bdb.Interface` to a package containing only interfaces (and no code) needed by drivers (the `bdb/drivers` package could eventually become this).
      - Create interfaces for `bdb.Column`, `bdb.PrimaryKey`, and `bdb.ForeignKey`, and change `bdb.Interface` to return these instead of struct types.

  - In general, it would be a best practice for plugins to avoid external dependencies other than the necessary database driver(s).

### Managing plugins
Unless we add some tooling, there are going to be a couple of pain points with managing plugins. The current process for installing and using a driver plugin looks something like this:

```
go get github.com/vattle/sqlboiler-mysql
go build -buildmode=plugin -o /path/to/mysql-driver.so github.com/vattle/sqlboiler-mysql
sqlboiler /path/to/mysql-driver.so
```

To work around this, we could add sub-commands that would help with this process. By default, SQLBoiler could use a directory like `~/.sqlboiler-plugins` to store plugins.

- `sqlboiler get-plugin [-u] github.com/vattle/sqlboiler-mysql`
- `sqlboiler rebuild-plugin github.com/vattle/sqlboiler-mysql`

___

### Alternative Implementation

This is mainly a thought for the future -- I don't think it presents enough of a benefit right now.

We could accept a slice of plugin paths via config and load all of them. Each one would have an `init` function that registers a driver (similar to `database/sql.Register`).

#### Advantages
- Plugins could be something other than a driver (for example, a plugin to generate additional code)
  - An example might be something that adds methods to model slices (similar to the collection methods generated by http://clipperhouse.github.io/gen/)

#### Disadvantages
- This would make it harder for plugins to avoid depending on SQLBoiler directly, so they would probably have to be recompiled more often.
  - Again, this could be painstakingly worked around using interfaces and dependency injection, but 😕 

___

### Trying it out

If you want to try it out, it's pretty easy to do so. If you're on Linux (and using Go 1.8), skip to the last 3 steps.

```
docker run -v "$(pwd):/go/src/github.com/vattle/sqlboiler" -it golang:1.8 bash

cd /go/src/github.com/vattle/sqlboiler

go get ./...
go install .
go build -buildmode=plugin -o mysql.so ./bdb/mysql/plugin.go
sqlboiler mysql.so
```

### DIFF
--- a/bdb/mysql/plugin.go
+++ b/bdb/mysql/plugin.go
@@ -1,0 +1,22 @@
+// +build plugin
+
+package main
+
+import (
+	"github.com/spf13/viper"
+	"github.com/vattle/sqlboiler/bdb"
+	"github.com/vattle/sqlboiler/bdb/drivers"
+)
+
+// InitDriver enables the default mysql driver to be used as a plugin. You could
+// also implement your own custom driver if you wanted.
+func InitDriver() (bdb.Interface, error) {
+	return drivers.NewMySQLDriver(
+		viper.GetString("mysql.user"),
+		viper.GetString("mysql.pass"),
+		viper.GetString("mysql.dbname"),
+		viper.GetString("mysql.host"),
+		viper.GetInt("mysql.port"),
+		viper.GetString("mysql.sslmode"),
+	), nil
+}
--- a/bdb/mysql/plugin.go
+++ b/bdb/mysql/plugin.go
@@ -3,20 +3,31 @@
 package main
 
 import (
-	"github.com/spf13/viper"
 	"github.com/vattle/sqlboiler/bdb"
 	"github.com/vattle/sqlboiler/bdb/drivers"
 )
+
+type configGetter interface {
+	GetString(string) string
+	GetInt(string) int
+}
+
+var config configGetter
+
+func SetConfig(c interface{}) error {
+	config = c.(configGetter)
+	return nil
+}
 
 // InitDriver enables the default mysql driver to be used as a plugin. You could
 // also implement your own custom driver if you wanted.
 func InitDriver() (bdb.Interface, error) {
 	return drivers.NewMySQLDriver(
-		viper.GetString("mysql.user"),
-		viper.GetString("mysql.pass"),
-		viper.GetString("mysql.dbname"),
-		viper.GetString("mysql.host"),
-		viper.GetInt("mysql.port"),
-		viper.GetString("mysql.sslmode"),
+		config.GetString("mysql.user"),
+		config.GetString("mysql.pass"),
+		config.GetString("mysql.dbname"),
+		config.GetString("mysql.host"),
+		config.GetInt("mysql.port"),
+		config.GetString("mysql.sslmode"),
 	), nil
 }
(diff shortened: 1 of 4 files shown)


aarondl:
First off, thanks for this. It's awesome.

Those are some pretty tough issues that we'll have to deal with in terms of managing the plugins. I think regardless we should have some plugin management commands. But it seems like it might be an overcomplication to have to be scared of vendoring and the like and specific versions for builds.

I almost hate to suggest this, but what if we went with a more decoupled plugin structure? We could use binaries that spoke JSON, and simply serialize it into the bdb/drivers.[]Table structure that it expects (from this function https://github.com/vattle/sqlboiler/blob/master/bdb/interface.go#L36). Configuration could be much more free form via ENV vars or JSON on stdin. We would simply have to add some kind of a versioning contract between plugins so we knew if they were within safe ranges. This puts the burden of maintenance of plugin compatibility mostly on the maintainers and not on the users who are trying to compile. Since we use github releases we could release the plugins built for the specific version of sqlboiler there too (and we could add the sqlboiler binary back with the template directories zipped up with a README on how to use it) and have our plugins compiled and deployed by CI.

If that's not appealing, then I think we should try to move forward with what you've got here and we should talk more in depth about what could make this smoother. What do you think?

nati:
I'm using mock driver and generate Table struct from my code now.  I know this is 
hacky way, so it is great if we can register user defined driver even if it isn't using plugin interface.

I really like Idea for having JSON file too.

bjyoungblood:
Binaries that speak JSON to each other makes sense and definitely eliminates the pitfalls associated with plugins. This would have some considerable benefits as well:

- `sqlboiler-mysql | sqlboiler` to generate Go code (we could provide `sqlboiler mysql` for sugar)
- `sqlboiler-mysql > db_schema.json` to commit to source control or to manually tweak the generated struct/field names for some tables/columns
- `sqlboiler < db_schema.json`

As a side note, I saw a package on /r/golang earlier today that invokes `go build` on a plugin prior to attempting to load it: https://github.com/sbinet/plugs.

aarondl:
@bjyoungblood Sorry for leaving this hang for an eternity. I think we'll go with the JSON model, seems everyone agrees with that. :) Not sure when we'll get to it. I'll v3 it if not sooner. PR's welcome of course!

PR #119, opened 2017-03-16
MS SQL engine support

nemec784:
Support of SQL Server

**Limitations:**
* No `mysqldump` equivalent, so generated tests must be supplemented by `tables_schema.sql` with `CREATE TABLE ...` queries
* `sqlcmd` utility must be installed. See https://docs.microsoft.com/en-us/sql/linux/sql-server-linux-setup-tools
* Limit with offset support only for SQL Server 2012 and above. In older versions it can be made only with ugly queries like in http://stackoverflow.com/questions/2135418/equivalent-of-limit-and-offset-for-sql-server
* Values of `UNIQUEIDENTIFIER` type will be treated as `UUID` string, not `GUID`
* NULL values for `BINARY` or `VARBINARY` are not supported  (at this moment) due to bug in `go-mssqldb` driver. See https://github.com/denisenkom/go-mssqldb/issues/196
* All columns with `IDENTITY` will be excluded from Insert and Update clause (even if `IDENTITY_INSERT` is set to `ON`)
* Special types are left behind

**SQL Server 2016 docker images:**
* Windows: https://hub.docker.com/r/microsoft/mssql-server-windows/
* Linux: https://hub.docker.com/r/microsoft/mssql-server-linux/

**Bugs are possible, but tests pass (without FK)**


nullbio:
Wow, there is an impressive amount of work here, thank you. Being primarily Linux users we'd not really given much thought to MS SQL. You've given us a lot to review and discuss, we'll try and get this done as quickly as possible. Thanks again.

README.md
```diff
@@ -319,6 +329,9 @@ sqlboiler -b goose_migrations postgres
 go test ./models
 ```
 
+*Note: No `mysqldump` or `pg_dump` equivalent for Microsoft SQL Server, so generated tests must be supplemented by `tables_schema.sql` with `CREATE TABLE ...` queries*
```
aarondl: Is there no third party command line utility that can do this that we can put a dependency on?

It's also possible that after reading the table, sqlboiler itself could generate the tables_schema.sql script since it has all the db info.

bdb/column.go
```diff
@@ -29,6 +29,11 @@ type Column struct {
 	// tinyint(1) instead of tinyint
 	// Used for "tinyint-as-bool" flag
 	FullDBType string
+
+	// MS SQL only bits
+	// Used to indicate that the value
+	// for this column is auto generated by database on insert (i.e. - timestamp (old) or rowversion (new))
+	AutoGenerated bool
```
aarondl: How does this differ from a default value? To me these read the same. Or is it different in that a value can never be provided for it?

aarondl: Also note that postgres/mysql implement auto-generated columns as a sort of complex default.

bdb/drivers/mock.go
```diff
@@ -118,6 +118,9 @@ func (m *MockDriver) PrimaryKeyInfo(schema, tableName string) (*bdb.PrimaryKey,
 // UseLastInsertID returns a database mock LastInsertID compatibility flag
 func (m *MockDriver) UseLastInsertID() bool { return false }
 
+// UseTopClause returns a database mock SQL TOP clause compatibility flag
+func (m *MockDriver) UseTopClause() bool { return false }
```
aarondl: This is just downright unfortunate. Nothing to do here, just wanted to voice my sadness at the T-SQL spec :)

bdb/drivers/mssql.go
```diff
@@ -0,0 +1,368 @@
...
+
+	return pkey, nil
+}
+
+// ForeignKeyInfo retrieves the foreign keys for a given table name.
+func (m *MSSQLDriver) ForeignKeyInfo(schema, tableName string) ([]bdb.ForeignKey, error) {
+	var fkeys []bdb.ForeignKey
+
+	query := `
+	SELECT
```
aarondl: This query is indented pretty strangely.

randomize/randomize.go
```diff
@@ -52,7 +52,7 @@ var (
 	rgxValidTime     = regexp.MustCompile(`[2-9]+`)
 
 	validatedTypes = []string{
-		"inet", "line", "uuid", "interval", "mediumint",
+		"inet", "line", "uuid", "uniqueidentifier", "interval", "mediumint",
```
aarondl: Is it possible to coerce the value of uniqueidentifier to uuid to make it simpler to deal with?

templates/01_types.tpl
```diff
@@ -32,8 +35,10 @@ var (
 	{{$varNameSingular}}InsertCache = make(map[string]insertCache)
 	{{$varNameSingular}}UpdateCacheMut sync.RWMutex
 	{{$varNameSingular}}UpdateCache = make(map[string]updateCache)
+	{{if ne .DriverName "mssql"}}
```
aarondl: Upsert seems like it's available in a few forms (I don't really know T-SQL, been a -long- time since I wrote it and never did an upsert, hah). Does it not fit nicely with the existing code? Or are there other issues with it?

templates/15_insert.tpl
```diff
@@ -66,7 +70,15 @@ func (o *{{$tableNameSingular}}) Insert(exec boil.Executor, whitelist ... string
 			return err
 		}
 		if len(wl) != 0 {
-			cache.query = fmt.Sprintf("INSERT INTO {{$schemaTable}} ({{.LQ}}%s{{.RQ}}) VALUES (%s)", strings.Join(wl, "{{.LQ}},{{.RQ}}"), strmangle.Placeholders(dialect.IndexPlaceholders, len(wl), 1, 1))
+			{{if ne .DriverName "mssql" -}}
```
aarondl: This is a pretty unhappy section. Before it was just cache/ret sections. But now it's much stranger.

We could actually have the cache.query = `insert into %%s values (%s)`, which leaves the cache.query as a format string set up for the ret portion. Then if there's a retMapping, we replace `%s` with `OUTPUT INSERTED` etc. If not we simply replace it with empty string. That's the only way I can think to keep the code a bit cleaner (though the method to clean the code up is sort of eh). What do you think of this?

templates/15_insert.tpl
```diff
@@ -75,13 +87,15 @@ func (o *{{$tableNameSingular}}) Insert(exec boil.Executor, whitelist ... string
...
 
+		{{if ne .DriverName "mssql" -}}
 		if len(cache.retMapping) != 0 {
 			{{if .UseLastInsertID -}}
-			cache.retQuery = fmt.Sprintf("SELECT {{.LQ}}%s{{.RQ}} FROM {{$schemaTable}} WHERE %s", strings.Join(returnColumns, "{{.LQ}},{{.RQ}}"), strmangle.WhereClause("{{.LQ}}", "{{.RQ}}", {{if .Dialect.IndexPlaceholders}}1{{else}}0{{end}}, {{$varNameSingular}}PrimaryKeyColumns))
-			{{else -}}
-			cache.query += fmt.Sprintf(" RETURNING {{.LQ}}%s{{.RQ}}", strings.Join(returnColumns, "{{.LQ}},{{.RQ}}"))
-			{{end -}}
+			cache.retQuery = fmt.Sprintf("SELECT {{.LQ}}%s{{.RQ}} FROM {{$schemaTable}} WHERE %s", strings.Join(returnColumns, "{{.RQ}},{{.LQ}}"), strmangle.WhereClause("{{.LQ}}", "{{.RQ}}", {{if .Dialect.IndexPlaceholders}}1{{else}}0{{end}}, {{$varNameSingular}}PrimaryKeyColumns))
+			{{else -}}
```
aarondl: Looks like dead whitespace here.

templates/16_update.tpl
```diff
@@ -48,7 +48,15 @@ func (o *{{$tableNameSingular}}) Update(exec boil.Executor, whitelist ... string
 	{{$varNameSingular}}UpdateCacheMut.RUnlock()
 
 	if !cached {
-		wl := strmangle.UpdateColumnSet({{$varNameSingular}}Columns, {{$varNameSingular}}PrimaryKeyColumns, whitelist)
+		wl := strmangle.UpdateColumnSet(
+			{{if ne .DriverName "mssql" -}}
+			{{$varNameSingular}}Columns,
+			{{- else -}}
+			{{$varNameSingular}}ColumnsWithoutAuto,
```
aarondl: Another place that would be cleaned up by coercing "autogenerated" into default/primary key fields as the other dbs if possible.

templates/16_update.tpl
```diff
@@ -157,12 +165,23 @@ func (o {{$tableNameSingular}}Slice) UpdateAll(exec boil.Executor, cols M) error
...
+	{{- else -}}
+	startIndex := len(colNames) + 1
+
+	if !dialect.IndexPlaceholders {
+		startIndex = 0
+	}
+
+	sql := fmt.Sprintf("UPDATE {{$schemaTable}} SET %s WHERE %s",
+		strmangle.SetParamNames("{{.LQ}}", "{{.RQ}}", {{if .Dialect.IndexPlaceholders}}1{{else}}0{{end}}, colNames),
+		strmangle.WhereClauseRepeated(string(dialect.LQ), string(dialect.RQ), startIndex, {{$varNameSingular}}PrimaryKeyColumns, len(o)))
```
aarondl: Does this function exist because MSSQL has no ability to do a "tuple" with in queries? I'll talk to @nullbio on this one, we may want to change the other DB's to use similar functionality to make it so we have less special cases for drivers.

aarondl: Even moreso now that I've scrolled down and saw how often this style of where clause is used.

strmangle/strmangle.go
```diff
@@ -520,6 +521,37 @@ func WhereClause(lq, rq string, start int, cols []string) string {
 	return buf.String()
 }
 
+// WhereClauseRepeated returns the where clause repeated with OR clause using start as the $ flag index
+// For example, if start was 2 output would be: "(colthing=$2 AND colstuff=$3) OR (colthing=$4 AND colstuff=$5)"
+func WhereClauseRepeated(lq, rq string, start int, cols []string, count int) string {
+
```
aarondl: Blank line not needed here.

bdb/drivers/mssql.go
```diff
@@ -0,0 +1,368 @@
...
+	driver := MSSQLDriver{
+		connStr: MSSQLBuildQueryString(user, pass, dbname, host, port, sslmode),
+	}
+
+	return &driver
+}
+
+// MSSQLBuildQueryString builds a query string for MSSQL.
+func MSSQLBuildQueryString(user, pass, dbname, host string, port int, sslmode string) string {
+
```
aarondl: Blank line not needed here.

templates_test/delete.tpl
```diff
@@ -8,7 +8,7 @@ func test{{$tableNamePlural}}Delete(t *testing.T) {
 	seed := randomize.NewSeed()
 	var err error
 	{{$varNameSingular}} := &{{$tableNameSingular}}{}
-	if err = randomize.Struct(seed, {{$varNameSingular}}, {{$varNameSingular}}DBTypes, true); err != nil {
+	if err = randomize.Struct(seed, {{$varNameSingular}}, {{$varNameSingular}}DBTypes, true, {{$varNameSingular}}ColumnsWithDefault...); err != nil {
```
aarondl: These changes are tough to understand by looking at it. Do you mind explaining what this one is for? This could break the other test runs unless this was an oversight by us and it should have been there from the beginning.

templates_test/insert.tpl
```diff
@@ -41,7 +41,11 @@ func test{{$tableNamePlural}}InsertWhitelist(t *testing.T) {
 
 	tx := MustTx(boil.Begin())
 	defer tx.Rollback()
+	{{if ne .DriverName "mssql"}}
 	if err = {{$varNameSingular}}.Insert(tx, {{$varNameSingular}}Columns...); err != nil {
+	{{- else -}}
+	if err = {{$varNameSingular}}.Insert(tx, {{$varNameSingular}}ColumnsWithoutDefault...); err != nil {
```
aarondl: This also seems pretty bizarre, why would it be the exact opposite of the other databases?

templates_test/main_test/mssql_main.tpl
```diff
@@ -0,0 +1,127 @@
...
+		return err
+	}
+	if err = m.createTestDB(); err != nil {
+		return err
+	}
+
+	createCmd := exec.Command("sqlcmd", "-S", m.host, "-U", m.user, "-P", m.pass, "-d", m.testDBName)
+
+	f, err := os.Open("tables_schema.sql")
+
```
aarondl: Blank line problems again :)

templates_test/main_test/mssql_main.tpl
```diff
@@ -0,0 +1,127 @@
...
+	if err = m.dropTestDB(); err != nil {
+		return err
+	}
+	if err = m.createTestDB(); err != nil {
+		return err
+	}
+
+	createCmd := exec.Command("sqlcmd", "-S", m.host, "-U", m.user, "-P", m.pass, "-d", m.testDBName)
+
+	f, err := os.Open("tables_schema.sql")
```
aarondl: I couldn't recall if the Go test will always run the test local to the package.

So I tested by creating a package called testdir that read it's own test file for the test. It passed when run from my homedir as `go test testdir`. This is just a comment for anyone else who might look at this as suspiciously as I did.

templates_test/main_test/mssql_main.tpl
```diff
@@ -0,0 +1,127 @@
...
+}
+
+func (m *mssqlTester) sslMode(mode string) string {
+	switch mode {
+	case "true":
+		return "true"
+	case "false":
+		return "false"
+	default:
+		return "disable"
```
aarondl: Is there a difference for disable vs false in this case?

strmangle/strmangle.go
```diff
@@ -520,6 +521,37 @@ func WhereClause(lq, rq string, start int, cols []string) string {
...
+	for i := 0; i < count; i++ {
+
+		if i != 0 {
+			buf.WriteString(") OR (")
+		}
+
+		if start > 0 {
+			startIndex = start + i*len(cols)
+		} else {
+			startIndex = 0
```
aarondl: Small nit, this code could look like:
```go
startIndex = 0
if start > 0 {
  startIndex = start + i*len(cols)
}
```

It's true that the assignment is then done uselessly, but it's one less jmp instruction. Not sure what modern cpus will do, I just like less else's. This is a suggestion not a direct request to change.

strmangle/strmangle.go
```diff
@@ -520,6 +521,37 @@ func WhereClause(lq, rq string, start int, cols []string) string {
...
+
+	var startIndex int
+
+	buf := GetBuffer()
+	defer PutBuffer(buf)
+
+	buf.WriteByte('(')
+
+	for i := 0; i < count; i++ {
+
```
aarondl: All these blank lines :)

aarondl:
I forgot to mention, excellent work and thank you very much for this! It looks like the code slid in pretty nicely for the most part :)

bdb/column.go
nemec784: Fields with default values can be explicitly inserted or updated, but fields of type rowversion, timestamp - can not.
According to the description of transact-sql (see https://technet.microsoft.com/en-us/library/ms182776(v=sql.110).aspx) those fileds totally controlled by database:

> Every time that a row with a rowversion column is modified or inserted, the incremented database rowversion value is inserted in the rowversion column.

If we try to update or insert data into this column, we will receive an error message.

templates_test/main_test/mssql_main.tpl
nemec784: Yes, according to the driver description:

"encrypt"
- disable - Data send between client and server is not encrypted.
- false - Data sent between client and server is not encrypted **beyond the login packet.** (Default)
- true - Data sent between client and server is encrypted.

bdb/drivers/mssql.go
nemec784: Fixed

templates/15_insert.tpl
nemec784: Fixed

strmangle/strmangle.go
nemec784: Fixed

strmangle/strmangle.go
nemec784: Fixed

templates_test/main_test/mssql_main.tpl
nemec784: Fixed

strmangle/strmangle.go
nemec784: I agree, fixed.

bdb/drivers/mssql.go
nemec784: Fixed

main.go
```diff
@@ -255,6 +255,42 @@ func preRun(cmd *cobra.Command, args []string) error {
...
+		// BUG: https://github.com/spf13/viper/issues/71
+		// Despite setting defaults, nested values don't get defaults
+		// Set them manually
+		if cmdConfig.MSSQL.SSLMode == "" {
+			cmdConfig.MSSQL.SSLMode = "true"
+			viper.Set("mssql.sslmode", cmdConfig.MSSQL.SSLMode)
+		}
+
+		if cmdConfig.MSSQL.Port == 0 {
+			cmdConfig.MSSQL.Port = 3306
```
nullbio: Should this not be port 1433?

nemec784: Fixed

templates/16_update.tpl
nemec784: Yes, unfortunately SQL Server has no ability to do a "tuple" with in queries.

nemec784:
@aarondl, about FK. Yes, tests are passing when foreign keys are disabled.

nemec784:
@aarondl, now all inserts processing without AutoGenerated columns list, but I do not know what to do with the updates, I do not yet see any other way.

templates/15_insert.tpl
nemec784: Yep, you are right. Please review updated code.

bdb/column.go
nemec784: This field now used only in Update statements.

randomize/randomize.go
nemec784: I think this will add more problems than benefit. It's a bit of two different types of data, even though they look the same in the form of a string.

From Wiki:
> The term GUID usually refers to **Microsoft's implementation** of the Universally Unique Identifier (UUID) standard.

templates_test/delete.tpl
nemec784: This parameter is passed in the remaining tests. In this case, it is necessary to exclude columns with automatically generated data from inserting.

templates_test/insert.tpl
nemec784: By default SQL Server prevents insert of explicit values for identity columns. 

If you try to perform an insert with such column, an error will be returned:

> Cannot insert explicit value for identity column in table 'Bar' when IDENTITY_INSERT is set to OFF.

README.md
nemec784: After doing some research I came to the conclusion that everything that is present on the network is oriented to Windows GUI / CLI / Powershell (uses .NET assemblies) or abandoned a long time ago (3-7 years).

We can generate a schema from the code, but the result will not be exactly accurate - there will be no, for example, triggers that can affect the tests results.

nemec784: Also it is possible to generate schema via SQL Server Management Studio (by Microsoft, goes with SQL Server)

For example:
https://blog.sqlauthority.com/2011/05/07/sql-server-2008-2008-r2-create-script-to-copy-database-schema-and-all-the-objects-data-schema-stored-procedure-functions-triggers-tables-views-constraints-and-all-other-database-objects/

glerchundi:
I would like to say such awesome work @nemec784!

templates/01_types.tpl
nemec784: Done, but smells a little.

nemec784:
Now all tests are passing (including Upsert)

aarondl:
Hey there. Planning on re-reviewing very soon. I've been out of town on business and couldn't get to this. Join us on Slack if you'd like to collaborate in something closer to real time: https://sqlboiler.from-the.cloud

nemec784:
Unfortunately, my knowledge of the language and the difference in time zones do not allow me to communicate in real time, but thanks for the invitation.

templates/15_insert.tpl
```diff
@@ -83,13 +75,21 @@ func (o *{{$tableNameSingular}}) Insert(exec boil.Executor, whitelist ... string
 			{{end -}}
 		}
 
+		queryOutput := ""
```
aarondl: Tiny nit.

```go
var queryOutput, queryReturning string
```

I probably would have let this one go, but `queryReturning` is also mispelled so you'll have to touch this code regardless.

templates/17_upsert.tpl
```diff
@@ -73,36 +72,64 @@ func (o *{{$tableNameSingular}}) Upsert(exec boil.Executor, {{if ne .DriverName
...
 		)
+		{{- if eq .DriverName "mssql"}}
+		insert = strmangle.SetComplement(insert, {{$varNameSingular}}ColumnsWithAuto)
+		for i, v := range insert {
+			if strmangle.ContainsAny({{$varNameSingular}}PrimaryKeyColumns, v) && strmangle.ContainsAny({{$varNameSingular}}ColumnsWithDefault, v) {
+				insert = append(insert[:i], insert[i+1:]...)
+			}
+		}
+		if len(insert) == 0 {
+			return errors.New("models: unable to upsert {{.Table.Name}}, could not build insert column list")
```
aarondl: `models:` should be templated to `{{.PkgName}}` like in other places.

templates/17_upsert.tpl
```diff
@@ -73,36 +72,64 @@ func (o *{{$tableNameSingular}}) Upsert(exec boil.Executor, {{if ne .DriverName
...
 			conflict = make([]string, len({{$varNameSingular}}PrimaryKeyColumns))
 			copy(conflict, {{$varNameSingular}}PrimaryKeyColumns)
 		}
-		cache.query = queries.BuildUpsertQueryPostgres(dialect, "{{$schemaTable}}", updateOnConflict, ret, update, conflict, whitelist)
-		{{- else -}}
-		cache.query = queries.BuildUpsertQueryMySQL(dialect, "{{.Table.Name}}", update, whitelist)
+		cache.query = queries.BuildUpsertQueryPostgres(dialect, "{{$schemaTable}}", updateOnConflict, ret, update, conflict, insert)
+		{{- end -}}
+
+		{{if eq .DriverName "mysql"}}
```
aarondl: Possible to clean this section up with `{{else if ...}}` instead of several if/ends.

templates_test/singleton/boil_suites_test.tpl
```diff
@@ -336,5 +336,4 @@ func TestUpsert(t *testing.T) {
   t.Run("{{$tableName}}", test{{$tableName}}Upsert)
   {{end -}}
   {{- end -}}
-}
-{{- end -}}
\ No newline at end of file
+}
```
aarondl: I notice whatever editor you're using is not putting newlines at the end of files. It would be a great favor if you could add that. Plays nice with git and other unix-like tools :) This is merely a friendly request and not a demand.

bdb/column.go
aarondl: Okay, I now understand the necessity of this and agree that we need it in update.

bdb/drivers/mssql.go
```diff
@@ -0,0 +1,367 @@
...
+			c.Type = "float64"
+		case "boolean", "bool", "bit":
+			c.Type = "bool"
+		case "date", "datetime", "datetime2", "smalldatetime", "time":
+			c.Type = "time.Time"
+		case "binary", "varbinary":
+			c.Type = "[]byte"
+		case "timestamp", "rowversion":
+			c.Type = "[]byte"
+		case "uniqueidentifier", "xml":
```
aarondl: Looking at this as a result of the above topic, is this supposed to be a null.String? It's clearly in the non-nullable field section.

randomize/randomize.go
aarondl: I remain unconvinced that there needs to be any separation for this. It's treated the same by randomize.go, and there are no special cases for the field that I can see in the code you've introduced.

There's two cases here:
1. If I have a uniqueidentifier in SQL Server field it will break in the generated tests since the uuid package is being used to generate a V4 UUID and not a GUID per se, then we need to fix this piece.
2. There's no special cases, and we can easily coerce this into the type uuid just to have less special cases (there's obviously plenty already from the introduction of mssql :) Keep in mind the driver still gets to determine what real type a "uuid" produces, so it shouldn't matter so much unless you want randomize.go to treat it differently.

I'm willing to be convinced, but I don't see anything convincing yet :)

templates/01_types.tpl
```diff
@@ -4,6 +4,9 @@
 {{- $tableNameSingular := .Table.Name | singular | titleCase -}}
 var (
 	{{$varNameSingular}}Columns               = []string{{"{"}}{{.Table.Columns | columnNames | stringMap .StringFuncs.quoteWrap | join ", "}}{{"}"}}
+	{{if eq .DriverName "mssql" -}}
+	{{$varNameSingular}}ColumnsWithAuto = []string{{"{"}}{{.Table.Columns | filterColumnsByAuto true | columnNames | stringMap .StringFuncs.quoteWrap | join ","}}{{"}"}}
+	{{- end}}
```
aarondl: Fairly sure it should be `{{end -}}` here. You'll wind up with a newline for no reason if this template doesn't end up evaluating true I think.

templates_test/insert.tpl
aarondl: Hm. We may make this the same for each database too then. I'll ask @nullbio on this one.

templates_test/main_test/mssql_main.tpl
aarondl: Dangerous default. Makes sense though.

templates/16_update.tpl
aarondl: Will talk to @nullbio on this one.

templates_test/singleton/boil_suites_test.tpl
nemec784: Well, finally I found the required parameter in VS Code. Now everything should be fine. I hope =)

templates/15_insert.tpl
nemec784: Fixed

templates/17_upsert.tpl
nemec784: Fixed

templates/17_upsert.tpl
nemec784: Fixed

bdb/drivers/mssql.go
nemec784: "Copy-paste" =(. Fixed

templates/01_types.tpl
nemec784: Fixed

nemec784:
Thanks for review!

I have made changes and now **uniqueidentifier** treats as **uuid**. All other mistakes are also fixed.

templates/16_update.tpl
aarondl: Okay. Done talking. We talked about the maintenance burden of having to ever touch these queries, and updating them in 6 places instead of 3 sort of thing, and setting a precedent for other database engines. If we ever were to add sqlite we don't want to have yet another huge amount of sections saying: "if sqlite", we tried to limit it to exceptional cases for mysql and we should do the same for mssql.

So we should simply use the bottom query (the one you wrote for MSSQL) for all database engines (at least I can't see a reason it wouldn't work for the others) and so I think we should do that. And we should do the same in both `18_delete.tpl` and `19_reload.tpl`.

Also one change to these queries is they're doing an if at runtime to choose startindex, if you see the original query that's done at template time and it should continue to be that way. If it's not runtime dependent it should always be done by templates.

templates_test/insert.tpl
aarondl: In my previous comment I do say we should get rid of all special casing, but this particular change actually weakens the test's purpose so we should leave this alone.

templates_test/upsert.tpl
```diff
@@ -47,4 +47,4 @@ func test{{$tableNamePlural}}Upsert(t *testing.T) {
 	if count != 1 {
 		t.Error("want one record, got:", count)
 	}
-}
+}
```
aarondl: One more newline fail :)

nemec784:
@aarondl, please review the latest changes.

aarondl:
Went to go test this. The postgres and mysql tests are broken. Template formatting errors:

```
Error: unable to generate output: failed to format template

 733 			eventOneColumns,
 734 			eventOneColumnsWithDefault,
 735 			eventOneColumnsWithoutDefault,
 736 			nzDefaults,
 737 			whitelist,
>>>> 		)update := strmangle.UpdateColumnSet(
 739 			eventOneColumns,
 740 			eventOnePrimaryKeyColumns,
 741 			updateColumns,
 742 		)if len(update) == 0 {
 743 			return errors.New("models: unable to upsert event_one, could not build update column list")
```

aarondl:
Also, I haven't looked at it, but it might be smart to make the default schema for mssql "dbo"?

aarondl:
Was working a bit on the CI. It's pretty complicated because typically you've got sqlcmd & the db in a container, and running sqlboiler's tests requires sqlcmd to be in PATH. I wonder if I could fake it out by placing a script in it's place... Hmm.

nemec784:
Fixed. Now the tests are passing for MSSQL and PostgreSQL (MySQL did not check).

nemec784:
@aarondl, I see only one way to set default schema name - remove default value from:
```
rootCmd.PersistentFlags().StringP("schema", "s", "public", "The name of your database schema, for databases that support real schemas")
```
update description to:
```
"The name of your database schema, for databases that support real schemas (defaults: Postgres - public, Sql Server - dbo)"
```
and set it in `preRun` depending on the `driverName`.

nemec784:
@aarondl, it looks like there is an error in circle.yml:

```
    - ./sqlboiler postgres -o "postgres"
    - ./sqlboiler postgres -o "mysql"
```
https://github.com/vattle/sqlboiler/blob/master/circle.yml#L11

nemec784:
@aarondl, sqlcmd can be installed separatly:
https://docs.microsoft.com/en-us/sql/linux/sql-server-linux-setup-tools

And circleci allow to install additional software via apt-get:
https://circleci.com/docs/1.0/installing-custom-software/#installing-via-circleyml

aarondl:
Let's iterate on any remaining pieces. I've done the change for the driver specific schemas. I also noticed that error in the circle.yml yesterday when I opened it and corrected it.

aarondl:
Thanks for all the hard work :)

nemec784:
It was a really hard work, but we did it! Thank you for your time, it was excellent experience for me! :-)

nullbio:
Thank you @nemec784, excellent job.

PR #127, opened 2017-04-16
Fixed writer to use io.Writer interface

Michsior14:
This commit allows devs to use their own loggers.


aarondl:
Would you mind opening against the dev branch? I'm going to add some documentation about contributing today. Sorry about not having it and causing more work for you.

PR #128, opened 2017-04-16
Fixed writer to use io.Writer interface

Michsior14:
This commit allows devs to use their own loggers. Merged to dev instead master - didn't noticed that branch last time. I could work on logging if u want later :)

### DIFF
--- a/boil/global.go
+++ b/boil/global.go
@@ -1,6 +1,7 @@
 package boil
 
 import (
+	"io"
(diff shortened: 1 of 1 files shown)


aarondl:
I'd like to hear a proposal on logging and what we should be adding. Open a new issue detailing your thoughts on the subject and we'll go from there :) Thanks for the PR!

PR #129, opened 2017-04-20
Allow type aliases in structs

liclac:
Given a schema like:

```sql
CREATE TABLE accounts (
    id INTEGER PRIMARY KEY,
    username VARCHAR(25),
    kind SMALLINT
);
```

sqlboiler will generate a struct along the lines of:

```go
type Account {
    ID int
    Username string
    Kind int16

    // …
}
```

So far so good, `go test ./...` passes with flying colors.

However, if I elsewhere in my program had this:

```go
type Kind int16

const (
    NormalUser Kind = iota
    Moderator
    Admin
)

func (k Kind) String() {
    switch k {
    case NormalUser:
        return "Normal User"
    case Moderator:
        return "Moderator"
    case Admin:
        return "Admin"
    default:
        return "[INVALID]"
    }
}
```

…I'd probably want that right in my models, because having to do `pkg.Kind(user.Kind)` everywhere else is kind of a pain. So I'd edit the model and all functions that take an argument called "kind" to use the Kind type instead - probably using a script that wraps `sed` or similar, for easy model regeneration.

However, at this point, unit tests will panic because randomization calls try to assign an `int16` value to a `Kind` field.

This change simply changes all `reflect.Value.Set()` calls to first cast to the desired type. This will still panic if something incompatible is assigned (eg. I changed `int16` to `string` for whatever reason), but will swallow constant types like these.

### DIFF
--- a/randomize/randomize.go
+++ b/randomize/randomize.go
@@ -151,6 +151,11 @@ func randDate(s *Seed) time.Time {
 	return t
 }
 
+// setField assigns a field, but also casts to the field's real type.
+func setField(field, value reflect.Value) {
+	field.Set(value.Convert(field.Type()))
+}
+
 // randomizeField changes the value at field to a "randomized" value.
 //
 // If canBeNull is false:
@@ -171,9 +176,9 @@ func randomizeField(s *Seed, field reflect.Value, fieldType string, canBeNull bo
 
 		if kind == reflect.Struct {
 			val := null.NewString(enum, rand.Intn(1) == 0)
-			field.Set(reflect.ValueOf(val))
+			setField(field, reflect.ValueOf(val))
 		} else {
-			field.Set(reflect.ValueOf(enum))
+			setField(field, reflect.ValueOf(enum))
 		}
 
 		return nil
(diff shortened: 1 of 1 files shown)


aarondl:
Hey @liclac. So we do *strongly* discourage editing models which is what it looks like you did here (not 1000% sure of course, correct me if I'm wrong). Have you considered using MySQL/Postgres enums? https://github.com/vattle/sqlboiler#enums

I'd say this is the supported way to do this. One reason is if you code it as you have above that means the database has no idea that these constraints exist and another database client (that doesn't have your custom enum code) may come along and shove a value you don't want into that field.
Also this survives model regenerations, as anything placed in the output folder (`models` by default) is considered volatile and will eventually be destroyed.

liclac:
Yeah, what I have right now is just a shell script that regenerates the models, then sed-patches in the enum types - I'm patching the json tags to match JSONAPI conventions either way.

I'm not using proper Postgres enums right now mainly because I had a chunk of existing code with its own constants set up, which I will most likely end up refactoring soon, but I figured I may as well get this change out there until then.

aarondl:
Hey @liclac, in the end we decided that we're not going to add code to support this. It really is an anti-pattern to `sed` models post-generation. Sorry :(

PR #133, opened 2017-04-21
fix column_type for non-enum user defined types

makslevental:
fix https://github.com/vattle/sqlboiler/issues/131

### DIFF
--- a/bdb/drivers/postgres.go
+++ b/bdb/drivers/postgres.go
@@ -131,7 +131,7 @@ func (p *PostgresDriver) Columns(schema, tableName string) ([]bdb.Column, error)
 	rows, err := p.dbConn.Query(`
 		select
 		c.column_name,
-		(
+		coalesce(
 			case when c.data_type = 'USER-DEFINED' and c.udt_name <> 'hstore'
 			then
 			(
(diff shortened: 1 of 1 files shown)


aarondl:
What version of postgres is this? I expect that the column type to read as USER-DEFINED. I'm a bit curious why it's null here.

makslevental:
@aarondl `PostgreSQL 9.6.2 on x86_64-pc-linux-gnu, compiled by gcc (Debian 4.9.2-10) 4.9.2, 64-bit`

i can write a docker script that'll demonstrate if you want

aarondl:
That's not necessary. I'll take a look at this soon. Thanks for the PR :)

makslevental:
@aarondl I don't know if this is a great solution to the problem. what I'd really like is the ability to specify a module with "custom types" like knq/xo allows but that'll require some work.

aarondl:
@makslevental We have a solution for this actually. It's just not built. We're going to pull the drivers out of sqlboiler and make them separate binaries that speak JSON. Then you can have any types you want.

aarondl:
See resolution here: https://github.com/vattle/sqlboiler/issues/131

PR #137, opened 2017-04-24
fix removing whitespace

makslevental:
fix https://github.com/vattle/sqlboiler/issues/136


aarondl:
This may leave blank lines, but gofmt should clean them up so I merged to get you going again. Thanks for the PR.

Issue #83, opened 2017-01-05
SQLite Support


jojomi:
If I am not mistaken in the README.md, there is no support for SQLite. Is there plans to add it or is there conceptual problems with it?

nullbio:
If there is a lot of demand for this we can look into it further. From our initial assessments it does seem possible. I'll leave this thread open so we can gauge interest.

aarondl:
@jojomi I wanted to add some more concrete wording around this issue. SQLite is not on the roadmap currently and we don't have a need for it so we won't be implementing it UNLESS there's a huge demand by the community surrounding sqlboiler (like there was for mysql). It 'should' be possible as it's very similar to mysql.

If anyone would like to pursue this path, the first step is always creating a driver in the file bdb/drivers/sqlite.go and ensuring that for similar schemas similar output is achieved. --debug can be used to create a schema dump and you can diff them to ensure two database drivers create the same data for the same schema.

Thanks for bringing this up :)

jojomi:
Thank your for the great tool, I imagine it is a lot of effort, but it sounds like just the thing you would need to have to implement database access in a golang way without being overly repetitive and verbose.

Also thank you for describing what one would have to do to implement a new driver, this might be README.md material :)

robks:
It's better to focus on the big two DBs, yes, but SQLite is good for keeping some data offline - like notes. Thanks for this marvelous work!

nullbio:
@robks Yeah, we'll need to change the way we handle database drivers before we implement sqlite3 support because the driver requires some manual compilation and dependency steps on Windows, and we don't want SQLBoiler to break on all Windows users if they aren't even using sqlite3 to begin with. It is something we want to have eventually though.

aarondl:
This work is complete in the v3 branch and can be done on top of it. There is still some breaking changes coming though I think a large majority of them that would affect drivers in any large way are done.

aarondl:
This is done for a while now.

bradleypeabody:
For posterity and people like me who spent 10 minutes reading through all this looking for this link: https://github.com/volatiletech/sqlboiler-sqlite3

PR #138, opened 2017-04-27
implement (optional) SQLite support

steffengy:
This implements optional SQLite support, which can be enabled using the `sqlite`
build tag.

This is a quite early version, just sharing it because I needed it anyways.

references #83

### DIFF
--- a/bdb/drivers/sqlite.go
+++ b/bdb/drivers/sqlite.go
@@ -53,7 +53,7 @@ func (m *SQLiteDriver) Close() {
 	m.dbConn.Close()
 }
 
-// UseLastInsertID returns false for postgres
+// UseLastInsertID returns false for sqlite
 func (m *SQLiteDriver) UseLastInsertID() bool {
 	return true
 }
@@ -63,7 +63,7 @@ func (m *SQLiteDriver) UseTopClause() bool {
 	return false
 }
 
-// TableNames connects to the postgres database and
+// TableNames connects to the sqlite database and
 // retrieves all table names from sqlite_master
 func (m *SQLiteDriver) TableNames(schema string, whitelist, blacklist []string) ([]string, error) {
 	var args []interface{}
@@ -193,11 +193,21 @@ func (m *SQLiteDriver) Columns(schema, tableName string) ([]bdb.Column, error) {
 		return nil, err
 	}
 
+	query := "SELECT 1 FROM sqlite_master WHERE type = 'table' AND name = ? AND sql LIKE '%AUTOINCREMENT%'"
+	result, err := m.dbConn.Query(query, tableName)
+	if err != nil {
+		return nil, err
+	}
+	autoIncr := result.Next()
+	if err := result.Close(); err != nil {
+		return nil, err
+	}
+
 	for _, column := range tinfo {
 		bColumn := bdb.Column{
 			Name:       column.Name,
-			FullDBType: column.Type,
-			DBType:     column.Type,
+			FullDBType: strings.ToUpper(column.Type),
+			DBType:     strings.ToUpper(column.Type),
 			Nullable:   !column.NotNull,
 		}
 
@@ -212,6 +222,8 @@ func (m *SQLiteDriver) Columns(schema, tableName string) ([]bdb.Column, error) {
 
 		if column.DefaultValue != nil && *column.DefaultValue != "NULL" {
 			bColumn.Default = *column.DefaultValue
+		} else if autoIncr {
+			bColumn.Default = "auto_increment"
 		}
 
 		columns = append(columns, bColumn)
@@ -277,7 +289,7 @@ func (m *SQLiteDriver) ForeignKeyInfo(schema, tableName string) ([]bdb.ForeignKe
 	return fkeys, nil
 }
 
-// TranslateColumnType converts postgres database types to Go types, for example
+// TranslateColumnType converts sqlite database types to Go types, for example
 // "varchar" to "string" and "bigint" to "int64". It returns this parsed data
 // as a Column object.
 // https://sqlite.org/datatype3.html
--- a/bdb/drivers/sqlite.go
+++ b/bdb/drivers/sqlite.go
@@ -1,0 +1,361 @@
+package drivers
+
+import (
+	"database/sql"
+	"fmt"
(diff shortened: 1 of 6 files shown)


aarondl:
Thank you very much for this! I'll attempt some code review soon in the future.

millerlogic:
Thanks for this, I've been hoping for sqlite support.

While giving it a try I ran into an issue with a composite primary key.  When getting the table_info, the pk column should be int instead of bool. It appears to indicate the order.

```sql
sqlite> .head on
sqlite> pragma table_info('mytest');
cid|name|type|notnull|dflt_value|pk
0|a_id|INTEGER|1||1
1|b_id|INTEGER|1||2
```

```
Error: sql: Scan error on column index 5: sql/driver: couldn't convert 2 into type bool
unable to scan for table mytest
github.com/vattle/sqlboiler/bdb/drivers.(*SQLiteDriver).tableInfo
        /home/dev/go/src/github.com/vattle/sqlboiler/bdb/drivers/sqlite.go:136
```

steffengy:
@millerlogic 
Yep, fixed.
```
The "pk" column in the result set is zero for columns that are not part of the primary key
 and is the index of the column in the primary key for columns that are part of the primary key.
```

aarondl:
Hey @steffengy, just to update you on this. We're actively working on a change that will abstract drivers into separate binaries and plan on getting this in after that. Hold on just a little while and we'll get to this! Thanks for your patience.

millerlogic:
Found another thing, the column types are case sensitive. If I create a not null datetime column, the sqlboiler type is string, if I use DATETIME it comes out as time.Time

millerlogic:
It appears that autoincrement isn't working. I try to insert two records and it complains that the UNIQUE constraint is violated on the primary key, it looks like it's trying to insert both with id 0. [Looked up](https://stackoverflow.com/questions/18694393/how-could-you-get-if-a-table-is-autoincrement-or-not-from-the-metadata-of-an-sql) how to detect if the primary key is autoincrement, seems the only way is to look for the word autoincrement for the table's sql in sqlite_master.

steffengy:
I'm not quite sure how to handle AUTOINCREMENT at this point,
The mysql driver simply seems to use `auto_increment` as default value when a column
is defined as `auto increment`, where I can't quite figure out how that works.

Does Mysql interpret `auto_increment` as the next value or is it translated somewhere in sqlboiler?

aarondl:
@steffengy `auto_increment` is a mysql concept. All databases we're using today have the ability to do auto increment as a column. So when you omit the value for it it will insert a special value.

SQLBoiler handles this like so:
- Whenever a pkey column that has a default value we assume it's an auto increment column and that the database will provide a value.
- The `database/sql` library provides https://golang.org/pkg/database/sql/#Result which has `LastInsertId` which in the mysql driver sets here https://github.com/go-sql-driver/mysql/blob/ae5a66df4d33ba404d639bd0fc9189cd5ee79bcf/packets.go#L584

The sqlite drivers should support this: https://github.com/mattn/go-sqlite3/blob/46e826d22aebbfda29523e86e3114c2beb4dab43/sqlite3.go#L833

So basically if your sqlite driver says that it uses lastinsertid (which it does: https://github.com/vattle/sqlboiler/pull/138/files#diff-c880f21b86d224c850e56d2d0ce28501R57) then it should just work using the templates here: https://github.com/vattle/sqlboiler/blob/master/templates/15_insert.tpl#L82

It's possible there's a bug. I can help debug if necessary.

aarondl:
Reading my own message one of the problems could be that the default value detection in the sqlite driver may not be working correctly?

steffengy:
@aarondl 
Thanks that clears things up.

So to detect autoincrement for sqlite I'll have to return a similar default
value as sqlboiler's mysql driver does. 
Detection of that is a bit icky, will probably be very similar
to https://github.com/vattle/sqlboiler/pull/138#issuecomment-301974304.

aarondl:
It should "just work" if the following is bug-free:

Default value checks:
https://github.com/vattle/sqlboiler/pull/138/files#diff-c880f21b86d224c850e56d2d0ce28501R213
Primary key checks:
https://github.com/vattle/sqlboiler/pull/138/files#diff-c880f21b86d224c850e56d2d0ce28501R226

If these are working we should be good to go!

steffengy:
It lacks the equivalent of what the MySQL sqlboiler driver is doing:
 `if(extra = 'auto_increment','auto_increment', c.column_default),`.

Currently there isn't any check for autoincrement, I'll look into getting the equivalent (which is an ugly substring to check for the AUTOINCREMENT keyword in the SQL) in soon.

steffengy:
@millerlogic Does this fix your issue with auto increment?

millerlogic:
@steffengy it works if line 196 uses `?` instead of `%s`, thanks!

steffengy:
Well OOPS. Still happy to hear that!

maxmeyer:
Any progress on that PR?

aarondl:
@maxmeyer The thing that's blocking it is the v3 way of doing drivers and the fact that merging it basically stops people from compiling `sqlboiler` on Windows unless they get the sqlite3 dll and mess around (same on Linux really, it just happens to be trivially easy on Linux).

We haven't been actively developing against v3 for some time now unfortunately but it's never left the backlog of things to do. We're about to kick off some new dev, and I think we might do it as a precursor to it.

maxmeyer:
@aarondl That sounds good! Thanks for the update.

aarondl:
Super happy to say that the work preceding this is finally done.

Here's an example of an external driver in the v3 branch:
https://github.com/volatiletech/sqlboiler/tree/v3/drivers/sqlboiler-psql

Because of the compilation trouble with this driver, I think it should remain outside the tree however. I will refactor and move it out if no one gets to it before me (this will be the last thing in v3 that I do, so there's plenty of time for someone to get to it before then!)

aarondl:
There now exists an sqlboiler v3 driver for sqlite3: https://github.com/volatiletech/sqlboiler-sqlite3

I'll be announcing this with the release of v3 sometime very soon :)

aarondl:
@steffengy If you desire we can add your name to copyrights in the license if you wish.

PR #140, opened 2017-04-28
*.gen suffix added to generated files

FourSigma:
This suffix will help differentiate generated source files in a project. Generated tests have the suffix gen_test.go


aarondl:
Hey @FourSigma. This PR makes it look like you're putting your own extension code into the same folder/package as the sqlboiler generated code. As seen in previous PR's and Slack discussions this is a pattern that is highly discouraged. Extensions should be placed outside the models.

See historical discussions.
https://github.com/vattle/sqlboiler/pull/96
https://sqlboiler.slack.com/archives/C3VDH4BU1/p1492271034443179

The recommended ways to extend models are documented here:
https://github.com/vattle/sqlboiler/#extending-generated-models

nullbio:
I'd also add that if you plan on contributing it's probably worth having a quick chat first (on an issue or on slack), just so you don't end up wasting time on something we have no intention of merging, and in case any unconsidered requirements come up. Thanks.

FourSigma:
@aarondl Would it be possible to add a suffix option to the command line parameters?

nullbio:
@FourSigma This is something we considered and decided against in the early days of making SQLBoiler, because it encourages patterns that we consider to be dangerous, for example: https://github.com/vattle/sqlboiler/pull/129#issuecomment-29652

People should be able to know that *every* file in the models folder is generated, and thus an extension is not necessary. Only would an extension be necessary if people are putting files inside the generated models package, which is something that should not be done.

PR #157, opened 2017-06-07
Mysql - decimal type fields treated as string

rmsj:
Decimal type fields are falling into the "default" and treated as string.
We should treat them in the same case as double, real, etc.

### DIFF
--- a/bdb/drivers/mysql.go
+++ b/bdb/drivers/mysql.go
@@ -310,7 +310,7 @@ func (m *MySQLDriver) TranslateColumnType(c bdb.Column) bdb.Column {
 			}
 		case "float":
 			c.Type = "null.Float32"
-		case "double", "double precision", "real":
+		case "double", "double precision", "real", "decimal":
 			c.Type = "null.Float64"
 		case "boolean", "bool":
 			c.Type = "null.Bool"


aarondl:
Hey there. So this has been brought up before, and it's actually in fact largely intentional that it's a string because of two reasons:
- Decimal types are fixed precision decimal numbers, whereas float64 is not. This impedence could cause people to lose/gain actual money in a web app for example. (And widen us all up to the Superman 3 hack!). So the great thing about this is you can use this in basically all ways that you'd want for display and input purposes, and it only breaks down when people go to do math with it, which is in fact where the problem lies.
- The database/sql package will transform numeric values into strings and vice versa for you (meaning no extra work for sqlboiler)

My comment on this in the past was we weren't looking to add a decimal library out of the box
(though maybe we should just bite that bullet honestly) and that we were going to eventually
allow people to use custom drivers that would be able to handle any data type in the way that
you want (like making decimal a float64, or a real decimal library if you're keen on that approach).

The good news is that the custom driver-ability is actually the next feature on our plate. :)
I'll wait for your reply, but my intention is to close this as I think what we're currently doing is safe and not
misleading for people making apps. As well as we have a solid way going forward for people to support this
properly by using a decimal library.

PR #161, opened 2017-06-14
Fix randomization for enums with single value

gsamokovarov:
Hey there,

I hit a bug that panicked the generated tests with:

    panic: invalid argument to Intn

The problem happened, because I have a Postgres `ENUM` with a single
value, that caused a `rand.Intn(0)` call, which panics on zero.

See: https://golang.org/src/math/rand/rand.go?s=4112:4142#L129


nullbio:
Thank you for fixing this. @aarondl Looks good to merge, yeah?

PR #163, opened 2017-06-16
Add columns name as constant

guytish:


nullbio:
If we're going to go down this route we should add table names and qm.Load parameter names (which are table names in a different format) as constants as well, to keep it consistent. For example, what should I be passing into qm.Load, "cool_cars", "CoolCars", "cool_car", "CoolCar", etc. It's not particularly clear and the most logical format depends on the perspective you hold, "am I writing the table name as is, or am I writing it based on the type of association?".

aarondl:
Hey there @guytish. Is there some sort of ideas you can give us on why this is desirable or what use case this enables? I have some ideas myself but I'm curious what yours are.

viaguyt:
Hi, thanks for the quick responses.
Thoughts:
1. Code readability.
2. Avoid typos.
3. Less strings allocations.

What do you have in mind?

aarondl:
Hey @guytish, we were talking about this and one of the things we try to do is cut down on the amount of identifiers there are for autocomplete against `models.`. This would sort of balloon it. We thought maybe it makes sense to put them into anonymous structs so that you have `models.ModelName.ID` as example. What do you think of this alternative?

PR #169, opened 2017-06-22
Fix area plural issue, add testcase

### LINKED ISSUE #168: Word `areas` will not convert to singular `area`. Generates name collision.

SQLBoiler v2.4.0

Using MySQL.

```sql
CREATE TABLE IF NOT EXISTS `areas` (
  `id` INT UNSIGNED NOT NULL AUTO_INCREMENT,
  `name` VARCHAR(128) NULL DEFAULT NULL,
  PRIMARY KEY (`id`)
);
```

Hi there, so the issue seems to be the table name `areas`. It will generate the struct `Areas` and also the global function `Areas`, resulting in a compilation error:

```bash
$ go build
# webservice/models
models/areas.go:836: Areas redeclared in this block
	previous declaration at models/areas.go:25
```

If I change the name to `cars`, everything works. I verified this by adding `areas` to the `TestSingular` test for `strmangle` and it failed, see diff below. Adding `areas` to `TestPlural` works as expected, the issue seems to be to convert `areas` to `area`. The diff below also contain a fix, but it might not be the general fix (why this is an issue and not a pull request).  It just adds `areas` and `area` as specific singulars in `strmangle/inflect.go`. 

```diff
diff --git a/strmangle/inflect.go b/strmangle/inflect.go
index 6238c39..c9ca8e4 100644
--- a/strmangle/inflect.go
+++ b/strmangle/inflect.go
@@ -100,6 +100,7 @@ func newBoilRuleset() *inflect.Ruleset {
        rs.AddSingular("ta", "tum")
        rs.AddSingular("ia", "ium")
        rs.AddSingular("analyses", "analysis")
+       rs.AddSingular("areas", "area")
        rs.AddSingular("bases", "basis")
        rs.AddSingular("diagnoses", "diagnosis")
        rs.AddSingular("parentheses", "parenthesis")
diff --git a/strmangle/strmangle_test.go b/strmangle/strmangle_test.go
index b07cee6..66bf582 100644
--- a/strmangle/strmangle_test.go
+++ b/strmangle/strmangle_test.go
@@ -139,6 +139,7 @@ func TestSingular(t *testing.T) {
                {"hello_people", "hello_person"},
                {"hello_person", "hello_person"},
                {"friends", "friend"},
+               {"areas", "area"},
                {"hello_there_people", "hello_there_person"},
        } 
```

Unclear how this affect other words but it's a problem for us. 

Thanks

jfernstad:
- Fix #168 

Added specific case for the word `area` as discussed in #168

### DIFF
--- a/strmangle/inflect.go
+++ b/strmangle/inflect.go
@@ -193,6 +193,7 @@ func newBoilRuleset() *inflect.Ruleset {
 	rs.AddSingularExact("is", "is", true)
 	rs.AddSingularExact("us", "us", true)
 	rs.AddSingularExact("as", "as", true)
+	rs.AddSingularExact("areas", "area", true)
 	rs.AddPluralExact("a", "a", true)
 	rs.AddPluralExact("i", "i", true)
 	rs.AddPluralExact("is", "is", true)
--- a/strmangle/strmangle_test.go
+++ b/strmangle/strmangle_test.go
@@ -139,6 +139,7 @@ func TestSingular(t *testing.T) {
 		{"hello_people", "hello_person"},
 		{"hello_person", "hello_person"},
 		{"friends", "friend"},
+		{"areas", "area"},
 		{"hello_there_people", "hello_there_person"},
 	}
 
@@ -159,6 +160,7 @@ func TestPlural(t *testing.T) {
 		{"hello_person", "hello_people"},
 		{"friend", "friends"},
 		{"friends", "friends"},
+		{"area", "areas"},
 		{"hello_there_person", "hello_there_people"},
 	}
 


strmangle/inflect.go
```diff
@@ -94,6 +94,7 @@ func newBoilRuleset() *inflect.Ruleset {
 	rs.AddSingular("s", "")
 	rs.AddSingular("ss", "ss")
 	rs.AddSingular("as", "as")
+	rs.AddSingularExact("areas", "area", true)
```
aarondl: I hate to make a comment on this PR that requires changes. But this should be with the other `AddSingularExact` calls.

jfernstad:
Understood. I put it there for the context of the row above that is probably causing the issues for us. I'll move it.

jfernstad:
Squashed updated fix.

aarondl:
Thanks very much. :)

PR #170, opened 2017-06-26
Added table columns and table names as anonymous struct

guytish:


aarondl:
Merged to `dev`. e58ab28

PR #174, opened 2017-07-05
Fix examples of Find in README

tchssk:


aarondl:
We thought we merged this already. Sorry it took a while.

PR #177, opened 2017-07-17
Add LAST_INSERT_ID to MySQL upsert query

tpisani:
It fixes an error caused by a query that uses the last insert ID to populate default values for structs after an upsert query at https://github.com/vattle/sqlboiler/blob/master/templates/17_upsert.tpl#L200, currently using `0` as the ID:

```
models: unable to populate default values for <model>: sql: no rows in result set
```

### DIFF
--- a/queries/query_builders.go
+++ b/queries/query_builders.go
@@ -231,6 +231,13 @@ func BuildUpsertQueryMySQL(dia Dialect, tableName string, update, whitelist []st
 		buf.WriteByte(')')
 	}
 
+	quoted := strmangle.IdentQuote(dia.LQ, dia.RQ, "id")
+	buf.WriteByte(',')
+	buf.WriteString(quoted)
+	buf.WriteString(" = LAST_INSERT_ID(")
+	buf.WriteString(quoted)
+	buf.WriteByte(')')
+
 	return buf.String()
 }
 


aarondl:
So we already have functionality in there to utilize the last insert id. I get the feeling you're looking for something else. We're going to need more details on what you're trying to accomplish and what's not working in order to figure out the proper solution. At first glance this doesn't seem necessary. Can you elaborate on what's going on?

tpisani:
I was trying to do something like this:
```go
type Team struct {
    ID         int
    ExternalID int
    Name       string
    Initials   string
}

type TeamStore struct {
    db *sql.DB
}

func (s TeamStore) SaveTeam(t *Team) error {
    o := models.Team{
        ExternalID: t.ExternalID,
        Name:       t.Name,
        Initials:   t.Initials,
    }
    err := o.Upsert(s.db, []string{"name", "initials"})
    t.ID = o.ID
    return err
}
```
_`ExternalID` is a field I'm using to store a unique ID from a third-party API._

But it kept returning this error:
```
models: unable to populate default values for teams: sql: no rows in result set
```

I noticed that the value of `o.ID` was `0`, then I checked the docs for something that could fill the last insert ID correctly but I failed to find a way to do it. So I decided to take a look at the upsert query and it looked like this:
```sql
INSERT INTO teams (`external_id`, `name`, `initials`) VALUES (?,?,?) ON DUPLICATE KEY UPDATE `name` = VALUES(`name`),`initials` = VALUES(`initials`)
```

And the query for updating default values that runs immediately after looked like this:
```sql
SELECT `id` FROM `teams` WHERE `id`=?
0
```

Then I tried to modify the query to "return" the last insert ID by adding `id = LAST_INSERT_ID(id)`, which sets `id` to the current ID if an update has occurred.

aarondl:
@tpisani Sorry I was on vacation. Back now :)

So the [code I mentioned before](https://github.com/vattle/sqlboiler/blob/master/templates/17_upsert.tpl#L174) is supposed to do this by getting the `last_insert_id` from the `mysql` driver. Maybe you're using a mysql driver that doesn't support it? Can you tell me which driver you're using?

It's possible it has a bug, but again I don't think Pull Request isn't the answer.

Is it possible you can give me the mysql schema involved here for the `Team` model?

tpisani:
It's OK. 😉 

Maybe the problem is the driver I'm using then: https://github.com/go-sql-driver/mysql.

Here is the create statement for `teams` table:
```sql
CREATE TABLE `teams` (
  `id` int(11) NOT NULL AUTO_INCREMENT,
  `external_id` int(11) NOT NULL,
  `name` varchar(100) NOT NULL,
  `initials` varchar(10) NOT NULL,
  PRIMARY KEY (`id`),
  UNIQUE KEY `external_id` (`external_id`)
);
```

Thank you for your time and patience.

aarondl:
Fixed in v3 branch. Sorry for the year delay :)

PR #183, opened 2017-08-03
fix null.Int64 when bigint is signed

ceshihao:
BIGINT should distinguish signed or unsigned for type int64 or uint64.


aarondl:
Thanks for this :)

ceshihao:
👍 cool

Issue #185, opened 2017-08-11
Ignore FK relationship if FK table is not in whitelist or in blacklist


ceshihao:
If you're having a generation problem please answer these questions before submitting your issue. Thanks!

### What version of SQLBoiler are you using (`sqlboiler --version`)?
SQLBoiler v2.5.1

### If this happened at generation time what was the full SQLBoiler command you used to generate your models? (if not applicable leave blank)
yes

### What is the output of the command above with the `-d` flag added to it? (Provided you are comfortable sharing this, it contains a blueprint of your schema)

```bash
sqlboiler --basedir ./ mysql -d
panic: could not find table name: pilots

goroutine 1 [running]:
panic(0x4dd940, 0xc42010a9d0)
	../.gvm/gos/go1.7.5/src/runtime/panic.go:500 +0x1a1
../github.com/volatiletech/sqlboiler/bdb.GetTable(0xc4200a81b0, 0x1, 0x1, 0xc42010a9a8, 0x6, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
	../github.com/volatiletech/sqlboiler/bdb/table.go:30 +0x23e
../github.com/volatiletech/sqlboiler/bdb.setForeignKeyConstraints(0xc4200a81b0, 0xc4200a81b0, 0x1, 0x1)
......
```

### Please provide a relevant database schema so we can replicate your issue (Provided you are comfortable sharing this)
```sql
CREATE TABLE pilots (
  id integer NOT NULL,
  name text NOT NULL
);

ALTER TABLE pilots ADD CONSTRAINT pilot_pkey PRIMARY KEY (id);

CREATE TABLE jets (
  id integer NOT NULL,
  pilot_id integer NOT NULL,
  age integer NOT NULL,
  name text NOT NULL,
  color text NOT NULL
);

ALTER TABLE jets ADD CONSTRAINT jet_pkey PRIMARY KEY (id);
ALTER TABLE jets ADD CONSTRAINT jet_pilots_fkey FOREIGN KEY (pilot_id) REFERENCES pilots(id);
```

### Further information. What did you do, what did you expect?
my sqlboiler.toml
```
whitelist=["jets"]

[mysql]
...
```

In my project there are lots of tables with FKs.
However, my app using sqlboiler only touches some of them, and I use `whitelist` option for models generation.

Current sqlboiler throws a panic out if FK table is not in whitelist or in blacklist, even though I do NOT need the relationship. 

[...]

PR #186, opened 2017-08-11
ignore FK if FK table not in whitelist or in blacklist

ceshihao:
A PR for #185 
Ignore FK relationship if FK table is not in whitelist or blacklist.

### DIFF
--- a/bdb/drivers/mock.go
+++ b/bdb/drivers/mock.go
@@ -59,7 +59,7 @@ func (m *MockDriver) Columns(schema, tableName string) ([]bdb.Column, error) {
 }
 
 // ForeignKeyInfo returns a list of mock foreignkeys
-func (m *MockDriver) ForeignKeyInfo(schema, tableName string, whitelist, blacklist []string) ([]bdb.ForeignKey, error) {
+func (m *MockDriver) ForeignKeyInfo(schema, tableName string) ([]bdb.ForeignKey, error) {
 	return map[string][]bdb.ForeignKey{
 		"jets": {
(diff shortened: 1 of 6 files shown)


aarondl:
Nice. This is a good feature to have. But I can't help but notice there's a lot of repetition at the driver level. I wonder if this functionality couldn't be created as a filter that runs around here: https://github.com/vattle/sqlboiler/blob/master/bdb/interface.go#L72

This absolves the drivers of the responsibility, creates fewer arguments to the functions downstream, and should also be able to provide the same functionality. What do you think?

ceshihao:
Perhaps it can be added in func setForeignKeyConstraints.

https://github.com/volatiletech/sqlboiler/blob/master/bdb/interface.go#L118

aarondl:
I'd say it should just be a new function, easier to test that way. It should take the foreign keys that exist + the list of tables, and cull foreign keys that don't belong to a table that exists in the list. I think this will work, yes?

ceshihao:
move it to func filterForeignKeys

aarondl:
Merged into `dev` branch: 9be8f90. This is our staging ground for "bigger" releases and is how you should submit PRs against dev branch usually (see [CONTRIBUTING.md](https://github.com/volatiletech/sqlboiler/blob/master/CONTRIBUTING.md)). Please help test it out :)

Thanks so much for your contributions!

ceshihao:
OK, i will submit new PR to dev branch.

Sorry for inconvenience.

saulortega:
This feature is not yet in version 2.6?

PR #193, opened 2017-08-23
use interface to enable custom Tx types

lyoshenka:
Use Transactor interface instead of *sql.Tx so developers are free to use a custom Tx type.

### DIFF
--- a/boil/db.go
+++ b/boil/db.go
@@ -19,7 +19,7 @@ type Transactor interface {
 
 // Beginner begins transactions.
 type Beginner interface {
-	Begin() (*sql.Tx, error)
+	Begin() (Transactor, error)
 }
 
 // Begin a transaction


aarondl:
This does make some sense. Will probably end up merging.

lyoshenka:
Upon further testing, I dont think this works. Turns out Go needs the return types to match exactly, or it does not consider an interface satisfied. So `*sql.DB` will not satisfy `Beginner` because it has 

    Begin() (*sql.Tx, error)

and not 

    Begin() (Transactor, error)

and Go does not care that `*sql.Tx` satisfies `Transactor`.

lyoshenka:
It's still possible to do this if you want to wrap `*sql.Tx` in some kind of wrapper satisfies the interfaces exactly and passes the calls through (which was my original intent with this, to create a transparent wrapper that logs all queries and passes them on). My guess is you don't want to do that though.

aarondl:
Well, my question would then be why don't you just create a thin wrapper for your code instead of having it be in the `sqlboiler` codebase. It seems like the problem is fairly localized to what you're doing. If it can't easily be generic I don't think we should fight to abstract it here. What do you think? Is it possible to keep it in your code?

lyoshenka:
I don't think its possible, because I have no control over the generated tests. At the very least the tests would have to be updated to use a wrapper.

lyoshenka:
Actually, it looks like all you need is https://github.com/volatiletech/sqlboiler/pull/205

PR #198, opened 2017-08-27
Modify disclaimer to make generated files be ignored by golint

VeryCB:
There is a new standard about generated file disclaimer. According to https://golang.org/s/generatedcode

```
Generated files are marked by a line of text that matches the regular expression, in Go syntax:

^// Code generated .* DO NOT EDIT\.$

The .* means the tool can put whatever folderol it wants in there, but the comment must be a single line and must start with Code generated and end with DO NOT EDIT., with a period.

```

Generated files with a disclaimer match above regular expression will be ignored by golint as well. That's actually what I want in my project.


aarondl:
Merged into dev branch: 5f4d922. This is our staging ground for "bigger" releases and is how you should submit PRs against dev branch usually (see CONTRIBUTING.md). Please help test it out :)
This is great, thanks! Wasn't aware this new standard existed.

PR #201, opened 2017-08-31
Fix missed `boil.Executor`s in `queries.Raw` calls

gsamokovarov:
I have stumbled on them reading the `queries.Raw` documentation.


aarondl:
Thanks :D

gsamokovarov:
🍻

PR #205, opened 2017-09-02
use interface to enable custom Tx types (fixed)

lyoshenka:
Updated version of https://github.com/volatiletech/sqlboiler/pull/193

### DIFF
--- a/boil/db.go
+++ b/boil/db.go
@@ -19,6 +19,11 @@ type Transactor interface {
 
 // Beginner begins transactions.
 type Beginner interface {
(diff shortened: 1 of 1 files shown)


aarondl:
Hey @lyoshenka, I don't really like having two separate interfaces that do the same thing, one being just slightly more generic. As we talked about in the previous pull request because there's nothing stopping you from including this in your own code and I think it's a bit of a special use case I think it should stay on the outside. I'm opened to hearing other arguments, but I'll close in the meantime. Sorry :(

PR #207, opened 2017-09-14
Expose DBType for columns

ceshihao:
In my project, I need DBType information for some business.

Therefore, I change the struct from string (DB column Name) to Column struct (Includes Name and DBType).

### DIFF
--- a/boilingcore/imports.go
+++ b/boilingcore/imports.go
@@ -171,6 +171,7 @@ func newImporter() importer {
 		},
 		thirdParty: importList{
 			`"github.com/pkg/errors"`,
+			`"github.com/volatiletech/sqlboiler/bdb"`,
 			`"github.com/volatiletech/sqlboiler/boil"`,
 			`"github.com/volatiletech/sqlboiler/queries"`,
 			`"github.com/volatiletech/sqlboiler/queries/qm"`,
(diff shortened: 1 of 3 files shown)


aarondl:
I'm fairly certain you can already get this information from the Table structure?

In `templateData` struct you have the member `Tables` as well as `Table` which is the current table:
https://github.com/volatiletech/sqlboiler/blob/master/boilingcore/templates.go#L18

The `Table` type is here which includes a list of `bdb.Column`
https://github.com/volatiletech/sqlboiler/blob/master/bdb/table.go#L6

In order to get the columns you can use this helper:
https://github.com/volatiletech/sqlboiler/blob/master/bdb/table.go#L33

There's helpers around columns here (some are actually exposed already in the templates too):
https://github.com/volatiletech/sqlboiler/blob/master/bdb/column.go#L39

I'd suggest you try my way first, if that's not doable lets re-evaluate and re-open. Thanks for submitting :) Sorry I got to this really late. Got busy with real life stuff and then a long business trip :(

ceshihao:
Do you mean I can use the whole bdb.Column variable instead of this?

https://github.com/volatiletech/sqlboiler/pull/207/files#diff-badd4b26638a77e3a1b14c4dde10efa1R26

In my case, I need dbtype in generated models so that my app can get dbtype and no need to access database.

aarondl:
Why does the app need that though? If anything it should just use the Go type, which is available via `reflect` if you need.

PR #210, opened 2017-09-26
make generated file comments compliant with rules

tarnowsc:
in order to make generated code ignored by linter it should follow the rules described here:
https://golang.org/s/generatedcode


ceshihao:
similar to #198 ?

aarondl:
Yeah, this is done in dev branch already. Thanks!

PR #219, opened 2017-11-09
Fix HAVING clause to join multiple conditions

knqyf263:
When having clause has multiple conditions, they are joined by `,`.
Then, the following error occurs.
```
mysql> SELECT * FROM personal GROUP BY id, name HAVING id <>1, length(name) > 5;
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near ' length(name) > 5' at line 1
```

I think they should be joined by `AND` or `OR`.
I support `AND` for the time being.

```
mysql> SELECT * FROM personal GROUP BY id, name HAVING id <>1 AND length(name) > 5;
+------+--------+
| id   | name   |
+------+--------+
|    2 | foobar |
+------+--------+
1 row in set (0.00 sec)
```


aarondl:
Thanks, it's sort of funny we missed this :)

PR #226, opened 2017-11-24
Query Select Hooks

pouyanh:
I'm using sqlboiler on a database containing spatial data and wanted to retrieve raw geometry data as binary on all direct/indirect retievements (finishers, eager loads) and convert them using [go-geom](https://github.com/twpayne/go-geom).

In order to do that:
* I've changed all geometry columns translation type to []byte
* Added QueryHooks to be able to change the query before commit esp. SELECT

Now I can make my geometry column to be queried as WKB on Finishers, **Eager Loads**, Reloads and Finds by adding the query hook as below:

```go
models.AddGeoAreaQueryHook(boil.SelectHook, func (exec boil.Executor, q *queries.Query) error {
	queries.AppendSelect(q,
		fmt.Sprintf("%s.*", models.TableNames.GeoAreas),
		fmt.Sprintf("ST_AsBinary(%s) AS %s", models.GeoAreaColumns.Geometry, models.GeoAreaColumns.Geometry),
	)

	return nil
})
```

And then scanning the geometry field using go-geom/wkb or any other decoders
```go
	var mp wkb.MultiPolygon
	if err := mp.Scan(geoArea.Geometry); nil == err {
		...
	}
```

### DIFF
--- a/boilingcore/text_helpers.go
+++ b/boilingcore/text_helpers.go
@@ -19,6 +19,7 @@ type TxtToOne struct {
 	}
 
 	ForeignTable struct {
+		Name        string
 		NameGo       string
 		NamePluralGo string
 		ColumnNameGo string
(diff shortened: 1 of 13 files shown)


boilingcore/text_helpers.go
```diff
@@ -19,6 +19,7 @@ type TxtToOne struct {
 	}
 
 	ForeignTable struct {
+		Name         string
```
aarondl: I don't actually see anywhere where this is used. Also this data is available already on the ForeignKey and Relationship structs so I'm not sure this is necessary?

Issue #196, opened 2017-08-26
Multiple database environment configuration selection


ivegotasthma:
Hello,

I have multiple databases setup for my application. I have a `prod`, `dev` and `test`, each of which has different credentials. Currently `sqlboiler` doesn't support switching between those in the config file. I'm also using `sql-migrate` who's config has all the variables that you want to have. My feature request is such: Allow a configuration in the format of `sql-migrate` to be specified to the `sqlboiler` executable with the `-config` flag and an `-env` flag which selects which environment of the driver to select. Do you think this is reasonable?

aarondl:
It's sort of at a higher level than the sqlboiler tool is at. I'm not sure if it's necessary to put it in. [abcweb](https://github.com/volatiletech/abcweb) is an example of a tool that brings the prod/dev/test sort of multiple environments along with it, embeds sqlboiler config, and then passes it in and I think this is probably the correct level for that functionality.

I think you could argue it's a bit odd that there's no `-c`, `--config` flag and I'd accept that to more easily accomodate this kind of workflow, so that you could have `sqlboiler-prod.toml` `sqlboiler-dev.toml` etc. I know that's not 100% the desired outcome, since the original request has nice things like common config that's shared between everything at the top level and then overidden at the lower level by env specific. But it also comes at the price of more code :p - though I do think viper can do this kind of straight-forwardly as the project above used it. Just seems like it's not worth the complexity at this level. Do any of these workarounds seem reasonable?

aarondl:
Closing for inactivity.

ivegotasthma:
Sorry for the late replay. I'd be okay with `--config` because every command I use is under a `Makefile` so command usability is not that big of an issue. I think this is the most minimal change that will fix the situation.

aarondl:
Okay, well I'd definitely accept a PR like that. Thanks for getting back to me.

PR #229, opened 2017-11-25
Add support for config flag

l4u:
related #196

### DIFF
--- a/main.go
+++ b/main.go
@@ -17,22 +17,22 @@ import (
 const sqlBoilerVersion = "2.6.0"
 
 var (
-	cmdState  *boilingcore.State
-	cmdConfig *boilingcore.Config
+	flagConfigFile string
+	cmdState       *boilingcore.State
+	cmdConfig      *boilingcore.Config
 )
 
-func main() {
-	var err error
-
-	// Too much happens between here and cobra's argument handling, for
-	// something so simple just do it immediately.
-	for _, arg := range os.Args {
-		if arg == "--version" {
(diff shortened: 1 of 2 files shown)


ryfloyd:
@l4u Won't this template `initViper` function need to change as well? https://github.com/volatiletech/sqlboiler/blob/master/templates_test/singleton/boil_main_test.tpl#L61

main.go
```diff
@@ -17,45 +17,51 @@ import (
 const sqlBoilerVersion = "2.6.0"
 
 var (
+	cfgFile   string
```
aarondl: For global vars (usually flags) I prefer to keep them separated by their type in some way, usually I'd call this `flagConfigFile`, there's no reason to abbreviate to `cfg` either.

main.go
```diff
@@ -71,7 +77,10 @@ func main() {
 		SilenceUsage:  true,
 	}
 
+	cobra.OnInitialize(initConfig)
+
 	// Set up the cobra root command flags
+	rootCmd.PersistentFlags().StringVar(&cfgFile, "config", "", "config file")
```
aarondl: The comment here should be more descriptive. "Use this flag to override the default config file lookup" or something along those lines

l4u:
Thanks for reviewing it. @aarondl 
Rebased and added changes to the boil_main_test.tpl

Dexus:
@aarondl can this be added shortly? need the function for multiple databases that used in on of our project.

Thanks, Josef

aarondl:
Will line this up for release when I figure out what's going on with these fixes around null foreign keys in #265 and #269

PR #233, opened 2017-12-08
Fix enum regexp to include digits

knqyf263:
SQLBoiler skips enum including digits (0-9) in a name (e.g. test_1).

However, enum including digits can be generated.

The following is the case of PostgreSQL.
```
testdb=# CREATE TYPE test_enum_1 AS ENUM ('test_1', 'test_2');
CREATE TYPE
```

Document: 
>SQL identifiers and key words must begin with a letter (a-z, but also letters with diacritical marks and non-Latin letters) or an underscore (_). Subsequent characters in an identifier or key word can be letters, underscores, digits (0-9), or dollar signs ($).
https://www.postgresql.org/docs/9.5/static/sql-syntax-lexical.html

Perhaps MySQL is the same.
https://dev.mysql.com/doc/refman/5.7/en/identifiers.html
>ASCII: [0-9,a-z,A-Z$_] (basic Latin letters, digits 0-9, dollar, underscore)

Thank you for the awesome tool!

### DIFF
--- a/strmangle/strmangle.go
+++ b/strmangle/strmangle.go
@@ -17,7 +17,7 @@ var (
 	idAlphabet    = []byte("abcdefghijklmnopqrstuvwxyz")
 	smartQuoteRgx = regexp.MustCompile(`^(?i)"?[a-z_][_a-z0-9]*"?(\."?[_a-z][_a-z0-9]*"?)*(\.\*)?$`)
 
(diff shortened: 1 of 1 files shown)


aarondl:
Merged via: eebe62f
Please open PR's against dev next time :)

knqyf263:
Oh, I made a mistake.  Sorry...

knqyf263:
I need this bug fix. 
It's helpful if this PR is merged into master early.
Thanks.

PR #235, opened 2017-12-22
Add Support for CockroachDB

baitxo:
We are planning to use SQLBoiler with CockroachDB and this PR adds support for it. 

CockroachDB does not support all data types yet so I've modified the test schema accordingly. 

Disclaimer: I'm pretty new to Go development ;)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,6 +18,9 @@ jobs:
 #         ACCEPT_EULA: 'Y'
 #         SA_PASSWORD: 'R@@tr@@t1234'
 
+      - image: cockroachdb/cockroach:v1.1.3
+        command: [start, --insecure]
+
     environment:
       GOPATH: /go
       ROOTPATH: /go/src/github.com/volatiletech/sqlboiler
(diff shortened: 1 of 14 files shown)


aarondl:
I wish I had known that this was happening. The v3 (https://github.com/volatiletech/sqlboiler/tree/v3) code rewrites a lot of the driver functionality and I would have told you to use that as your base (it still works just fine it's parity with v2 code) probably :( :( We'll either have to port it now or later which is extra work. But it's still an undetermined amount of time away from actually becoming the master branch. Thoughts?

baitxo:
I will check the v3 out to stimate how much work is needed to port it. I will let you know ;)

glerchundi:
Great work @baitxo!

> But it's still an undetermined amount of time away from actually becoming the master branch.

When you say undetermined we're talking about weeks or months? What about cutting alpha versions from v3/develop to start testing it out? In case that is an option and until it gets into beta/stable phase this PR could be developed against the march-expected (they are already pushing v2.0-alpha releases) cockroach version which supports JSON as well.

Putting this issue as a ref, which could lead to simplify templating code: https://github.com/cockroachdb/cockroach/issues/20985

WDYT?

aarondl:
v3 is an alpha branch. It's never going to be broken, as features are developed in isolation and then merged. You could use it today and it should work fine. If there are bugs in it I'd take a look at them.

baitxo:
Hi @aarondl,

Sorry for the delay, I've been unable to work on this... I've already checked the V3 and will port it.

I'm closing this PR in favor of the new one.

PR #236, opened 2018-01-04
Adjust randomize/randomize.go to conform with the changes in satori/go.uuid

mpecan:
satori/go.uuid changed the return signature of the New{version} function to return a value and error instead of just a single value.

The change is limited to the randomize file (as go.uuid is used nowhere else).

### DIFF
--- a/randomize/randomize.go
+++ b/randomize/randomize.go
@@ -195,7 +195,11 @@ func randomizeField(s *Seed, field reflect.Value, fieldType string, canBeNull bo
 					return nil
 				}
 				if fieldType == "uuid" {
-					value = null.NewString(uuid.NewV4().String(), true)
+					randomUuid, err := uuid.NewV4()
+					if err != nil {
+						return err
+					}
+					value = null.NewString(randomUuid.String(), true)
 					field.Set(reflect.ValueOf(value))
 					return nil
 				}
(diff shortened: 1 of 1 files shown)


mpecan:
The upstream task discussing the unannounced change can be found here: https://github.com/satori/go.uuid/issues/66

film42:
Thanks for beating me to the fix @mpecan ! LGTM

aarondl:
Kind of waiting to see where it goes. So far the maintainer of the package hasn't replied to any of the calls for reversion or anything. In the meantime, you'll just have to check out the older version of that package to build. Sorry :( My intention is to wait for him to reply (hopefully soon) and then we'll figure out the appropriate action to take. Since if he does revert, it'll mean another breaking change for `sqlboiler` users.

mpecan:
@aarondl I was expecting that, which is why I also linked the issue in the comment afterwards. We have to wait and see.
Since the tests do not run on CI this does not really break my workflow too much, so no worries.

glerchundi:
FYI https://github.com/satori/go.uuid/issues/48#issuecomment-356706335.

It seems like this PR applies then.

aarondl:
Thanks for the PR. Sorry for the delay. Please branch off dev in the future though.

PR #256, opened 2018-03-13
build: use make in dev & ci workflows 

glerchundi:


glerchundi:
PTAL

/cc @aarondl @nullbio

glerchundi:
Ok, after a conversation on Slack with @aarondl, we decided to make dev and ci workflow as similar as possible. This required a bunch of changes in the `Makefile` as well as in the CircleCI `config.yml` file.

I also moved `.PHONY`s near to the corresponding target following this guideline: http://clarkgrubb.com/makefile-style-guide#phony-targets so that no target will exists without its twin `.PHONY`.

It should be ready for a review pass.

testdata/Dockerfile
```diff
@@ -26,10 +26,10 @@ RUN curl https://www.postgresql.org/media/keys/ACCC4CF8.asc | apt-key add - \
     && apt-get update \
     && env ACCEPT_EULA=Y apt-get install -y git postgresql-client-9.6 mysql-client-5.7 mssql-tools unixodbc-dev
 
-# Install Go
-RUN curl -o $GODIST https://storage.googleapis.com/golang/$GODIST \
+# Install Go + Go based tooling
+ENV GOLANG_VERSION 1.10
+RUN curl -o go.tar.gz https://storage.googleapis.com/golang/go$GOLANG_VERSION.linux-amd64.tar.gz \
```
aarondl: As a matter of principle I think this line should be written as:
```
RUN curl -o go.tar.gz "https://storage.googleapis.com/golang/go${GOLANG_VERSION}.linux-amd64.tar.gz" \
```

I'm not a fan of undelimited replacements in shell scripts, because it leaves it up to wishy washy rules and it's somewhat unclear to read.

Makefile
```diff
@@ -22,126 +12,170 @@ DRIVER_USER=sqlboiler_driver_user
...
-	go test ./models
+.PHONY: test
+test:
+	go test -v -race $(go list ./... | grep -v /drivers/ | grep -v /vendor/)
 
+.PHONY: test-user-psql
 test-user-psql:
 	# Can't use createuser because it interactively promtps for a password
-	psql --command "create user $(USER) with superuser password '$(PASS)';" postgres
+	psql --host localhost --username postgres -c "create user $(USER) with superuser password '$(PASS)';"
```
aarondl: The expansion of `--command` here was intentional. In all scripts it's generally better to use the expanded form of an argument. The `-c` is to save you typing once you know it when executing via the command line, but when we only have to read it because it's in a script, it becomes better to have the full argument visible to the reader since quite often you won't require a trip to the manpage if you see the full name. (And yea, I'm aware that `-c` is a common "run command" argument in shells etc, I don't think that makes the argument weaker).

.circleci/config.yml
```diff
@@ -13,151 +13,121 @@ jobs:
...
       - run:
-          name: Add PSQL Creds
+          name: 'Make GOPATH'
+          command: mkdir -p $ROOTPATH
+
+      - checkout:
+          name: 'Checkout'
+          path: /go/src/github.com/volatiletech/sqlboiler
+
+      # Workaround to allow the use of the circleci local cli.
```
aarondl: What's the point of this when the checkout step has already completed above? Just curious.

.circleci/config.yml
```diff
@@ -3,7 +3,7 @@ jobs:
   build:
     working_directory: /root
     docker:
-      - image: aarondl0/sqlboiler-test:latest
+      - image: quay.io/glerchundi/sqlboiler-test:v3
```
aarondl: I've built and published this tag to `aarondl0/sqlboiler-test:v3`, please change to reflect this.

The old one will reside on aarondl0/sqlboiler-test:v2 (and it also pulls as latest since I pushed with a specific tag).

glerchundi: Ok.

.circleci/config.yml
glerchundi: It seems like circleci cli when running in a local environment checkouts (copies all the files) everything in `/root`.

```
[...]
====>> Checkout
  #!/bin/sh
mkdir -p /root && cp -r /tmp/_circleci_local_build_repo/. /root
[...]
```

So my strategy was to "in case checkout didn't work by checking if there isn't any file in there, use the local environment files"

But i think it would be a good idea to test it out directly in circleci once we merge this.

Makefile
glerchundi: Ok.

glerchundi:
> For example the way I currently run tests is not with the make file (since to me it's really good at set up, but poor at running the tests), but with ginkgo because despite my hatred of the framework side of it it happens to be a good test runner. Now I need to also add all the arguments in to iterate on the tests in my normal way. Same thing is true if I want color output on my tests (which I then pipe output through my own colortest program). In general the complete reliance on the makefile to get these parameters right instead of the previous hardcoding is that it hurts our ability to run them easily outside the make file. Food for thought while you fix these other issues, please comment on this.

Personally i like `make` as the entrypoint for everything and, without any numbers at hand, I think it is the default choice of most Go developers (for example, [`dep`](https://github.com/golang/dep), [`kubernetes`](https://github.com/kubernetes/kubernetes), [`docker/moby`](https://github.com/moby/moby), [`cockroachdb`](https://github.com/cockroachdb/cockroach), [`contour`](https://github.com/heptio/contour), ...). So, without reliable evidences, I think we'll not have any problem from people (and us/you too! ;) getting started with the current implementation.

That being said, I think we can do something to solve your use case by doing something like:

```
source testdata/*.env
ginkgo
```

By inferring username/password/database from environment variables.

I can address it in a further PR, WDYT?

aarondl:
Yep. That was my thinking as well since env doesn't require any typing. The makefile hardly changes in terms of command invocation.

I've opened this issue since it sounds like you'd like to defer doing the work.
https://github.com/volatiletech/sqlboiler/issues/257

glerchundi:
Perfect, thanks @aarondl.

Will work on two new PRs, one for integrating CockroachDB and the other to solve #256, have a good weekend!

PR #258, opened 2018-03-19
build: make use of envs instead of args

### LINKED ISSUE #257: v3 Code should use env not arguments to fuel test binaries
Currently the driver tests all accept arguments as a means of configuring what credentials they will use to access a database. This is undesirable because it makes invoking them via the command line (ie. through `go test` or some other test runner) with your own arguments obnoxious since you have to provide the credentials on the command line.

Instead we should replace those command line arguments with environment variables, and put those environment variables in a file under `testdata`, the `Makefile` will need to be adjusted to use these environment variables as well, and then a choice needs to be made:

Do we ask users to source the file before running makefile pieces, and bail in the makefile if these vars are not present? This seems undesirable.
Or do we source before every invocation time? This is going to ugly up the makefile something awful. And we'd have to run everything via `sh`/`bash` to be able to `source`.
Or do we put the values in two places again? This is probably the correct answer, since they'll never change. They were previously hardcoded in two places after all.

glerchundi:
Fixes #257

Changes:
* When it comes to configuring the driver testing environment, flags are no longer used. This was almost (except `overwrite-golden`) entirely replaced by environment variables.
* A new target called `testdata/env.mk` is automatically executed before any target thus not polluting other targets with useless dependencies to that one.
* `SQLBOILER_TEST_` was used as a prefix in order to avoid environment name clashes.
* Now people (like @aarondl ;) can use other testing tools by doing `source testdata/env.sh` first and just once.
* Used `*_PASS` for any driver by converging `MSSQLPASS` and `PASS` and using the most restrictive password requirements.

Signed-off-by: Gorka Lerchundi Osa <glertxundi@gmail.com>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ sqlboiler.toml
 models/
 testschema.sql
 .cover
+testdata/env.mk
--- a/Makefile
+++ b/Makefile
@@ -1,15 +1,14 @@
 # Everything in this make file assumes you have the ability
 # to run the commands as they are written (meaning .my.cnf and .pgpass
 # files set up with admin users) as well as the mssql db set up with
-# the sa / Sqlboiler@1234 credentials (see docker run below for example)
+# the sa / Sqlboiler@1234 credentials. See testdata/env.sh.
 
-USER=sqlboiler_root_user
-DB=sqlboiler_model_test
-PASS=sqlboiler
-MSSQLPASS=Sqlboiler@1234
+-include testdata/env.mk
 
-DRIVER_USER=sqlboiler_driver_user
-DRIVER_DB=sqlboiler_driver_test
+# Automatically import test related environments
+.PHONY: testdata/env.mk
+testdata/env.mk: testdata/env.sh
+	sed 's/"//g ; s/=/:=/' < $< > $@
 
 # Builds all software and runs model tests
 .PHONY: tests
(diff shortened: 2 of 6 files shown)


glerchundi:
PTAL @aarondl.

aarondl:
While I'm staring at this I realize that I've lost compatibility with my own shell with this PR. One thing that stood out to me is the include hack. Since we're now dependent on bash for setting environment variables in the `.sh` file. Since we have a dependency on `sed` and `sh` now I think the best idea is to abandon make for a straight-up shell script. I'm not super happy about this obviously but at least we'd be able to provide `--help` on the script so people can see the targets, and we can all use `source` like normal human beings to deal with this. I want to hear your thoughts about this, but I could help out here depending on what we decide.

glerchundi:
@aarondl My opinion on this is that asking for bash & makefile to _build_ SQLBoiler is not that much. When I work on windows and I see myself in need of linux/bash i usually use [WLS](https://en.wikipedia.org/wiki/Windows_Subsystem_for_Linux) or Docker. Not to mention that the CIs we're already using are Linux based which eases CI testing as well (`circleci build`).

aarondl:
Closing this in favor of d53a22c.

The reasoning is in the commit message. I don't care about the duplication so much as I care about convenience it turns out. The shell script is also more useful than the make file is as it allows you to pass arguments along if you need to. Hopefully you don't disagree. I just didn't want to deal with make include and regexes to replace things and having to source files and the possible inclusion of direnv later because of this, it's just not worth it.

glerchundi:
I would like to take a deeper look but it look good to me!

PR #260, opened 2018-03-26
*: initial crdb support

glerchundi:


glerchundi:
PTAL @aarondl

glerchundi:
@aarondl any chance to look at this?

Now that CockroachDB [has launched the new version](https://www.cockroachlabs.com/blog/cockroachdb-2-0-release/) seems like a good moment to push this forward ;)

importers/imports.go
```diff
@@ -308,6 +308,24 @@ func NewDefaultImports() Collection {
...
+				`"fmt"`,
+				`"io"`,
+				`"os"`,
+				`"os/exec"`,
+				`"strings"`,
+			},
+			ThirdParty: List{
+				`"github.com/pkg/errors"`,
+				`"github.com/spf13/viper"`,
+				`"github.com/volatiletech/sqlboiler/drivers/sqlboiler-crdb/driver"`,
```
aarondl: We may have to figure out a way around this when we go to separate drivers out. sqlite3 would have the same issue.

main.go
```diff
@@ -175,6 +175,11 @@ func preRun(cmd *cobra.Command, args []string) error {
 		viper.SetDefault("mssql.sslmode", "true")
 		viper.SetDefault("mssql.port", 1433)
 		required = append(required, "schema")
+	case "crdb":
```
aarondl: Can we have these directly in the driver? As the above TODO states, need to move them into the drivers anyway at some point.

aarondl: It'll also affect breaking this one out.

queries/query_builders.go
```diff
@@ -327,6 +327,58 @@ func BuildUpsertQueryMSSQL(dia drivers.Dialect, tableName string, primary, updat
 	return buf.String()
 }
 
+// BuildUpsertQueryCockroachDB builds a SQL statement string using the upsertData provided.
+func BuildUpsertQueryCockroachDB(dia drivers.Dialect, tableName string, updateOnConflict bool, ret, update, conflict, whitelist []string) string {
```
aarondl: We need to come up with a mechanism to break this out. I think it should be simple, each driver should be able to create its own
upsert query. And we should just provide all the information to it that we can, and then it can make the decision what to do with it.

It involves figuring out a superset of all the required info then executing the driver's method.

templates/17_upsert.tpl
```diff
@@ -39,7 +39,7 @@ func (o *{{$tableNameSingular}}) Upsert(exec boil.Executor, {{if eq .DriverName
 
 	// Build cache key in-line uglily - mysql vs psql problems
 	buf := strmangle.GetBuffer()
-	{{if eq .DriverName "psql"}}
+	{{if (or (eq .DriverName "psql") (eq .DriverName "crdb"))}}
```
aarondl: Ignoring the upsert portions of this stuff, these conditionals are really hard for out of band drivers. Since we do have a lot of conditional logic that says "if postgres" or "if mysql". We'll probably have to do some additional work to clean these up into more abstract pieces or the entire system will suffer a bit.

templates_test/main_test/crdb_main.tpl
```diff
@@ -0,0 +1,130 @@
+type pgTester struct {
```
aarondl: I notice this file is formatted very strangely, not like Go code (no tabs for example). Might be worth it to run gofmt on it once.

templates_test/singleton/boil_queries_test.tpl
```diff
@@ -10,6 +10,7 @@ func MustTx(transactor boil.Transactor, err error) boil.Transactor {
 var rgxPGFkey = regexp.MustCompile(`(?m)^ALTER TABLE ONLY .*\n\s+ADD CONSTRAINT .*? FOREIGN KEY .*?;\n`)
 var rgxMySQLkey = regexp.MustCompile(`(?m)((,\n)?\s+CONSTRAINT.*?FOREIGN KEY.*?\n)+`)
 var rgxMSSQLkey = regexp.MustCompile(`(?m)^ALTER TABLE .*ADD\s+CONSTRAINT .* FOREIGN KEY.*?.*\n?REFERENCES.*`)
+var rgxCDBFkey = regexp.MustCompile(`(?m)((\n)?\s+CONSTRAINT.*?FOREIGN KEY.*?\n)+`)
```
aarondl: Another part for driver separation.

types/array.go
```diff
@@ -519,13 +519,13 @@ func (a *BoolArray) scanBytes(src []byte) error {
 	} else {
 		b := make(BoolArray, len(elems))
 		for i, v := range elems {
-			if len(v) != 1 {
+			if len(v) < 1 {
```
aarondl: These fixes seem generic, should they make their way into `master` earlier than the v3 code?

glerchundi: Yep, will do.

templates_test/main_test/crdb_main.tpl
glerchundi: Umm..., strange. Will take a look.

glerchundi:
As we talked yesterday or the day before yesterday (i can't remember 😄), it seems like we're no fully comfortable with the driver splitting and hence with integrating new drivers/dialects that will suffer from the same splitting issues.

@aarondl if you don't mind, I would prefer to merge this (once you feel it's mergeable of course) and then do further PRs to improve the driver splitting itself. Also, as we mentioned, once we feel that splitting is good enough it would be a good idea to test it out by creating another driver in a fully separated repository and see what happens. Obviously the first candidate would be SQLite which requires fuckin' cgo...

What do you think? I'm open to work on those further PRs.

aarondl:
@glerchundi Sorry for the insane slowness. Gonna be working on this stuff a lot as of tomorrow. Just got through the backlog of bugfixes and PR's today. Would like to re-engage here if you're good to.

So the driver split stuff is obviously a bit wonky, we can agree there. But I'd much rather us work at that than merge this in and later attempt to split it out again. That will force us to fix the issues with the driver splitting. So we get two good outcomes and the path to sqlite3 becomes easier as well. I will provide support for the driver split fixes, if you could do me the favor and start attempting to use it.

Driver split issues I'll be taking care of:
https://github.com/volatiletech/sqlboiler/issues/279
https://github.com/volatiletech/sqlboiler/issues/280
https://github.com/volatiletech/sqlboiler/issues/281
https://github.com/volatiletech/sqlboiler/issues/282
https://github.com/volatiletech/sqlboiler/issues/283

Also a few of my comments were not addressed by you yet, comment about the `!=` vs `<`, the file not being go-formatted.

glerchundi:
> @glerchundi Sorry for the insane slowness. Gonna be working on this stuff a lot as of tomorrow. Just got through the backlog of bugfixes and PR's today. Would like to re-engage here if you're good to.

Don't worry, I'm really busy at work ATM.

I'll try to revisit everything you did in these days and tell you something about the splitting issues.

glerchundi:
FANTASTIC JOB @aarondl! I'm planning to work on this next week, will point here with my progress.

glerchundi:
@aarondl i finally converted from the PR to an actual driver, you can take a look to it here: [glerchundi/sqlboiler-crdb](https://github.com/glerchundi/sqlboiler-crdb).

Everything looks promising but I had to make some changes before it finally worked:
* I modified the logic to lookup the binary.
* Boolean array values can now be defined as `T,t,F,f,False,false,True,true`.

Also, I discovered that SQLBoiler "Inserts" are not working properly when the table has nothing more than a primary key, it fails with a `models: unable to insert into users: sql: no rows in result set`. I don't know why or how but an insert without a returning statement is getting inside this piece of code:

```
[...]
	if len(cache.retMapping) != 0 {
		err = exec.QueryRow(cache.query, vals...).Scan(queries.PtrsFromMapping(value, cache.retMapping)...)
	}
[...]
```

I can send further PR's to make things work as I explained here but I want to double check before spending some time on this.

Thanks!

#### main.go
```
[...]
	driverName := args[0]
	driverPath, err := exec.LookPath(fmt.Sprintf("sqlboiler-%s", driverName))
	if err != nil {
		return errors.Wrap(err, "could not find absolute path to driver")
	}

	drivers.RegisterBinary(driverName, driverPath)
	fmt.Println("Will be executing the binary at: ", driverPath)
[...]
```

#### types/array.go
```
[...]
			if len(v) < 1 {
				return fmt.Errorf("boil: could not parse boolean array index %d: invalid boolean %q", i, v)
			}
			switch v[0:1][0] {
			case 't', 'T':
				b[i] = true
			case 'f', 'F':
				b[i] = false
[...]
```

aarondl:
Finding a driver on `PATH` is nice. Currently it handles two cases: A binary given by full path, a short name of a driver that's in the current dir.

I think we shouldn't lose these two abilities, but instead keep all three. Ideally it'd have three ways to find the binary:
* Full path specified (os.Separator present), just use it directly and truncate the `sqlboiler-` prefix as it does now.
* Otherwise assume short name specified in which case it'll check current directory and `exec.LookPath` for `sqlboiler-{driverName}`.
* The println should only happen if debug has been turned on IMHO. But we have to be careful since the debug output is all JSON I think atm :( Maybe we could print that to stderr.

Unconcerned about the changes to types/array.go, I'd be shocked if there was any actual problem because of that across any database :p

In terms of inserts not working did you need me to try to dig into that? Let's start by at least filing an issue so we can track it and make sure it gets fixed appropriately.

glerchundi:
> inding a driver on PATH is nice. Currently it handles two cases: A binary given by full path, a short name of a driver that's in the current dir.

AFAIK it does handle just the first one not the second. Current dir is checked if PATH contains `.` which is not a very common and not a very good security practice unless you're the only guy in that machine.

> I think we shouldn't lose these two abilities, but instead keep all three. Ideally it'd have three ways to find the binary:

I think that less is more and having just one logic to lookup the binary is enough, more considering that every usecase I can imagine of can be solved. Take `protoc` as an example, it follows a pre-established naming convention like we're doing here (`protoc --go_out=` will look for a `protoc-gen-go`) and lookups the corresponding plugin/binary in PATH.

> The println should only happen if debug has been turned on IMHO. But we have to be careful since the debug output is all JSON I think atm :( Maybe we could print that to stderr.

Right, will send a PR and lets discuss that there!

> Unconcerned about the changes to types/array.go, I'd be shocked if there was any actual problem because of that across any database :p

Actually, there is! CockroachDB is encoding array values like this: `{false,true,false}`!

> In terms of inserts not working did you need me to try to dig into that? Let's start by at least filing an issue so we can track it and make sure it gets fixed appropriately.

Will file an issue and lets see if something can/should be done.

Thanks 👍

Issue #259, opened 2018-03-20
Geometric types for Postgres


saulortega:
It was not easy to try it in v3. It's very broken! ("postgres" changes to "psql", missing ALL imports in main_test.go, and so on...)

But in the end I succeeded. :)

Well, before doing PR, I want to show here the changes that would be added to the code, and to know your opinions. For now, I did tests with the point type. These same changes should be made for the other geometric types.

OK, the point type: https://github.com/saulortega/pgeo/blob/master/point.go

randomize/randomize.go:
```go
import (
        //...
	"github.com/volatiletech/sqlboilerv3/strmangle"
	"github.com/volatiletech/sqlboilerv3/types"
	"github.com/saulortega/pgeo" // -------------- New import added --------------
)

//...

var (
        //...
	typeStringArray  = reflect.TypeOf(types.StringArray{})
	typeHStore       = reflect.TypeOf(types.HStore{})
	typePoint        = reflect.TypeOf(pgeo.Point{}) // -------------- New added --------------
)

//...

func randomizeField(s *Seed, field reflect.Value, fieldType string, canBeNull bool) error {
//...
			case typeHStore:
				value := types.HStore{}
				value[randStr(s, 3)] = sql.NullString{String: randStr(s, 3), Valid: s.nextInt()%3 == 0}
				value[randStr(s, 3)] = sql.NullString{String: randStr(s, 3), Valid: s.nextInt()%3 == 0}
				field.Set(reflect.ValueOf(value))
				return nil
				// --------------------- BEGIN ADDED LINES (none random yet, really :) ) -----------------
			case typePoint:
				value = pgeo.Point{}
				field.Set(reflect.ValueOf(value))
				return nil
				// ---------------------- END ADDED LINES -------------------
			}
//...
}

```

importers/imports.go:
```go
func NewDefaultImports() Collection {
		//...
		"types.Hstore": {
			ThirdParty: List{`"github.com/volatiletech/sqlboilerv3/types"`},
		},
		// --------------------- BEGIN ADDED LINES -----------------
		"pgeo.Point": {
			ThirdParty: List{`"github.com/saulortega/pgeo"`},
		},
		// ---------------------- END ADDED LINES -------------------
		//...
}
```

drivers/sqlboiler-psql/driver/psql.go:
```go
func (p *PostgresDriver) TranslateColumnType(c drivers.Column) drivers.Column {
		//...
		case "date", "time", "timestamp without time zone", "timestamp with time zone":
[...]

Issue #143, opened 2017-04-29
External Driver Support


aarondl:
Use JSON as a data-passing mechanism to other binaries.

These binaries will be able to receive JSON configuration, talk to a database, return the set of tables and other data.
- Keep in mind the imports for custom data types
- Make sure that this is possible to use as a library with side effect imported packages

aarondl:
Done in v3 branch.

PR #263, opened 2018-03-28
Add geometric types for Postgres

saulortega:
Well here it is.

Reference. https://github.com/volatiletech/sqlboiler/issues/259

### DIFF
--- a/drivers/sqlboiler-psql/driver/psql.go
+++ b/drivers/sqlboiler-psql/driver/psql.go
@@ -362,8 +362,22 @@ func (p *PostgresDriver) TranslateColumnType(c drivers.Column) drivers.Column {
 			c.Type = "null.JSON"
 		case "boolean":
 			c.Type = "null.Bool"
-		case "date", "time", "timestamp without time zone", "timestamp with time zone":
+		case "date", "time", "timestamp without time zone", "timestamp with time zone", "time without time zone", "time with time zone":
 			c.Type = "null.Time"
+		case "point":
+			c.Type = "pgeo.NullPoint"
+		case "line":
+			c.Type = "pgeo.NullLine"
+		case "lseg":
+			c.Type = "pgeo.NullLseg"
+		case "box":
+			c.Type = "pgeo.NullBox"
+		case "path":
+			c.Type = "pgeo.NullPath"
+		case "polygon":
+			c.Type = "pgeo.NullPolygon"
+		case "circle":
+			c.Type = "pgeo.NullCircle"
 		case "ARRAY":
 			if c.ArrType == nil {
 				panic("unable to get postgres ARRAY underlying type")
@@ -404,8 +418,22 @@ func (p *PostgresDriver) TranslateColumnType(c drivers.Column) drivers.Column {
 			c.Type = "[]byte"
 		case "boolean":
 			c.Type = "bool"
-		case "date", "time", "timestamp without time zone", "timestamp with time zone":
+		case "date", "time", "timestamp without time zone", "timestamp with time zone", "time without time zone", "time with time zone":
 			c.Type = "time.Time"
+		case "point":
+			c.Type = "pgeo.Point"
+		case "line":
+			c.Type = "pgeo.Line"
+		case "lseg":
+			c.Type = "pgeo.Lseg"
+		case "box":
+			c.Type = "pgeo.Box"
+		case "path":
+			c.Type = "pgeo.Path"
+		case "polygon":
+			c.Type = "pgeo.Polygon"
+		case "circle":
+			c.Type = "pgeo.Circle"
 		case "ARRAY":
 			c.Type = getArrayType(c)
 			// Make DBType something like ARRAYinteger for parsing with randomize.Struct
--- a/importers/imports.go
+++ b/importers/imports.go
@@ -389,6 +389,48 @@ func NewDefaultImports() Collection {
 		"types.Hstore": {
 			ThirdParty: List{`"github.com/volatiletech/sqlboiler/types"`},
 		},
+		"pgeo.Point": {
+			ThirdParty: List{`"github.com/volatiletech/sqlboiler/types/pgeo"`},
+		},
+		"pgeo.Line": {
+			ThirdParty: List{`"github.com/volatiletech/sqlboiler/types/pgeo"`},
+		},
+		"pgeo.Lseg": {
+			ThirdParty: List{`"github.com/volatiletech/sqlboiler/types/pgeo"`},
+		},
+		"pgeo.Box": {
+			ThirdParty: List{`"github.com/volatiletech/sqlboiler/types/pgeo"`},
+		},
+		"pgeo.Path": {
+			ThirdParty: List{`"github.com/volatiletech/sqlboiler/types/pgeo"`},
+		},
+		"pgeo.Polygon": {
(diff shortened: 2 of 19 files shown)


aarondl:
Similar to another issue here #268 where we have extensions being supported in the core though, I wonder what we should be doing in terms of support for this. Since it's not completely a part of postgres, should it be part of sqlboiler directly? Maybe it makes sense to have a forked psql driver for this. It's just a conversation I'd like to have, I'm not saying we won't merge these.

saulortega:
I did tests in my database with my fork. I do not know if you mean some code to do type tests. How is that?

These typres are part of postgres (no extension): https://www.postgresql.org/docs/current/static/datatype.html
Even so, I would also prefer some kind of external implementation, because there can be several ways to implement these types (i.e. They can be simple things such as to name the coordinates of a point X/Y or Longitude/Latitude, but there can also be more important differences).

I'm not sure if this is already implemented in V3. You mentioned something about importing custom data types in https://github.com/volatiletech/sqlboiler/issues/143 . Is it possible in V3 to implement an external library for any type? If so, I would like to do it that way.

aarondl:
There is code here to do type tests:
https://github.com/volatiletech/sqlboiler/blob/master/testdata/postgres_test_schema.sql

It ensures that these types generate and run tests.

As for an external implementation, the splitting of the drivers has been completed in the v3 branch. As well as custom imports being possible now too. So that means this work could be done in a fork of the postgres driver code (from v3 as a base). The main problem I foresee is the tests. Since the randomize.go has all these types caked into it, I wonder if we could actually create a new interface for types that want to be randomized for tests.

saulortega:
I'm not sure if I understand you.

I don't like the idea of custom driver forks. I like this: https://github.com/volatiletech/sqlboiler/issues/100#issuecomment-276423951

So, that toml config is done and working in V3 without the need of a fork driver? If not, what is the custom imports to which you refer?

(sorry if mi English is bad)

aarondl:
The toml config is *not* done to be clear. We didn't go that direction we went the direction of having the drivers be external binaries so they can be replaced by literally anything.

Here's an example of the postgres driver code as it is today in v3, this ends up as a separate binary that speaks JSON:
https://github.com/volatiletech/sqlboiler/blob/v3/drivers/sqlboiler-psql/main.go
It still uses a LOT of help from the sqlboiler `drivers` package, but the nice part is that it doesn't need to. This is easily forkable. So we could have a separate repository that's called `sqlboiler-psql-gis` if we wanted that contains code to also do the gis extension (I don't know if we want to have this, it's just a thought at the moment).

In terms of the custom imports, the code is here for that:
https://github.com/volatiletech/sqlboiler/blob/v3/main.go#L222
What this does is allow users to specify imports without having to change any code inside `sqlboiler`. They can replace the driver, and specify their own type imports to supplement the driver and ensure that the correct imports get added into the generated code to use any extra libraries the driver wants to, like this:

```toml
[[imports.all.based_on_type."null.Decimal"]]
third_party = ["\"github.com/ericlagergen/decimal\""]
```

I generally agree less forks is better though remember that everything that lands in the sqlboiler `types` package burdens `mysql`, `mssql`, `sqlite3`, and `cockroachdb` users as well, but maybe not as badly (it's just dead code) because you've put things in `types/pgeo` which was a good decision.

So I have one pointed question:
When you wrote your tests, did the test database have to be created with a line that looked like this:

```sql
create extension if not exists ...
```

If so, then that's not part of postgres core. If not, then this conversation is done and I can look at merging this!

saulortega:
Hello @aarondl.

I have not done any tests yet, but I'm looking at the code for custom imports and there's something I do not understand: I do not see a way to map the database type to the Go type.

There should be something like this, I think:

```json
{
    "db_type": "decimal",
    "go_type": "decimal.Decimal",
    "third_party_library": "\"github.com/ericlagergen/decimal\"",
}
```

ceshihao:
It is seems there is no such a map.
There is a mapping relationship in this driver function.

https://github.com/volatiletech/sqlboiler/blob/06841eda617526f93b198243d39bcd8774e39c33/drivers/sqlboiler-psql/driver/psql.go#L339-L342

saulortega:
That means that the only thing I can do is replace third-party libraries, like null, with others. That is not very useful.

Aaron, would you accept a PR that adds this DB-to-Go mapping functionality (using viper)?

aarondl:
Hey @saulortega, sorry for the insane slowness on this. The extension business has been cleared up on the other issue as well, and this one looks clean so we can go ahead with this implementation.

In terms of how we MIGHT have implemented it, people use a forked driver that contains the TranslateColumnType function that has the extra types, and then they additionally use a config file that helps do the custom imports. It's a bit ugly actually, I'm thinking maybe the imports need to be part of the driver now.

aarondl:
I'm going to merge this so we can move forward with it. Do you mind adding some tests though? The type tests are in https://github.com/volatiletech/sqlboiler/blob/v3/testdata/psql_test_schema.sql

aarondl:
One more thing @saulortega, if you could squash your commits (down to at most 3, just want all those extraneous one-file creation commits gone) that'd be great. Will merge IMMEDIATELY after you do so.

saulortega:
Hello.

At https://github.com/volatiletech/sqlboiler/blob/master/testdata/postgres_test_schema.sql you already have all these seven geometric types (magic table). Do I have to add something else?

I did squash (ceb8d08), but the previous commits still appear. I do not know if this is normal or I need to do something else.

aarondl:
Hey Saul. The history has gotten quite messy now because this branch contains two versions of the same changes from what I can see.

```
*   3593f27 - (saulortega/v3) Merge remote-tracking branch 'origin/v3' into v3 (31 minutes ago)
|\  
| * 4c84977 - Create polygon.go (7 weeks ago)
| * 6d9bce2 - Create point.go (7 weeks ago)
... ELIDED FOR BREVITY ...
| * 7f93905 - Added geotypes for Postgres (7 weeks ago)
* | ceb8d08 - Added geotypes for Postgres (54 minutes ago) # THIS IS THE ONLY COMMIT WE WANT
|/  
*   fce9cf3 - Merge branch 'glerchundi/v3' into v3 (9 weeks ago)
|\  
| * 502a18b - build: use make in dev & ci workflows (9 weeks ago)
|/  
* e327cdf - (origin/split-drivers, split-drivers, fix-mysql-upsert, fix-all-upserts) Add a makefile. (6 months ago)
```

I think the proper thing to do here is to create a new branch with the one commit that adds everything that's on top of the current v3.
```bash
git fetch -p
git checkout -b pgis-types origin/v3 # Assuming origin means volatiletech/sqlboiler
git cherry-pick ceb8d08
git push -u saulortega pgis-types
```

And then open up a pull request with that new branch. It should do the trick, and leaves your current branch alone just in case so we don't lose anything :)

aarondl:
Actually I realized since you squashed into a commit, I could do it myself!

```text
*   b37d51b - (HEAD -> v3, origin/v3) Merge branch 'pgis-types' into v3 (66 seconds ago)
|\  
| * f425832 - (pgis-types) Added geotypes for Postgres (2 minutes ago)
|/  
* 9ecfa11 - Support psql citext (3 days ago)
```

Please verify that things look like they should in the v3 branch. Thanks very much for this Saul.

PR #266, opened 2018-04-04
Ignore virtual columns

Nykakin:
In MySQL you can create virtual columns. You can't set them directly since they are generated on the fly:
```sql
mysql root@localhost:test> CREATE TABLE t (
                        ->   a DOUBLE,
                        ->   b DOUBLE,
                        ->   c DOUBLE AS (a + b)
                        -> );
Query OK, 0 rows affected
Time: 0.022s
mysql root@localhost:test> DESCRIBE t;
+---------+--------+--------+-------+-----------+-------------------+
| Field   | Type   | Null   | Key   |   Default | Extra             |
|---------+--------+--------+-------+-----------+-------------------|
| a       | double | YES    |       |    <null> |                   |
| b       | double | YES    |       |    <null> |                   |
| c       | double | YES    |       |    <null> | VIRTUAL GENERATED |
+---------+--------+--------+-------+-----------+-------------------+
3 rows in set
Time: 0.001s
mysql root@localhost:test> INSERT INTO t VALUES(1, 5, 6)
(3105, "The value specified for generated column 'c' in table 't' is not allowed.")
```
To insert data in such a table you have to omit these virtual columns:
```sql
mysql root@localhost:test> INSERT INTO t(a, b) VALUES(1, 5)
Query OK, 1 row affected
Time: 0.010s
mysql root@localhost:test> SELECT * FROM t
+-----+-----+-----+
|   a |   b |   c |
|-----+-----+-----|
|   1 |   5 |   6 |
+-----+-----+-----+
1 row in set
Time: 0.002s
```
This caused me problems when working with sqlboiler because it treated virtual columns as normal ones and generated boilerplate code for them as well. As a result in complex queries I got same error as presented above: `The value specified for generated column 'c' in table 't' is not allowed.`.

I've add a flag to ignore these as a simple workaround of this problem. I guess the correct approach would be to alter generated insert/set methods to make them ignore virtual columns but I don't want to stall the project I'm working on with too much changes in the underlying libs. For now I'd be happy to just omit these.

### DIFF
--- a/bdb/drivers/mysql.go
+++ b/bdb/drivers/mysql.go
@@ -148,7 +148,7 @@ func (m *MySQLDriver) Columns(schema, tableName string) ([]bdb.Column, error) {
 				(select count(*) from information_schema.key_column_usage where table_schema = kcu.table_schema and table_name = tc.table_name and constraint_name = tc.constraint_name) = 1
 		) as is_unique
 	from information_schema.columns as c
-	where table_name = ? and table_schema = ?;
+	where table_name = ? and table_schema = ? and c.extra not like '%VIRTUAL%';
 	`, tableName, schema)
 
 	if err != nil {
--- a/bdb/drivers/mysql.go
+++ b/bdb/drivers/mysql.go
@@ -17,6 +17,12 @@ import (
 // then tinyint(1) will be mapped in your generated structs to bool opposed to int8.
 var TinyintAsBool bool
 
+// IgnoreVirtualColumns is a global that is set from main.go if a user specifies
+// this flag when generating. This flag only applies to MySQL so we're using a global
+// instead, to avoid breaking the interface. If IgnoreVirtualColumns is true then
+// virtual columns present in the schema will be ignored during the code generation
+var IgnoreVirtualColumns bool
+
 // MySQLDriver holds the database connection string and a handle
 // to the database connection.
 type MySQLDriver struct {
@@ -148,8 +154,8 @@ func (m *MySQLDriver) Columns(schema, tableName string) ([]bdb.Column, error) {
 				(select count(*) from information_schema.key_column_usage where table_schema = kcu.table_schema and table_name = tc.table_name and constraint_name = tc.constraint_name) = 1
 		) as is_unique
 	from information_schema.columns as c
-	where table_name = ? and table_schema = ?;
-	`, tableName, schema)
+	where table_name = ? and table_schema = ? and (? or c.extra != 'VIRTUAL GENERATED');
+	`, tableName, schema, IgnoreVirtualColumns)
 
 	if err != nil {
 		return nil, err
--- a/bdb/drivers/mysql.go
+++ b/bdb/drivers/mysql.go
@@ -17,12 +17,6 @@ import (
 // then tinyint(1) will be mapped in your generated structs to bool opposed to int8.
 var TinyintAsBool bool
 
-// IgnoreVirtualColumns is a global that is set from main.go if a user specifies
-// this flag when generating. This flag only applies to MySQL so we're using a global
-// instead, to avoid breaking the interface. If IgnoreVirtualColumns is true then
-// virtual columns present in the schema will be ignored during the code generation
-var IgnoreVirtualColumns bool
-
 // MySQLDriver holds the database connection string and a handle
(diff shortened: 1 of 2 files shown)


aarondl:
Hm. So I've done a bunch of work in the v3 branch to eliminate the need for driver-specific baked in flags like this. I'd much rather just disable these entirely, and optionally bring them back in v3. What do you think about that?

Nykakin:
I don't really need the flag itself, I only added it to ensure backward compatibility. All I want is to prevent sqlboiler from trying to insert values into virtual columns. Can we actually settle on 

```go
	rows, err := m.conn.Query(`

        ...

	from information_schema.columns as c
	where table_name = ? and table_schema = ? and c.extra != 'VIRTUAL GENERATED';
	`, tableName, schema)
```
in both master and v3 branches?

aarondl:
Sounds fine to me. Though I worry about binding when there's a `select *`. Not sure if that'll work correctly, but otherwise I'm all for it.

Nykakin:
I've made some tests on following schema:
```sql
CREATE TABLE foo (
    id INT(6) UNSIGNED AUTO_INCREMENT PRIMARY KEY,
    foo_field CHAR(8) DEFAULT '',
    double_foo_field CHAR(16) as (concat(foo_field, foo_field))
);

CREATE TABLE bar (
    id INT(6) UNSIGNED AUTO_INCREMENT PRIMARY KEY,
    bar_field CHAR(8) DEFAULT '',
    double_bar_field CHAR(16) as (concat(bar_field, bar_field))
);

CREATE TABLE foo_bar (
    foo_id INT(6) UNSIGNED,
    bar_id INT(6) UNSIGNED,
    PRIMARY KEY(foo_id, bar_id),
    FOREIGN KEY(foo_id) REFERENCES foo(id),
    FOREIGN KEY(bar_id) REFERENCES bar(id)
);
```

Using following code:

```go
package main

import (
    "database/sql"
    "fmt"

    "./models"

    "github.com/volatiletech/sqlboiler/boil"
    "github.com/volatiletech/sqlboiler/queries/qm"
    _ "github.com/go-sql-driver/mysql"
	"gopkg.in/volatiletech/null.v6"
)

func main() {
    db, err := sql.Open("mysql", "root:1234@/test")
    if err != nil {
        panic(err)
    }
    boil.SetDB(db)

    bar1 := models.Bar {
        BarField: null.StringFrom("test1"),
    }
    bar2 := models.Bar {
        BarField: null.StringFrom("test1"),
    }
    foo := models.Foo {
        FooField: null.StringFrom("foo"),
    }

    if err := foo.InsertG(); err != nil {
        panic(err)
    }
    if err := bar1.InsertG(); err != nil {
        panic(err)
    }
    if err := bar2.InsertG(); err != nil {
        panic(err)
    }
    if err := foo.SetBarsG(false, &bar1, &bar2); err != nil {
        panic(err)
    }

    bars, err := models.BarsG(qm.Select("*")).All()
    if err != nil {
        panic(err)
    }
    for _, bar := range bars {
        fmt.Printf("%+v\n", bar)
    }
}
```
The code above fails in MySQL if virtual columns are not omitted (`Error 3105: The value specified for generated column 'double_foo_field' in table 'foo' is not allowed`). With patch applied it seems to work correctly. Also it doesn't seem to be any problem with using `qm.Select("*")`.

On MariaDB both patched and non patched code works fine.

bdb/drivers/mysql.go
```diff
@@ -148,7 +148,7 @@ func (m *MySQLDriver) Columns(schema, tableName string) ([]bdb.Column, error) {
 				(select count(*) from information_schema.key_column_usage where table_schema = kcu.table_schema and table_name = tc.table_name and constraint_name = tc.constraint_name) = 1
 		) as is_unique
 	from information_schema.columns as c
-	where table_name = ? and table_schema = ?;
+	where table_name = ? and table_schema = ? and c.extra not like '%VIRTUAL%';
```
mickeyreiss, line 151: @aarondl This fix does not seem to work for `GENERATED ALWAYS … STORED` columns, which are supported in MySQL[1] and do not include the keyword `VIRTUAL` in the column type.

1: https://dev.mysql.com/doc/refman/5.7/en/create-table-generated-columns.html

PR #268, opened 2018-04-06
Support PostgreSQL citext fields

boxofrad:
Postgres has a [citext type](https://www.postgresql.org/docs/9.1/static/citext.html) for storing textual data that you often want to perform case-insensitive operations on (e.g. usernames). It can be treated as a normal string.

Have at it! 💝

### DIFF
--- a/bdb/drivers/postgres.go
+++ b/bdb/drivers/postgres.go
@@ -348,10 +348,13 @@ func (p *PostgresDriver) TranslateColumnType(c bdb.Column) bdb.Column {
 			// Make DBType something like ARRAYinteger for parsing with randomize.Struct
 			c.DBType = c.DBType + *c.ArrType
 		case "USER-DEFINED":
-			if c.UDTName == "hstore" {
+			switch c.UDTName {
+			case "hstore":
 				c.Type = "types.HStore"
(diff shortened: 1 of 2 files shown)


aarondl:
Hm. It's nice to support more things, but I need to be convinced that we need to support all of the extension types for postgres in the core of sqlboiler. I'd rather support less things officially to be honest.
It's a long list: https://www.postgresql.org/docs/current/static/contrib.html

The cool thing about the v3 code is you can easily add support for this in a fork of the driver code which is sort of the intention, since everybody has their own take on what they need (for example postgres GIS extensions).

boxofrad:
Sure, makes sense. I'm happy to try the fork approach in our app if that's better, is there documentation on how somebody might do that?

That said, citext is a fairly simple addition as it's really just a string (compared to some of the GIS types for example) and it's particularly useful when building web apps that often need to look up records by a case-insensitive identifier such as an email address or slug.

aarondl:
So, it turns out we support `hstore` which I actually didn't know is part of contrib as well, so I suppose this is a done deal :) The only thing is we didn't test hstore so we never had the `create extension` line that set of warning alarms in my head.

Issue #265, opened 2018-04-02
If the obj is nil in the second level will panic, such as XXX.XXX


renshaohai83:
https://github.com/volatiletech/sqlboiler/blob/7e0eca627f995e2476d637ada0c3b70ecb5d837c/templates/09_relationship_to_many_eager.tpl#L31

If the obj is nil in the second level will panic, such as XXX.XXX.
Could we add a return if the obj is nil?

Thanks,
renshaohai83

aarondl:
What's the condition that triggers a null to appear there? Nullable foreign keys?

aarondl:
Stale. Try dev branch it has improvements around eager loading nullable things. With one related bug in progress.

PR #269, opened 2018-04-08
fix eager load issue when 1st loadingFrom value is nil

ceshihao:
When foreign key allows null, eager load does NOT work correctly if `val.Index(0)` is nil.

I add a helper function to return the first non nil value, so that eager loading still works even if 1st value is nil.

### DIFF
--- a/queries/eager_load.go
+++ b/queries/eager_load.go
@@ -167,7 +167,7 @@ func (l loadRelationshipState) callLoadFunction(depth int, loadingFrom reflect.V
 		if val.Len() == 0 {
 			return nil
 		}
-		val = val.Index(0)
+		val = firstNonNilValue(val)
 		if val.IsNil() {
 			return nil
 		}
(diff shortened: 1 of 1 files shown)


aarondl:
We might want to bundle this with a set of larger fixes so we don't piecemeal these to completion. I don't think we've ever truly considered nullable foreign keys nor tested them for all operations. Seems like there's another issue here about it #265

Thoughts?

ceshihao:
@aarondl 
I think it is another bug for nullable foreign key, and #265 happens during 2 level eager loading.

e.g. A->B->C
if A's foreign key (A.b_id) is nil, A.R.B is nil. Then it will panic when second level eager loading.

I have a simple solution that `continue` the loop if any `obj` is nil.

aarondl:
Yeah. This is my worry, that we haven't tested null foreign keys nearly well enough :( I think we should dedicate a branch to this. Starting with your fixes you have here.

ceshihao:
Great.
I think a `fix_nullable_fk` branch can be created, so that everyone can pull their fix.

I will create another pull for the fix mentioned in my previous comment, if I can help later.

aarondl:
Hey @ceshihao, trying to fix all the problems with nullable foreign keys. I've pushed a change to the dev branch that fixes one issue that creates bad queries by including nulls ids in the query.

I was looking at your change here and trying to understand why this is necessary. When would `loadingFrom` ever have a nil element in it?

aarondl:
It looks like #265 is related to the same thing, when there's a null object somewhere in the slice. I'm trying to understand why this would ever happen.

aarondl:
Perhaps this is a problem with the `collectLoaded` function. Maybe it needs to be checking for nils?

ceshihao:
In my case, there maybe a slice `val` from loadingFrom. Because nullable foreign key,  the first load value may be nil and the second is not nil. 
In previous codes, it returns nil if the first load value is nil which doesn't do the load function.
In my fix, it does the load function if any value in the slice is not nil.

aarondl:
I just don't understand how that case could ever happen. I'm gonna have to do some more investigating. In the meantime if you come up with a recreate let me know.

PR #275, opened 2018-04-12
Update randomize.go

brianvoe:
Doesnt return an error anymore

### DIFF
--- a/randomize/randomize.go
+++ b/randomize/randomize.go
@@ -195,10 +195,7 @@ func randomizeField(s *Seed, field reflect.Value, fieldType string, canBeNull bo
 					return nil
 				}
 				if fieldType == "uuid" {
-					randomUuid, err := uuid.NewV4()
-					if err != nil {
-						return err
-					}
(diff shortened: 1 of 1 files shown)


glerchundi:
This change rollbacks sqlboiler to an older version of satori/uuid so it is
not eligible to be merged.

brianvoe:
I would suggest add dep to ensure things are using certain versions of dependencies cause right now i get build errors.

saulortega:
Just update your library satori/uuid.

aarondl:
As @saulortega mentioned, this is not eligible to be merged. The current version of satori/uuid has the error parameter, and if we merged this many users of sqlboiler will no longer be able to build. This is also a breaking change so something I can't make lightly. I think the best path is to update your version of uuid.

sqlboiler won't commit to dep because it doesn't look like the community has, probably looking at vgo when that comes out.

CameronAckermanSEL:
@saulortega fyi https://github.com/satori/go.uuid/issues/73

PR #288, opened 2018-05-23
fix strmangle tests

ann-kilzer:
I found that the unit tests were failing due to a typo where %i was in the fmt string instead of %d. Now all unit tests pass.


aarondl:
Fixed in 331a9f6. Thanks for bringing it up. This was fixed in v3 but not backported. I didn't merge this directly because although you said the verb should be `%d` you made it `%v` and I preferred your initial assessment so I made a commit with `%d`.

Also never was a problem until Go 1.10 when `go test` started automatically running `go vet`. So long as the tests never failed, the bad printf would have never been a problem (not saying that it shouldn't have been fixed, I was just surprised by the failure! lol)

Thanks for the PR!

Issue #296, opened 2018-06-04
Eager loading of relationships may not always populate object references correctly


n3integration:
If you're having a generation problem please answer these questions before submitting your issue. Thanks!

### What version of SQLBoiler are you using (`sqlboiler --version`)?

2.5.0

### Please provide a relevant database schema so we can replicate your issue (Provided you are comfortable sharing this)

Although I cannot provide the exact schema, a fairly simple schema can reproduce the behavior.

### Further information. What did you do, what did you expect?

When eagerly loading a foreign constraint reference via `qm.Load("References")`, the resulting object graph is not always correct. If for example, loading a series of objects that reference the same foreign constraint, the SQL query produced at runtime will result a single row being returned. For example: `select * from table where id in ('abc', 'abc')`. In this case, only the first object will have its object reference set; whilst any remaining objects will not have their object reference set.

PR #297, opened 2018-06-04
[#296] - swap order of range to ensure all foreign references are set properly

n3integration:
By changing the order of range loops, the values are set correctly if less rows are returned from the SQL query.

### DIFF
--- a/templates/09_relationship_to_many_eager.tpl
+++ b/templates/09_relationship_to_many_eager.tpl
@@ -116,8 +116,8 @@ func ({{$varNameSingular}}L) Load{{$txt.Function.Name}}(e boil.Executor, singula
 		}
 	}
 	{{else -}}
-	for _, foreign := range resultSlice {
-		for _, local := range slice {
+	for _, local := range slice {
+		for _, foreign := range resultSlice {
 			{{if $txt.Function.UsesBytes -}}
 			if 0 == bytes.Compare(local.{{$txt.Function.LocalAssignment}}, foreign.{{$txt.Function.ForeignAssignment}}) {
 			{{else -}}


aarondl:
Guess this comment came back to bite me: 59c238539ddc3e0ee964e24f0f30910875c88334 haha.

Although I'm not 100% sure this is correct. The postgres tests fail to pass with this change. (You can find the test schema in the `testdata` folder).

```text
    --- FAIL: TestToMany/OwnerToCats (0.00s)
    	owner_test.go:523: number of eager loaded records wrong, got: 1
... Many others
```
I'm need to meditate on this but I'm thinking in a 1-n or n-m relationship it may be that we have to give up on using the `break` keyword completely.

Just today I was about to work on an issue where the IDs were duplicated for the select btw, so you shouldn't see `select * from table where id in ('abc', 'abc')` this kind of thing anymore after I fix that. Maybe that's part of the issue that's conflating things?

aarondl:
Also I think I might be misunderstanding what your schema looks like. I made a table that has two separate ids that are foreign keys into a second table, I tried it and it worked both with and without the patch. Maybe you could expand on the reproduce a bit?

n3integration:
Hey @aarondl - Thanks for the quick reply. Here is an example database schema (with data) and a program that demonstrates the issue.

#### SQL
```
CREATE TABLE X(
  ID SERIAL PRIMARY KEY
);

CREATE TABLE Y(
  ID SERIAL PRIMARY KEY,
  X_ID SERIAL REFERENCES X(ID)
);

INSERT INTO X VALUES(nextval('x_id_seq'));
INSERT INTO Y VALUES(nextval('y_id_seq'), (SELECT ID FROM X));
INSERT INTO Y VALUES(nextval('y_id_seq'), (SELECT ID FROM X));
INSERT INTO Y VALUES(nextval('y_id_seq'), (SELECT ID FROM X));
```

#### Test Driver
```
package main

import (
	"database/sql"
	"log"

	_ "github.com/lib/pq"
	"github.com/n3integration/sqlboilerfailure/models"
	"github.com/volatiletech/sqlboiler/boil"
	"github.com/volatiletech/sqlboiler/queries/qm"
)

func main() {
	boil.DebugMode = true
	connStr := "user=test dbname=test password=test sslmode=disable"
	db, err := sql.Open("postgres", connStr)
	if err != nil {
		log.Fatal(err)
	}

	ys, err := models.Ys(db, qm.Load("X")).All()
	if err != nil {
		panic(err)
	}
	for _, y := range ys {
		log.Println("Y =>", y)
		if y.R != nil {
			log.Println("Y->X =>", y.R.X)
		}
	}
}
```

#### Execution

```
SELECT * FROM "y";
[]
select * from "x" where "id" in ($1,$2,$3)
[1 1 1]
2018/06/05 07:29:38 Y => &{1 1 0xc42000ee38 {}}
2018/06/05 07:29:38 Y->X => &{1 <nil> {}}
2018/06/05 07:29:38 Y => &{2 1 0xc42000ee40 {}}
2018/06/05 07:29:38 Y->X => <nil>
2018/06/05 07:29:38 Y => &{3 1 0xc42000ee48 {}}
2018/06/05 07:29:38 Y->X => <nil>
```

n3integration:
Hey @aarondl - Thanks for the quick reply. Here is an example database schema (with data) and a program that demonstrates the issue.

#### SQL
```
CREATE TABLE X(
  ID SERIAL PRIMARY KEY
);

CREATE TABLE Y(
  ID SERIAL PRIMARY KEY,
  X_ID SERIAL REFERENCES X(ID)
);

INSERT INTO X VALUES(nextval('x_id_seq'));
INSERT INTO Y VALUES(nextval('y_id_seq'), (SELECT ID FROM X));
INSERT INTO Y VALUES(nextval('y_id_seq'), (SELECT ID FROM X));
INSERT INTO Y VALUES(nextval('y_id_seq'), (SELECT ID FROM X));
```

#### Test Driver
```
package main

import (
	"database/sql"
	"log"

	_ "github.com/lib/pq"
	"github.com/n3integration/sqlboilertest/models"
	"github.com/volatiletech/sqlboiler/boil"
	"github.com/volatiletech/sqlboiler/queries/qm"
)

func main() {
	boil.DebugMode = true
	connStr := "user=test dbname=test password=test sslmode=disable"
	db, err := sql.Open("postgres", connStr)
	if err != nil {
		log.Fatal(err)
	}

	ys, err := models.Ys(db, qm.Load("X")).All()
	if err != nil {
		panic(err)
	}
	for _, y := range ys {
		log.Println("Y =>", y)
		if y.R != nil {
			log.Println("Y->X =>", y.R.X)
		}
	}
}
```

#### Execution

```
SELECT * FROM "y";
[]
select * from "x" where "id" in ($1,$2,$3)
[1 1 1]
2018/06/05 07:29:38 Y => &{1 1 0xc42000ee38 {}}
2018/06/05 07:29:38 Y->X => &{1 <nil> {}}
2018/06/05 07:29:38 Y => &{2 1 0xc42000ee40 {}}
2018/06/05 07:29:38 Y->X => <nil>
2018/06/05 07:29:38 Y => &{3 1 0xc42000ee48 {}}
2018/06/05 07:29:38 Y->X => <nil>
```

aarondl:
So, I've tried the same program and the same schema and cannot recreate. I tweaked the output a bit but my is copy paste:

```
	t, err := models.Ys(db, qm.Load("X")).All()
	if err != nil {
		panic(err)
	}

	for _, v := range t {
		fmt.Println(v.ID)
		if v.R.X == nil {
			fmt.Println("  No X")
		} else {
			fmt.Println("  ", v.R.X.ID)
		}
	}
```

Which is the same as yours.

But my output differs a lot:

```
SELECT * FROM "y";
[]
select * from "x" where "id" in ($1,$2,$3)
[1 1 1]
1
   1
2
   1
3
   1
```

I noticed in your bug report your version of sqlboiler is older. Though I've looked in the history and didn't immediately find anything related to this issue, can you humor me and try the head of `dev` branch? Specifically I'm on beb0fe8d519dad9b4e88bbed77819baa770410d9 this commit just so we're clear whose using what.

n3integration:
Hey @aarondl - Given that the SQL query returns a single row, I'm not sure how iterating over the `resultSlice` could result in the correct behavior. Although I could certainly be overlooking something. I upgraded to both `2.7.0` and the `dev` branch revision that you've indicated above without any luck.

#### Gopkg.toml
```
[prune]
  go-tests = true
  unused-packages = true

[[constraint]]
  name = "github.com/volatiletech/sqlboiler"
  revision = "beb0fe8d519dad9b4e88bbed77819baa770410d9"

[[constraint]]
  branch = "master"
  name = "github.com/lib/pq"
```

#### Execution

```
SELECT * FROM "y";
[]
select * from "x" where "id" in ($1,$2,$3)
[1 1 1]
1
   1
2
  No X
3
  No X
````

aarondl:
Are you vendoring or doing anything else odd? sqlboiler has a nasty habit of looking in your `GOPATH` for templates. And even if you have a correct binary, it may be finding older templates somewhere. Can you ensure that that's not the case?

aarondl:
Also FWIW in this particular case I think you're looking at the wrong template. This should concern `07_relationship_to_one` I believe. Since Y is the "local" and X is the "foreign". In that sense the eager load we're looking at is "to_one". Which would mean that the patch that I pointed you to long ago should have fixed this, and now I think there's an even stronger likelihood of out of date templates hanging around somewhere.

n3integration:
Sure enough. I had `2.1.7` in my GOPATH. After updating to `2.7.0`, the issue that I was experiencing is resolved. I appreciate you taking the time to work through this and I think that we can close out this pull request and issue.

aarondl:
Awesome! Glad we got to the bottom of this. Sorry about that behavior, we're considering bindata for v3 so this stops occurring. And also thanks for having the gumption to dig in and submit a PR. Always nice to have people helping out :)

PR #299, opened 2018-06-06
Test authchanges

tiger5226:


aarondl:
Wow. Although it seems to have been in error that this pull request was opened. It's a pretty interesting one. Several fixes that I don't think we have, and even some extra features. I wonder why the author never tried to pull request any of them.

tiger5226:
Sorry that is our local fork. And I was trying to fix the issue because it was urgent. We are happy to share these updates I am sure.

lyoshenka:
I have filed a few PRs, but it takes time and I've been pretty busy. I could certainly do more. As for new features, some of them are actually not very good (e.g. the Filter stuff is a mistake. using qm is more flexible, but I didn't understand it at first), some make assumptions about the structure of the db (e.g. that there's always an id column), and others only target mysql, since that's all we use, so I don't test other dbs.

lyoshenka:
I'm happy you think its interesting though :smile:

PR #306, opened 2018-06-16
types/array: support full boolean values in scanBytes

### LINKED ISSUE #305: types/array: expect full boolean definitions in array
This is not needed for the current drivers but it is in order to implement the upcoming CockroachDB one.

CockroachDB serializes boolean arrays like `true,false,...` when other drivers do in a different way: `t,f,...` or `T,F`.

Will send a PR soon.

glerchundi:
Fixes #305

Signed-off-by: Gorka Lerchundi Osa <glertxundi@gmail.com>


glerchundi:
Failed to do the fix in the correct branch, superseeded by #307.

Issue #311, opened 2018-06-19
Nullable JSON field does not produce a null.JSON for MySQL.


pranava:
A nullable JSON should produce a null.JSON type but is instead producing a types.JSON in MySQL.

### What version of SQLBoiler are you using (`sqlboiler --version`)?
2.7.1

### If this happened at generation time what was the full SQLBoiler command you used to generate your models? (if not applicable leave blank)
`
sqlboiler --whitelist test mysql;
`

### Please provide a relevant database schema so we can replicate your issue (Provided you are comfortable sharing this)
`
CREATE TABLE test (
  `json_field` json DEFAULT NULL
);
`

### Further information. What did you do, what did you expect?
`
type Test struct {
	JsonField null.JSON  `...`
}
`

PR #312, opened 2018-06-19
Fix generation of nullable JSON type for MySQL

pranava:
- A nullable JSON field was being generated as types.JSON
  for MySQL
- This fix changes the generation to be null.JSON

https://github.com/volatiletech/sqlboiler/issues/311


aarondl:
Typically open your branches against dev (as mentioned in contribution guidelines). In this case there was nothing in that branch so it worked out fine though. Thanks for the bug fix!

PR #221, opened 2017-11-14
fix mysql type TIME should not covert to time.Time

ceshihao:
As per mysql driver document.

https://github.com/go-sql-driver/mysql#parsetime
```
`parseTime=true` changes the output type of `DATE` and `DATETIME` values to `time.Time` instead of `[]byte` / `string`
```
However type `TIME` is scaned as raw bytes.
https://github.com/go-sql-driver/mysql/blob/master/fields.go#L124

Therefore, models throw out an error when converting `TIME` type column to `time.Time`

PR #325, opened 2018-07-04
fix random time issue in test

ceshihao:
After #221

Mysql type TIME is scanned as `string`
But test will fail if random `TIME` as general string.

I can also port this fix to **v3** later.

### DIFF
--- a/randomize/random.go
+++ b/randomize/random.go
@@ -90,6 +90,10 @@ func randMoney(s *Seed) string {
 	return fmt.Sprintf("%d.00", s.nextInt())
 }
 
+func randTime() string {
+	return fmt.Sprintf("%d:%d:%d", rand.Intn(24), rand.Intn(60), rand.Intn(60))
+}
+
 // StableDBName takes a database name in, and generates
 // a random string using the database name as the rand Seed.
 // getDBNameHash is used to generate unique test database names.
--- a/randomize/randomize.go
+++ b/randomize/randomize.go
@@ -55,7 +55,7 @@ var (
 		"inet", "line", "uuid", "interval", "mediumint",
 		"json", "jsonb", "box", "cidr", "circle",
 		"lseg", "macaddr", "path", "pg_lsn", "point",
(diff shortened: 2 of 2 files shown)


aarondl:
Hey @ceshihao. I expected this to be fixed in a different place in the randomize code because it shouldn't be a string. And looking into it sure enough
I found that v2 seems to be doing the wrong thing:

mysql driver v3 code properly handles this (I think):
https://github.com/volatiletech/sqlboiler/blob/7c76df8250d329a928e6259d4fe2cd9de89f8606/drivers/sqlboiler-mysql/driver/mysql.go#L389
https://github.com/volatiletech/sqlboiler/blob/7c76df8250d329a928e6259d4fe2cd9de89f8606/drivers/sqlboiler-mysql/driver/mysql.go#L441

mysql driver v2 code is missing a case for `time`:
https://github.com/volatiletech/sqlboiler/blob/d4bc831c7df09032b8290a3ddfb7bb41a0866934/bdb/drivers/mysql.go#L317
https://github.com/volatiletech/sqlboiler/blob/d4bc831c7df09032b8290a3ddfb7bb41a0866934/bdb/drivers/mysql.go#L367

Assuming that v3 isn't doing the incorrect thing, we should fix v2 to do the same. Once that's fixed, there's a different place in the randomize
code to fix this:

v2:
https://github.com/volatiletech/sqlboiler/blob/d4bc831c7df09032b8290a3ddfb7bb41a0866934/randomize/randomize.go#L457
https://github.com/volatiletech/sqlboiler/blob/d4bc831c7df09032b8290a3ddfb7bb41a0866934/randomize/randomize.go#L503

v3:
https://github.com/volatiletech/sqlboiler/blob/7c76df8250d329a928e6259d4fe2cd9de89f8606/randomize/randomize.go#L164
https://github.com/volatiletech/sqlboiler/blob/7c76df8250d329a928e6259d4fe2cd9de89f8606/randomize/randomize.go#L176

aarondl:
Merged. Thx. Working on fixing up these problems in v3 now.

PR #326, opened 2018-07-04
v3 fix random time issue in test

ceshihao:
v3 port for #325


aarondl:
Can you verify that this works for you @ceshihao? :)

PR #332, opened 2018-07-11
[Feature] Extensible strmangle

shousper:
We want to use sqlboiler, but our existing database isn't using snake_case. This is an attempt to expose a few of the internals related to this requirement so we can support our convention (a mixture of TitleCase table names and camelCase columns).

I should point out that no changes have been made to the `strmangle` public function implementations. I just moved them to `impl.go` and added an interface in `strmangle.go` with the `DefaultMangler` global, of which all the global `strmangle` functions now use 👍 The change should be completely non-breaking.

Hopefully this is an acceptable change :)

### DIFF
--- a/boilingcore/text_helpers.go
+++ b/boilingcore/text_helpers.go
@@ -272,12 +272,12 @@ func mkFunctionName(fkeyTableSingular, foreignTablePluralGo, fkeyColumn string, 
 	return strmangle.TitleCase(colName) + foreignTablePluralGo
 }
 
-var identifierSuffixes = []string{"_id", "_uuid", "_guid", "_oid"}
+var IdentifierSuffixes = []string{"_id", "_uuid", "_guid", "_oid"}
 
 // trimSuffixes from the identifier
 func trimSuffixes(str string) string {
 	ln := len(str)
-	for _, s := range identifierSuffixes {
+	for _, s := range IdentifierSuffixes {
 		str = strings.TrimSuffix(str, s)
 		if len(str) != ln {
 			break
--- a/boilingcore/text_helpers_test.go
+++ b/boilingcore/text_helpers_test.go
@@ -264,7 +264,7 @@ func TestTxtNameToMany(t *testing.T) {
 func TestTrimSuffixes(t *testing.T) {
 	t.Parallel()
 
-	for _, s := range identifierSuffixes {
+	for _, s := range IdentifierSuffixes {
 		a := "hello" + s
 
 		if z := trimSuffixes(a); z != "hello" {
--- a/strmangle/impl.go
+++ b/strmangle/impl.go
@@ -1,0 +1,693 @@
+// Package strmangle is a collection of string manipulation functions.
+// Primarily used by boil and templates for code generation.
+// Because it is focused on pipelining inside templates
+// you will see some odd parameter ordering.
+package strmangle
+
+import (
+	"fmt"
+	"math"
+	"regexp"
+	"sort"
+	"strings"
+	"sync"
+)
+
+var (
+	idAlphabet    = []byte("abcdefghijklmnopqrstuvwxyz")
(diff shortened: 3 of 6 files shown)


aarondl:
It's definitely an interesting proposition. The feature itself is great. Not happy with the implementation because I don't like the huge interface. Not dismissing this implementation completely, I'd just like a bit of time when I'm back to dig into the problem in more detail and consider other solutions.

Thanks very much for the PR. I'll be looking at it next week in more detail.

shousper:
Thanks @aarondl, I'm glad you feel that way :) I'm also not a fan of the huge interface, but I needed to get something operational for us, and I wanted to share that we're quite keen to use sqlboiler, just with a non-snake-case convention.

If you'd like my 2 cents, I was thinking along the lines of:
- Implementing a "wrapper" struct for the names of tables and columns as pulled from DBMS metadata, similar to [this little gem](https://github.com/markbates/inflect/blob/master/name.go).
- Adding a "naming strategy" interface that only exposes methods that name outputs in the templates, similar to [this](https://github.com/typeorm/typeorm/blob/master/src/naming-strategy/NamingStrategyInterface.ts) but reversed (because typeorm is not a database-first ORM).
- Implementing 2 naming strategies for the current snake case pattern and our camel case usage.
- Adding a global registry for the naming strategies (similar to how `database/sql` registers drivers).
- Adding a configuration parameter to set the naming strategy to use.

Some of the methods currently on the `Mangler` interface I've got here probably don't belong as part of a naming strategy, because Go dictates the format for interface, struct, field, function and method names -- so we don't need different strategies for that :) (thanks Go!)

Another option that could be attractive would be [plugins](https://golang.org/pkg/plugin/) if you really wanted to expose the core of sqlboiler for vast extension.

Anyway, I wasn't going to go too far out on a limb and implement this kind of thing straight up in case you didn't like it, so here we are :)

I've [raised an issue](https://github.com/volatiletech/sqlboiler/issues/334) for this as well

aarondl:
Thanks for raising the issue. I also want to ask the question if you've considered aliasing your tables. Are you aware of the aliasing feature in v3? This allows you to control the generated names completely. It might not be completely ideal but it would allow you use a non-fork immediately.

aarondl:
Oh and furthermore reading some of your suggestions definitely not looking for any new pluigin mechanisms. The new driver model is painful enough!

shousper:
Wow, I was not at all aware of the v3 branch, haha!
I see its at RC9 so hopefully pretty stable and close to release :)

I'll take a look at this more closely and probably closer this PR.

Thanks @aarondl!

shousper:
Hey @aarondl, it [looks like](https://github.com/volatiletech/sqlboiler/blob/v3/boilingcore/text_helpers.go#L112) I'm still going to have some trouble with aliasing because these hard-coded identifier suffixes.

Ours are all `ID` so tables with relationships using this pattern create fields and methods with the same name, resulting in this:

```
models/ClientMessages.go:315:6: type ClientMessage has both field and method named ClientMessageTypeID
models/ClientSettings.go:303:6: type ClientSetting has both field and method named ClientID
models/Clients.go:445:6: type Client has both field and method named ClientStatusID
```

Is there a way to override these? Could be just another thing to add to the config 😃

aarondl:
Relationship names are also configurable. Have you seen the Readme that documents the feature? There's also an alternative syntax if your toml keys need case sensitivity.

https://github.com/volatiletech/sqlboiler/blob/v3/README.md#aliases

If something is still not doable/not working let me know.

shousper:
Thanks @aarondl. Yeah, I found the docs 👍 We'll definitely need this if we want any hope of using sqlboiler 😃 

I eventually bent my head around TOML. I switched to a JSON configuration file and got it working, but was determined to understand TOML better.

In the end, I couldn't get this syntax working:

```
[[aliases.tables.ClientMessages.relationships]]
```

I needed to use this:

```
[[aliases.tables]]
name = "ClientMessages"

    [[aliases.tables.relationships]]
```

Cheers for your help, I really appreciate it! I think I can close this off for now, but I'm happy to chat more about naming strategies in the #334 👍

PR #333, opened 2018-07-11
Added singular version of 'menus'

jonas747:
Pretty self explanatory, i ran into a issue when i had a table named "menus"


aarondl:
Thanks for the PR. I've also added the change to the v3 branch. Which if you're not aware of you should check out. It will become the master branch in fairly short order and v2 will be getting less and less attention. Thanks again :)

PR #347, opened 2018-07-17
Allow empty pass in drivers

izumin5210:
sqlboiler v3's README describes that the pass is an optional parameter.
But current implementation does not allow an empty pass.


izumin5210:
I fixed in the mssql driver.
But the DSN functionality seems good to me 😄

aarondl:
Merged into master. Thanks for the PR, sorry about the massive delays :)

aarondl:
Forgot to mention v3 became master today. That's why I did it this way :) Yay for release.

Issue #337, opened 2018-07-13
[Feature Request] Allow connection without password


shousper:
For local development and CI we don't use a password for our database connections. Could this requirement could be lifted from sqlboiler?

Alternatively, supplying the DSN instead of individual connection string components could be good 👍 

Thoughts? :)

aarondl:
I would consider having a DSN parameter that when present took precedence over the piecemeal ones. Later we could maybe deprecate the individual args if it was desirable. I'm not really sure why they ended up separate, hold over from rails world I guess.

aarondl:
See #351 comments for why DSN won't be added as a feature.

PR #347 fixes this issue and will be merged shortly.

PR #351, opened 2018-07-21
Support DSN parameter

izumin5210:
https://github.com/volatiletech/sqlboiler/issues/337

I tried to implement parsing DSN parameter.
But I have no knowledge of mssql, so I don't implement for the mssql driver...

### DIFF
--- a/drivers/sqlboiler-mysql/driver/mysql.go
+++ b/drivers/sqlboiler-mysql/driver/mysql.go
@@ -60,34 +60,28 @@ func (m *MySQLDriver) Assemble(config drivers.Config) (dbinfo *drivers.DBInfo, e
 		}
 	}()
 
-	var (
-		user, pass, dbname, host string
-		port                     int
-	)
+	var schema string
 
 	if dsn, ok := config.String(drivers.ConfigDSN); ok {
 		cfg, err := mysql.ParseDSN(dsn)
 		if err != nil {
 			return nil, errors.WithStack(err)
 		}
-		user = cfg.User
-		pass = cfg.Passwd
-		dbname = cfg.DBName
-		hostAndPort := strings.SplitN(cfg.Addr, ":", 2)
-		host = hostAndPort[0]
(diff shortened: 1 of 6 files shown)


drivers/sqlboiler-mysql/driver/mysql.go
```diff
@@ -60,13 +60,33 @@ func (m *MySQLDriver) Assemble(config drivers.Config) (dbinfo *drivers.DBInfo, e
...
-	dbname := config.MustString(drivers.ConfigDBName)
-	host := config.MustString(drivers.ConfigHost)
-	port := config.DefaultInt(drivers.ConfigPort, 3306)
-	sslmode := config.DefaultString(drivers.ConfigSSLMode, "true")
+	var (
+		user, pass, dbname, host string
+		port                     int
+	)
+
+	if dsn, ok := config.String(drivers.ConfigDSN); ok {
```
aarondl: Rather than looking up the DSN, parsing it, then assembling it into a DSN again, you should just assign `m.connStr = dsn`, half of the reason for taking the dsn directly is that there's features the config file doesn't allow you to provide, and you want to override it's choices.

izumin5210:
Fixed! 😄

aarondl:
I really really hate to do this. But I'm going to actually close this PR without merging it in.

After some thought, I decided that this feature isn't necessary. The reasoning is that right now this dsn string is only used to connect to the database for generation. When in reality the most important time to have a correct dsn string is when we're connecting to the database for actual use of generated code, not at generation time.

It's also true that there was no way the generated tests could use the same dsn parameter, since they use the parameterized form (user, pass, host etc.) to write out password files for mysql/psql for various reasons.

This inconsistency between generation/tests/your application makes it a pretty unsuitable feature. Although it sounded good at the time, it was not fully thought through, and I apologize for asking for effort before I had considered it enough :( I've merged your other PR though, thanks for that :)

PR #357, opened 2018-08-01
enable support for IDENTITY on PostgreSQL 10.2+

### LINKED ISSUE #356: PostgreSQL 10.2+ IDENTITY support (SERIAL replacement)
v3.0.0-rc10

PostgreSQL introduced `GENERATE IDENTITY` feature in 10.2 to drop-in replacement for `SERIAL`.

When a table primary key `id` generated using an `IDENTITY` it does not have a default value like `SERIAL` has which is `nextval('test_id_seq'::regclass)`.

Due to this, `ID` inserted as `0` all the time and causes duplicate pkey.

So this is an actually a request to support `IDENTITY` for auto-increment. It will be very handy if we at least add this to `3.0.0-stable`.

### Workaround

If we do Blacklist for ID then everything goes smooth.

`p1.Insert(db, boil.Blacklist("id"))`

But this has a downside too. We do not get `p1.ID` back as it is blacklisted.

Or, manually edit models as below:

### Solution

If `ID` is an `IDENTITY` remove "id" from `messageColumnsWithoutDefault` and add it to the `messageColumnsWithDefault` as follows:

```
	messageColumnsWithoutDefault = []string{"id", "user_id", "target_user_id", "message", "created_at", "updated_at"}
	messageColumnsWithDefault    = []string{}
```

_should be:_

```
	messageColumnsWithoutDefault = []string{"user_id", "target_user_id", "message", "created_at", "updated_at"}
	messageColumnsWithDefault    = []string{"id"}
```

gencer:
PostgreSQL 10.2+ introduces IDENTITY columns as drop-in replacement for SERIAL. This fix allow identifying identities and mark them as with defaults.

Before doing this, we check if PostgreSQL information_schema has is_identity field. If not then false is returned. This provides backward-compatibility for older versions of PostgreSQL that does not have this field.

This fixes #356.

---

Let me explain in steps:

1. I check `is_identity` column exists in `information_schema`. If exists, this means this is PostgreSQL 10.2+.
2. If first is `TRUE` then I check `is_identity` column in our generated `table` (inside the query)
3. If 1 `TRUE` and 2 is `YES`, we have a winner here. This is `IDENTITY` field.

If in any case 1 or 2 fails, `FALSE` returns. Otherwise `TRUE`.

### DIFF
--- a/drivers/sqlboiler-psql/driver/psql.go
+++ b/drivers/sqlboiler-psql/driver/psql.go
@@ -216,6 +216,14 @@ func (p *PostgresDriver) Columns(schema, tableName string, whitelist, blacklist 
 		c.column_default,
 
 		c.is_nullable = 'YES' as is_nullable,
+		(case
+			when (select 
+		    case
+			    when column_name = 'is_identity' then (select c.is_identity = 'YES' as is_identity)
+		    else 
+			    false
+		    end as is_identity from information_schema.columns
+		    WHERE table_schema='information_schema' and table_name='columns' and column_name='is_identity') IS NULL then 'NO' else is_identity end) = 'YES' as is_identity,
 		(select exists(
 			select 1
 			from information_schema.table_constraints tc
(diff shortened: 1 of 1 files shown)


aarondl:
Merged into master (v3 has been released today, and includes this change). Thanks so much for the PR!

PR #368, opened 2018-08-13
Add import for nullable JSON fields in MySQL

mickeyreiss:
Fixes a possible regression from #311. Before this fix, generated models with nullable JSON columns are missing the import for `null`.


aarondl:
Merged into master (v3 is now released), thanks for the PR!

Issue #364, opened 2018-08-10
Bulk Insert


taxio:
Is there support for Bulk Insert? I cannot find this function in this project... : (
Is there plans to add it?

aarondl:
It doesn't exist right now in sqlboiler. No plans to add currently. The `lib/pq` bulk insertion API is really weird and so I decided just not to bother for the time being. Use the driver directly to get the job done.

ORESoftware:
any word on this? thanks

ORESoftware:
bulk insert into postgres at least has a simple api:

```sql
INSERT INTO products (product_no, name, price) VALUES
    (1, 'Cheese', 9.99),
    (2, 'Bread', 1.99),
    (3, 'Milk', 2.99);
```

I assume other relational db engines have very similar syntax, so I don't see why all pg drivers wouldn't support this, seems worth implementing to me.

PR #374, opened 2018-08-21
Multi Insert

taxio:
#364 

Hi, aarondl.
I want sqlboiler to support Multi Insert (e.g `INSERT INTO "hoge" VALUES (a, b, c), (d, e, f)`).

Multi insert is required even in other libraries. https://github.com/jmoiron/sqlx/issues/134 

It is my first time to send PR to OSS, so please tell me anything if there is a fix.


aarondl:
So I just double-checked and ensured that all of our database engines support this type of insert (and it looks like they do). I'll probably accept this PR with some tweaks. But first a thorough review. I'll find time when I can to get that done. Thanks for submitting it.

aarondl:
It's acceptance is also conditional on whether or not we can actually populate the default values correctly. I have a feeling for `mysql` this is impossible.

taxio:
Thanks for comments.

I don't understand enough about mysql...
Since I am in your slack, please feel free to advise.

templates/15_insert.go.tpl
```diff
@@ -201,3 +201,168 @@ CacheNoHooks:
...
+	{{- end}}
+	if err != nil {
+		return errors.Wrap(err, "{{.PkgName}}: unable to insert into {{.Table.Name}}")
+	}
+	{{if $canLastInsertID -}}
+	var lastID int64
+	{{- end}}
+
+	{{if $canLastInsertID -}}
+	lastID, err = result.LastInsertId()
```
aarondl: If this code is not looped over as the rest of it is, there's no chance that mysql values will be returned for all inserted rows.

taxio:
I fixed getting default values.


Maybe, I mistook the operation of git, so if you need, I will reissue PR later...

aarondl:
Hey @taxio, can you clean up this PR first? There's too much noise in the diff.

PR #375, opened 2018-08-22
Fix blacklist filter for mssql driver

DukeLog:
Blacklist option for mssql doesn't work correctly. Here is a little fix.


aarondl:
Thanks for the PR!

PR #378, opened 2018-08-24
Add ToOneRels and ToManyRels constants to struct template

visortim:
@mickeyreiss


visortim:
We changed our use case on this one, deleting the pull request.

PR #387, opened 2018-09-10
Fix: `panic: assignment to entry in nil map`

nadilas:
```
panic: assignment to entry in nil map

goroutine 1 [running]:
github.com/volatiletech/sqlboiler/boilingcore.FillAliases(0xc04215e9d8, 0xc04229c000, 0x8, 0x9)
	C:/Users/veres/go/src/github.com/volatiletech/sqlboiler/boilingcore/aliases.go:154 +0x1274
```

For existing tables the Relationships is not getting initialized, hence a join table runs into a nil map problem

### DIFF
--- a/boilingcore/aliases.go
+++ b/boilingcore/aliases.go
@@ -42,8 +42,11 @@ func FillAliases(a *Aliases, tables []drivers.Table) {
 
 	for _, t := range tables {
 		if t.IsJoinTable {
-			if _, ok := a.Tables[t.Name]; !ok  || a.Tables[t.Name].Relationships == nil {
+			jt, ok := a.Tables[t.Name];
+			if !ok {
 				a.Tables[t.Name] = TableAlias{Relationships: make(map[string]RelationshipAlias)}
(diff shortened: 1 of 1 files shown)


nadilas:
Hey @aarondl, no worries. :)

I actually did have a separate else if branch in the first place but in my case that was never hit, since I had no other info attached to the join table in my defs. But true, we are recreating the whole table, deleting anything what was there before. This was not intentional and requires the change you mentioned to preserve the original table and only initialize the relationships if need be. 

The double read from the table, is a good catch, I’ll rewrite it. ;)

Could you give me an example of aliases on the relationships with the alternative syntax (case sensitive) so I can debug both ways? I never tried to set it to be honest.

aarondl:
Sounds good.

An alternative syntax example is at the very bottom of this section in the readme:
https://github.com/volatiletech/sqlboiler#aliases

nadilas:
@aarondl I modified the condition, let me know if it's ok.

aarondl:
LGTM

PR #389, opened 2018-09-11
fix MySQL bigint type

nazo:


aarondl:
I swear we've fixed this before. Thank you for the PR. Normally you should make your pull requests against the dev branch as it states in the contributing guidelines. So I can't merge this PR directly. Instead it's merged here ddfa759 and will be in master once the next release hits.

PR #393, opened 2018-09-22
Allow output directory to be multiple levels deep

gedorinku:
When `output` is set to a directory multiple levels deep, the generated tests fail...
### sqlboiler.toml
``` toml
output = "foo/models"
pkgname = "models"

[psql]
  dbname = "boiler_sample"
  host = "localhost"
  port = 5432
  user = "postgres"
  pass = ""
  sslmode = "disable"

```
```
$ sqlboiler psql
$ go test ./...
Unable to execute setup: Parameter validation failed:
        Parameter is an empty string: psql.user
        Parameter is an empty string: psql.host
        Parameter is an empty string: psql.dbname
FAIL    github.com/gedorinku/sqlboiler-sample/foo/models        0.013s
```

This is caused by the following template, so I fixed it.
https://github.com/volatiletech/sqlboiler/blob/1340e5544bc5880f10808da79ebb3ee80cf8abb1/templates_test/singleton/boil_main_test.go.tpl#L73-L78


aarondl:
Thanks for this :)

PR #401, opened 2018-10-10
Add UpdateAllG function for query object.

gemscng:


CONTRIBUTING.md
```diff
@@ -40,6 +40,34 @@ up as an issue to avoid penalizing the rest of the users for not being on Slack.
 the way to do something, then open the PR against the **dev** branch and we'll commence code review
 with the Github code review tools. Then it will be merged into dev, and later go out in a release.
 
+# Templates
+
+1. add sqlboiler [Configuration files](https://github.com/volatiletech/sqlboiler#configuration).
+1. You may need to install go-bindata first
+
+   ```
+   brew install go-bindata
```
aarondl: This is mac specific and does not work on other platforms though we support three other ones. This should be replaced with a `go get`

CONTRIBUTING.md
```diff
@@ -40,6 +40,34 @@ up as an issue to avoid penalizing the rest of the users for not being on Slack.
...
+1. generate code for template file using following command
+
+   ```
+   ./boil.sh go-generate all
+   ```
+
+1. run command to generate the models
+
+   ```
+   go run main.go [driver]
```
aarondl: This is also not the recommended way to run sqlboiler and only works because there's no other files other than `main.go` in package main. If we ever add a second file, this command will fail. Better to simply recommend: `sqlboiler [driver]`. You could also use `./boil.sh gen psql`.

CONTRIBUTING.md
```diff
@@ -40,6 +40,34 @@ up as an issue to avoid penalizing the rest of the users for not being on Slack.
...
+   ```
+
+1. run test
+
+   ```
+   go test ./models/...
```
aarondl: Worth noting that `./boil.sh test` also exists which adds `-race` and `-v` to the flags.

aarondl:
Sorry for the late merge.

PR #406, opened 2018-10-24
Add ctx to Find examples

jones77:


aarondl:
Thanks for the PR. Merged, will push in a short while here.

Issue #405, opened 2018-10-24
Eager loading a many-to-many relationship seems to break when columns in eager-loaded table are not in alphabetical order


cpickett-ml:
### What version of SQLBoiler are you using (`sqlboiler --version`)?
SQLBoiler v3.0.1

### If this happened at runtime what code produced the issue? (if not applicable leave blank)
```
package main

import (
	"database/sql"
	"fmt"

	"github.com/volatiletech/sqlboiler/boil"
	_ "github.com/volatiletech/sqlboiler/drivers/sqlboiler-mysql/driver" // mysql connection
	"github.com/volatiletech/sqlboiler/queries/qm"
	"pernicious.games/test/app/models"
)

//go:generate sqlboiler --wipe --add-global-variants --no-context --no-tests --no-hooks -o app\models mysql

func main() {
	db, _ := sql.Open("mysql", "root:3.asi79>%swB@tcp/sqlboil_test?charset=utf8&parseTime=True")

	player := &models.Player{Name: "Test player"}
	player.Insert(db, boil.Infer())
	team := &models.Team{Name: "Test team"}
	team.Insert(db, boil.Infer())
	player.SetTeams(db, false, team)

	foundPlayer, err := models.Players(qm.Load("Teams"), qm.Where("id = ?", player.ID)).One(db)
	fmt.Println(foundPlayer, err)
}
```

### What is the output of the command above with the `-d` flag added to it? (Provided you are comfortable sharing this, it contains a blueprint of your schema)
```
[...]

PR #410, opened 2018-10-30
Issue #405 - Fixed MySQL column ordering when building model

cpickett-ml:
https://dev.mysql.com/doc/refman/8.0/en/columns-table.html indicates that selecting from the columns table has no guarantee of column order. In my case, I was getting the columns back (sometimes) in alphabetical order instead of the order they appeared in the table.


aarondl:
Awesome. Nice investigation in the issue thread. Thanks for the PR

PR #428, opened 2018-11-18
Issue #224 - write stderr into buffer for mssql, mysql and psql tests

### LINKED ISSUE #224: Model tests should show stdout/stderr of commands on failure
pg_dump may fail under certain conditions (in my case it was due to client-server version mismatch, since I'm running pg inside a Docker container). If so, you get a not very informative "exit status 1" error:

```
$ go test ./models
exit status 1
Unable to execute setup: failed to wait for pg_dump command: exit status 1
FAIL	github.com/XX/XX/models	0.275s
```

whereas when running pg_dump directly you clearly see the issue:
```
$ pg_dump -s -U postgres -h localhost
pg_dump: server version: 10.1; pg_dump version: 9.6.5
pg_dump: aborting because of server version mismatch
```

This could be solved by printing pg_dump's stderr in case of failure.

LukasAuerbeck:
Fixes #224 

Print stderr if the commands in test setup fail.

### DIFF
--- a/drivers/sqlboiler-mssql/driver/override/templates_test/singleton/mssql_main_test.go.tpl
+++ b/drivers/sqlboiler-mssql/driver/override/templates_test/singleton/mssql_main_test.go.tpl
@@ -60,7 +60,9 @@ func (m *mssqlTester) setup() error {
 
 	defer func() { _ = f.Close() }()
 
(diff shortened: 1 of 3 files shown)


aarondl:
Thanks very much for this! Amazed someone got around to it haha.

PR #444, opened 2018-12-14
add err check after rows.Next() in bind function

ceshihao:
Add err check after rows.Next() in bind function, like the example of `QueryContext`.
Some sql errors will be ignored in previous version.

```go
	rows, err := db.QueryContext(ctx, "SELECT name FROM users WHERE age=?", age)
	if err != nil {
		log.Fatal(err)
	}
	defer rows.Close()
	names := make([]string, 0)
	for rows.Next() {
		var name string
		if err := rows.Scan(&name); err != nil {
			log.Fatal(err)
		}
		names = append(names, name)
	}
	if err := rows.Err(); err != nil {
		log.Fatal(err)
	}
```
https://golang.org/pkg/database/sql/#DB.QueryContext

### DIFF
--- a/queries/reflect.go
+++ b/queries/reflect.go
@@ -288,6 +288,9 @@ Rows:
 			ptrSlice.Set(reflect.Append(ptrSlice, newStruct))
 		}
 	}
+	if err := rows.Err(); err != nil {
+		return err
+	}
 
 	if bkind == kindStruct && !foundOne {
 		return sql.ErrNoRows


aarondl:
Hey @ceshihao. This is actually intentionally missing this error check because it's supposed to be checked by the caller.

You can see that a proper caller cleans up the rows:
https://github.com/volatiletech/sqlboiler/blob/dev/queries/reflect.go#L119

Whereas if you pass in the rows, you are responsible for creation and all of: `.Close`, `.Err`.
https://github.com/volatiletech/sqlboiler/blob/dev/queries/reflect.go#L101

ceshihao:
@aarondl 
I see what you mean.
But I think `Bind` is better to have some necessary defensive code since it is an exported function.

aarondl:
I think `Bind` is a special use case (especially that particular version of bind). Normally you'll be binding off a `Query` which does the appropriate error checks etc. Due to the specialized use case of this I think additional documentation should suffice. Thoughts?

ceshihao:
I think additional documentation is an option.

But it is very easy to be ignored for lots of `specialized use cases`.

aarondl:
Hm. I think I like the api this creates more. I'll probably merge this. Otherwise when calling this method you end up doing two error checks per call.

ceshihao:
Glad to hear that. Thanks for all your efforts.

aarondl:
Unfortunately after looking into this it can't actually work.

The `bind` method you had modified was the wrong place, since that is not an exported function and the callers are definitely responsible in that case for checking errors. There are only two callers in sqlboiler:

`(*Query).Bind` already closes things and checks for errors, this is not an issue.
`queries.Bind` does not `Close` the rows (as a feature for advanced use cases) and therefore an `Err` check is premature. The documentation for rows says:

> Err returns the error, if any, that was encountered during iteration. Err may be called after an explicit or implicit Close.

https://golang.org/pkg/database/sql/#Rows.Err

So in one case (where you bind to a slice) the error check will be useful since the rows will be implicitly closed by the fact that we've iterated through all of them.

However in the other (where you have multiple rows and you're binding one at a time) not only is the error check extraneous since you shouldn't check for errors until you're done iteration, but it may not report an error until the `.Close` has been called at the end of iteration which means you would have ignored your error.

Due to this inconsistency I don't think it's a good idea to do it this way. I've amended the documentation instead for now.

PR #465, opened 2019-01-29
Update hstore.go

mayank1010:
changed type in hQuote() from sql.NullString to null.String


aarondl:
As I mentioned in Slack I don't think we can change this currently as it's a breaking change. But it would definitely be included in a v4.0

aarondl:
Only after I fixed this bug did I see this PR as necessary. I'm sorry I didn't look hard enough at this PR and would have accepted it if I did :( The issue is fixed in dev though if that's any consolation.

PR #468, opened 2019-02-03
Improve documentation regarding aliases

NickyMateev:
# Improve documentation regarding aliases

I recently had an issue with generating my sqlboiler models.
Turned out I had collisions with naming and so I had to read up on the aliasing features of sqlboiler.

I'll be honest, it took me a fair amount of time to understand the difference between the **local** and **foreign** attributes and how they behave in different kinds of relationships (one-to-many, many-to-many, etc.)

I did my best to try to improve the aliases documentation in a way that I believe will be at least a little more descriptive so that it's easier to understand for newcomers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -446,7 +446,7 @@ available via the config file and can use some explanation.
 ##### Aliases
 
 In sqlboiler, names are automatically generated for you. If you name your database
-nice things you will likely have nice names in the end. However in the case where your
+entitiese properly you will likely have descriptive names generated in the end. However in the case where thee
 names in your database are bad AND unchangeable, or sqlboiler's inference doesn't understand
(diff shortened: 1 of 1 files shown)


aarondl:
I don't mind this documentation, but I don't think it's a huge improvement over the original. It conveys the same information in more words but it doesn't seem to add very much. I think what we really need is several concrete examples (a toml config file that shows all alias possibilities, and a set of tables in SQL that have all possible relationships to go with it) for people to grasp this concept as it's quite simple in reality but very difficult to explain. This would be a more visual approach to showing people how this works and I think it would compliment what you've written to really help people understand. I really appreciate the effort though, thanks for this.

PR #472, opened 2019-02-07
Fix handling of custom Pg domain that is based on an array

autarch:
Previously this would lead to a runtime nil pointer panic because *c.ArrType
was not populated. Now we check for this case and use the c.UDTName value,
which should be populated in these cases.

To see this in action create a table like this:

```sql
  CREATE DOMAIN text_array AS TEXT[] CHECK ( ARRAY_LENGTH( value, 1 ) >= 1 );

  CREATE TABLE foo (
      foo_id  INT  PRIMARY KEY,
      col1    text_array
  );
```


aarondl:
At first glance this seems fine, will take a deeper look and merge a bit later.

aarondl:
Thanks for the PR! Merged here on dev: 3dfda68
Please keep all contributions targetting the `dev` branch as outlined in CONTRIBUTING.md

PR #475, opened 2019-02-08
Make sqlboiler dump idempotent

autarch:
This makes dumping the same schema repeatedly idempotent (except for MS SQL which is idempotent only if you explicitly name all your FKs).


autarch:
Note that this and #472 have some conflicts. If you merge this first I'll need to regen the psql golden JSON in that one. If you merge that one first I'll need to regen it here.

autarch:
I'm fine with removing the loops although I'd note that on my system running with the loops as is takes 2.7s for the psql test, which isn't much. I removed the loops and it took 0.7s. So it basically costs 2s for psql.

Anyway, I'll remove the loops.

autarch:
I force pushed an update sans loops.

autarch:
I'll also rebase this off dev to deal with the conflict with my other PR.

autarch:
Alright, this should up to date with dev. I also changed the PR to be against dev instead of master.

aarondl:
Forgot to ask, what's the reasoning behind indenting golden files?

autarch:
If you regen the golden file and it's changed, having it indented makes it possible to see what the changes are using `git diff`.

aarondl:
Thanks. This was a great change.

aarondl:
I missed it but this feature actually breaks when there are blacklists in the config file. Fixed on dev branch. This comment is for posterity.

PR #478, opened 2019-02-11
Fix an accidental rename of an imported package

autarch:


autarch:
I'm really kind of hating Go rigth now.

autarch:
Sorry about leaving this in here.

aarondl:
Ugh. Should have seen that :( I think Go modules allows you to rewrite to a particular fork fairly easily. They're trying to address the problem.

autarch:
I think Go modules might help the importer of a package but it doesn't help in this particular case (unless I've missed something).

It'd be good to add a test that checks for imports of "/*/sqlboiler" that aren't from volatiletech.

aarondl:
It'd be good to have any amount of CI that's not 0. Ever since CircleCI got weird and broke several times I've sort of neglected to get it back up and running.

Issue #480, opened 2019-02-15
Feature request: Add a query mod to allow passing in CTEs


lucaslsl:
Currently the only way to write CTEs (Common Table Expressions) is writing raw queries (`SQL()` mod). But by doing so you lose the richness of dynamically building queries using mods `Select(), Where(), ...`.

Would be nice to have a query mod called `With()` to allow for CTEs to be included in a query (SELECT/UPDATE/DELETE). Working like below:
```Go
models.Pilots(
    With("cte_0 AS (SELECT * FROM other_0)"),
    With("cte_1 AS (SELECT * FROM other_1 WHERE stuff=?)", 3),
    InnerJoin("cte_0 ON cte_0.stuff = pilots.stuff"),
).All()
```
producing:
```SQL
WITH cte_0 AS (SELECT * FROM other_0), cte_1 AS (SELECT * FROM other_1 WHERE stuff=$1) SELECT * FROM "pilots" INNER JOIN cte_0 ON cte_0.stuff = pilots.stuff
```

PR #481, opened 2019-02-15
Adds "With" Query Mod (CTEs)

lucaslsl:
- Adds a query mod called `With(clause string, args ...interface{})` to allow setting Common Table Expressions in a query.

Proposed in: https://github.com/volatiletech/sqlboiler/issues/480

### DIFF
--- a/queries/_fixtures/16.sql
+++ b/queries/_fixtures/16.sql
@@ -1,0 +1,1 @@
+WITH cte_0 AS (SELECT * FROM other_t0), cte_1 AS (SELECT * FROM other_t1 WHERE thing=$1 AND stuff=$2) SELECT * FROM "t";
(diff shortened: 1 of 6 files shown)


aarondl:
Hi Lucas. Thanks again for this PR. I've merged it however it was of pretty poor quality. The tests had syntax errors which is pretty egregious as it means they've never been run and it's clear that a linter had never been run on it as there was an easily invertible if that it should have flagged.

See commits: 733c00d and 2d2903b for the changes I made.

Because I took so long to get to this I simply did the changes myself and merged it but next time please be a bit more careful :D The lint error is whatever but the tests not even being able to run was pretty bad :|

lucaslsl:
sorry :|, next time i'll be more careful, hadnt actually finished/ran tests, it was more so you could check if i were in the right path (changing in the right places). glad you accepted/merged though, thanks a lot.

PR #495, opened 2019-03-08
add full_db_type

### LINKED ISSUE #494: full_db_type in postgres
First of all, let me say I love sqlboiler!

Then, I've been using it for a few projects with MySQL and great results.
I've decided to see if I can switch to postgres and then came an issue.

At generation, I'm using the full_db_type as a type.match, and it allows me to be quite precise on which field should be casted to what.
For example, I always use github.com/rs/xid as a primary key (id) in most of my tables.
I store this either as BINARY(12) or CHAR(20) in mysql, and the match is easy to do with full_db_type char(20) > xid.ID 

When I switched to postgresql, I realized the full_db_type is always empty, and the db_type doesn't include the max length of the character, and then it's impossible to differentiate the id char(20) from the other string ids (for example another country code table, with id as string).

I tried to use the default value as type.match, but it's not taken into account.
I also tried to generate a alternative "full_db_type" with udt_name and character_max_length but at this point of time, before submitting any PR, I'd like to have an opinion.

Specifically for psql:
would you accept a PR to match on the column default value?
would you accept a full_db_type with udt_name and char length?

@aarondl

vincentserpoul:
fixes #494


aarondl:
@vincentserpoul I tried this change `6b49f65 - (HEAD, vincentserpoul/master) add full_db_type (10 days ago)`

Rebuilt both sqlboiler and the driver but when I attempt to use it it doesn't seem to work. The output in `--debug` still shows `full_db_type` as "" for everything.

vincentserpoul:
So, I only set the full_db_type in case I have:

```
c.character_maximum_length != 0
```

But I can fill it for any type with the default udt_name instead of "" (empty string) if you prefer

aarondl:
Ugh, this got lost. I think it makes sense to let it duplicate. Let's go for consistency :D

vincentserpoul:
@aarondl

aarondl:
Merged in 3.4.0, thanks for the PR @vincentserpoul!

PR #514, opened 2019-04-16
Set \'1=0` when args is empty on WhereIn clause

izumin5210:
`WHERE ("a" IN ())` causes syntax error in postgresql.
in activerecord of rails/rails, it inserts `1=0` instead of empty IN clause when args is empty 

discusion: https://github.com/rails/rails/pull/9216
code https://github.com/rails/rails/blob/v5.2.3/activerecord/lib/active_record/relation/predicate_builder.rb#L69


aarondl:
Seems like it was a pretty poorly received change? Is this something we want?

PR #528, opened 2019-05-11
Rename variable *Columns to *AllColumns to allow a table named "updates"

### LINKED ISSUE #526: Table named "updates"
If my table is named "updates" (and I imagine "inserts" will have the same issue), sqlboiler generates code that doesn't compile.
This is because it generates `updateColumns` in two places.

1. As a var:
```updateColumns               = []string{...}```
This comes from the table name, I believe.

2. In the Upsert method there's an argument named updateColumns (and I imagine that one is hard-coded).

SQLBoiler v3.1.0

`sqlboiler -c ./sqlboiler.toml -o ./internal/models --wipe psql`

### PR DESCRIPTION
**Pull request by stuartgrigg**
I believe this fixes #526.
I have decided to change the dynamically named private variable generated for all models from *Columns to *AllColumns rather than renaming the updateColumns argument to Upsert etc. to avoid changing the argument names of any generated public methods. I have added a new table named "updates" to the test DBs to prevent regressions.

I have only tested this locally with Postgres - is there any CI for the other drivers?

### CODE REVIEW

**Comment by aarondl** [comment]
There is no CI for the other drivers currently. Permission issues with the org currently that's preventing people from attaching something that touches private repos with impunity. One day maybe we'll be fortunate enough that Github would bless us with a better permission model for OAuth2 but until then we might be stuck for a bit longer (we have some internal CI we might be cooking up soon).

As for the PR itself, I think it's great. Thanks. Merging now.

File: templates/00_struct.go.tpl [hunk 1]
```diff
@@ -19,7 +19,7 @@ type {{$alias.UpSingular}} struct {
 	{{end -}}
 }
 
-var {{$alias.UpSingular}}Columns = struct {
+var {{$alias.UpSingular}}AllColumns = struct {
```
aarondl: Spoke a bit too soon hah! This can't be changed, this is breaking and will affect all sqlboiler users that are currently referring to columns. Because this is an uppercase variant, I don't believe this needed to be changed in the first place. `UpdateColumns` and `updateColumns` cannot conflict. Will need this fixed before I can merge.

stuartgrigg:
Thanks for the review! I have restored the name of that variable - I don't think it is referenced anywhere else and it doesn't break the test I added. This might not have been the most pressing issue but I was looking for somthing that was staightforward.

aarondl:
Merged and in 3.4.0 thanks for the PR!

PR #538, opened 2019-05-23
Type replacement by PostgreSQL domain

### LINKED ISSUE #530: Type replacement by PostgreSQL domain
In my application I use PostgreSQL's `domain`s to define types with check conditions. For example
```sql
CREATE DOMAIN uint256 AS numeric(78, 0)
	CHECK (
		value >= 0
			AND value < 115792089237316195423570985008687907853269984665640564039457584007913129639936);
```
and use this domain in a lot of columns across the database. 
When it comes to generating the SQL, I would like to map this into a `big.Int` (or a wrapper of it, written by me), but this is not doable without replacing the whole `numeric` type, but we also use the generic `numeric` type for other columns.

I propose we add a `DomainName` field into `drivers.Column`, or modify the postgreSQL driver to provide the domain name as part of an existing field (`FullDBType`...?)

I believe the practice of using domains for refined types is common, so this is a valuable feature to add. Thank you!

natsukagami:
Resolves #530 

## Changes

- Added a new field `DomainName` in `drivers.Column`. This column should be `psql` only, and currently only the psql driver writes to it.
- The psql driver writes the `information_schema.columns`'s `domain_name` value in `DomainName`.
- Type replacement now accepts a `domain_name` parameter on the `match` side, which matches against the `DomainName` field.
- Bonus: fixed a bug where sqlboiler would crash when an `array_type` type match is used but matched against a column without one.


natsukagami:
The default value of the domain column in Postgres is the empty string `""`, which stands for no domain. I prefer it to have the NULL value instead of the empty string because I believe it's pretty close to the `ArrType` case (you either have a domain with a name, or have no domains at all and then a name is irrelevant)
However I do understand that `*string` is a bit annoying in Go code, so I'll leave the decision to you.

natsukagami:
@aarondl what do you think?

aarondl:
@natsukagami I agreed with your assessment so I merged it. Thanks so much for the PR :)

aarondl:
(Released in 3.4.0)

PR #539, opened 2019-05-23
Correct queries.Raw example

jrmycanady:
One queries.Raw example had an incorrect db parameter.


aarondl:
Merged, thanks for the PR!

PR #545, opened 2019-06-05
Fix mistype in README

dreadatour:


aarondl:
Merged into dev ce8a254. Thanks! Would that every language was like Go and enforced trailing commas :(

PR #549, opened 2019-06-12
Fix column order bug

tiger5226:
Added objectID position tracking. When the template is generated the ordering of conflicting unique columns is alphabetical. This means we cannot assume the objectID can always be appended. It might need to be inserted at a specific position. This adds support for tracking the position and inserting it at the correct position so the delete query correctly deletes the conflicts.


tiger5226:
grrr...sorry Aaron, I did it again...wrong repo

PR #550, opened 2019-06-15
fix: Invalid variable reference

cliedeman:


aarondl:
Thanks for the PR! Merged into dev. d974063

PR #554, opened 2019-06-22
Use new go-sqlmock import path

### LINKED ISSUE #553: go-sqlmock's canonical import path has changed from gopkg.in to github.com
`go-sqlmock`'s canonical import path was changed from `gopkg.in/DATA-DOG/go-sqlmock.v1` to `github.com/DATA-DOG/go-sqlmock` when they started supporting go modules.

Since `sqlboiler` still imports the library using the old path, projects using go modules to depend on `sqlboiler` will get messages about the "missing" dependency until a replace directive is added to point the old path to the new one, like:

```
replace gopkg.in/DATA-DOG/go-sqlmock.v1 => github.com/DATA-DOG/go-sqlmock v1.3.3
```

Not doing so doesn't have any major effects unless you do something like `go test all`, but I think it'd be a good idea to import the new recommended import path (and I'd be happier with one fewer `replace` to maintain).

Forgive my deletion of the template, but I didn't think it really fit.

zikaeroh:
Fixes #553.

The canonical/recommended import path for sqlmock has changed. Using the correct path makes the Go compiler happy in module mode.

```
$ go test ./queries                  
ok  	github.com/volatiletech/sqlboiler/queries	0.004s
```

Opened PR to the `dev` branch, as per the contributing docs.

### DIFF
--- a/queries/reflect_test.go
+++ b/queries/reflect_test.go
@@ -14,7 +14,7 @@ import (
 
 	"github.com/volatiletech/sqlboiler/drivers"
 
-	"gopkg.in/DATA-DOG/go-sqlmock.v1"
+	"github.com/DATA-DOG/go-sqlmock"
 )
 
 func bin64(i uint64) string {


aarondl:
Thanks for the PR! Yay, less gopkg.in!

PR #561, opened 2019-06-28
Add support for camelcase db tables

bugzpodder:
If a FK is TableId, be able to strip Id from the end.
Allow case sensitive Pluralize and Sigularize


aarondl:
I'm worried about a potential misuse of the inflect library here. As it's not my code I don't understand exactly how it works, but in general we've avoided feeding it any upper case input and I think that's probably for the best. As there are already so many rules it needs to look at each time it's evaluating a given word, adding another load like this doesn't seem like the right way to go. Thoughts?

bugzpodder:
inflect library is actually fairly straightforward.  It's looping through all the rules, and find any suffix matches and replace it with the pluralized version:
https://github.com/volatiletech/inflect/blob/master/inflect.go#L350

Also, I think Pluralize only affects codegen time instead of runtime so any perf concerns are contained.

The other approach is to add a flag for upper case vs lower case.  I thought about it first but there are quiet a bit of places where the conditional is needed and it wasn't completely obvious where I make the changes.  I could look into it if its preferred way..

aarondl:
If you wanted to do a flag you could consider a separate ruleset that could be activated by said flag that would work for upper/camelcase. If you look in boilingcore.go#New() there's no usage of strmangle in the binary up until around there. If a flag was to be added it could be switched on there and the new ruleset could be applied. This would require a new method in strmangle like `SetUppercaseRuleset()` or something.

Thoughts?

bugzpodder:
It's your call here :) But I think a different ruleset may make sense.  
Basically there are three options:
1) a flag should be added because it changes behavior for upper case schemas
2) a flag should be added because it potentially negatively impacts perf.
3) this change doesn't really affect lower case schemas, but improves support for upper case schemas, so we don't need the flag.

For 1), we should definitely add the flag.  For 2), we could potentially do a few tests on the timing on schema generation (my current schema takes a minute to generate so it might serve as a good benchmark).  For 3) my current PR should suffice.

bugzpodder:
timing on my system:
'''
time go run github.com/bugzpodder/sqlboiler --wipe --no-auto-timestamps --no-rows-affected --no-context --no-tests mysql

real	1m7.660s
user	0m2.540s
sys	0m0.765s

time go run github.com/bugzpodder/sqlboiler --wipe --no-auto-timestamps --no-rows-affected --no-context --no-tests mysql

real	1m5.840s
user	0m2.578s
sys	0m0.724s

time go run github.com/bugzpodder/sqlboiler --wipe --no-auto-timestamps --no-rows-affected --no-context --no-tests mysql

real	1m5.971s
user	0m2.566s
sys	0m0.717s

time go run github.com/volatiletech/sqlboiler --wipe --no-auto-timestamps --no-rows-affected --no-context --no-tests mysql

real	1m18.961s
user	0m2.784s
sys	0m0.775s
time go run github.com/volatiletech/sqlboiler --wipe --no-auto-timestamps --no-rows-affected --no-context --no-tests mysql

real	1m9.693s
user	0m2.700s
sys	0m0.783s
time go run github.com/volatiletech/sqlboiler --wipe --no-auto-timestamps --no-rows-affected --no-context --no-tests mysql

real	1m7.873s
user	0m2.534s
sys	0m0.742s
'''

aarondl:
I think we add the flag. Uppercase schemas are a second class citizen in sqlboiler and should remain as such. As well as it's a breaking change for any uppercase schemas that are currently working which I think is very serious.

bugzpodder:
I didnt have a chance to implement a flag, and I'm no longer working in Go so I'll leave it for someone else.

Issue #277, opened 2018-04-28
Test ToMany function fails due to unique column randomization not being adhered to


tiger5226:
If you're having a generation problem please answer these questions before submitting your issue. Thanks!

### What version of SQLBoiler are you using (`sqlboiler --version`)?

latest

### If this happened at generation time what was the full SQLBoiler command you used to generate your models? (if not applicable leave blank)

sqlboiler --no-auto-timestamps --no-hooks --tinyint-as-bool --wipe mysql

### If this happened at runtime what code produced the issue? (if not applicable leave blank)

N/A

### What is the output of the command above with the `-d` flag added to it? (Provided you are comfortable sharing this, it contains a blueprint of your schema)

No errors

### Please provide a relevant database schema so we can replicate your issue (Provided you are comfortable sharing this)

Can't share schema

### Further information. What did you do, what did you expect?

So if you look at the test code generation template file below

https://github.com/volatiletech/sqlboiler/blob/master/templates_test/relationship_to_many_setops.tpl#L19

The columns that it takes as input are the primary key columns, and columns without defaults. The problem here is that only the primary key is considered as being unique. It does not consider that columns other than the primary key might need to be unique. 

This obviously causes test case failures which impedes the auto generation capabilities if they have to be adjusted after each generation.

ceshihao:
Do you mean you have a non-primary-key and unique column with default value?

It seems not to be very reasonable.

tiger5226:
So I have 2 columns, the id (serial) and another address. The uniqueness of a row cannot be a composite primary key. Therefore the standard is to use a unique constraint in addition to the primary key in mysql...as far as I understand. Also this is common to improve query performance when you know a column is unique.

aarondl:
I mean, it doesn't matter what you're using the unique column for. In general we should be able to have unique columns and not have them die during testing. However what's confusing to me is that the whole purpose of the random seed is that it is an ever-increasing integer that acts as the seed for each column's value. Which means that unless this seed isn't being incremented, or the value range of your unique column is extremely constrained (like it's a boolean), it shouldn't be failing? Given this information, can you at least tell me the type of column that's failing so I can look into the problem in the `randomize` package?

aarondl:
It'd be helpful to know both the type and any constraints other than unique on it.

tiger5226:
[...]

PR #566, opened 2019-07-04
Remove periodicitly from picking string

marutanm:
This change may good effect to #277

In case define 1-many relationship table and child table has multiple varchar column, value of one of unique column are get duplicated.

```sql
CREATE TABLE `parents` (
  `id` INT NOT NULL AUTO_INCREMENT
)

CREATE TABLE `childs` (
  `id` INT NOT NULL AUTO_INCREMENT,
  `parent_id` INT NOT NULL,
  `foo` VARCHAR(255) NOT NULL,
  `bar` VARCHAR(255) NOT NULL,
  PRIMARY KEY (`id`),
  UNIQUE KEY `idx_foo` (`foo`),
  KEY `parent_id_idx` (`parent_id`),
  CONSTRAINT `ibfk` FOREIGN KEY (`parent_id`) REFERENCES `parents` (`id`)
)
```
generated test named `testParentToManyAddOpChilds` will fail like `Duplicate entry 'k' for key 'idx_foo'`

because the value set to VARCHAR column has periodicity.
```
b.Foo: "K"
c.Foo: "W"
d.Foo: "8"
e.Foo: "k"   # <= duplicated
seed: 1562208263
```
OR
```
b.Foo: "G"
c.Foo: "s"
d.Foo: "4"
e.Foo: "g"   # <= duplicated
seed: 1562208755
```
and so on...

Pick a random positions of word is enough to avoid this collision instead of using `nextInt()`.

### DIFF
--- a/randomize/random.go
+++ b/randomize/random.go
@@ -19,7 +19,7 @@ const alphabetLowerAlpha = "abcdefghijklmnopqrstuvwxyz"
 func Str(nextInt func() int64, ln int) string {
 	str := make([]byte, ln)
 	for i := 0; i < ln; i++ {
-		str[i] = byte(alphabetAll[nextInt()%int64(len(alphabetAll))])
+		str[i] = byte(alphabetAll[rand.Intn(len(alphabetAll))])
 	}
 
 	return string(str)
--- a/randomize/randomize.go
+++ b/randomize/randomize.go
@@ -276,7 +276,7 @@ func getVariableRandValue(s *Seed, fieldType string, kind reflect.Kind, typ refl
 		if ok {
 			return str
 		}
-		return Str(s.NextInt, 1)
+		return Str(s.NextInt, 2)
 	case reflect.Slice:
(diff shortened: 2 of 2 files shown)


aarondl:
The usage of `rand.Intn` here is undesirable. Pure randomness can fail the tests randomly, which is worse than them failing all the time with a specific seed value.

marutanm:
> Pure randomness can fail the tests randomly

Yes, that's right. Understood.
but test fails too much in my condition. It feels like 50% or more cases failing.
Is there any other way to avoid this without using `rand.Intn`...

Picked character with `nextInt()` from fixed string `alphabetAll` can be considered as pure random 🤔

aarondl:
Well, all we're doing is saying: get the next one in the list. So it's possible that it's just doing too many insertions in general into that table? Not sure. I'm opened to suggestions. Keep in mind that we chose 1 character to avoid hurting schemas with varchar limits etc. For things like country code or whatever.

Issue #570, opened 2019-07-09
whereHelper* should support IN


bugzpodder:
whereHelperstring, whereHelperint should support IN operator, which converts []string/[]int to []interface{}

aarondl:
Not opposed to this.

PR #575, opened 2019-07-13
implement whereIN for int and string

bugzpodder:
https://github.com/volatiletech/sqlboiler/pull/570

### DIFF
--- a/templates/00_struct.go.tpl
+++ b/templates/00_struct.go.tpl
@@ -47,6 +47,15 @@ func (w {{$name}}) LT(x {{.Type}}) qm.QueryMod { return qmhelper.Where(w.field, 
 func (w {{$name}}) LTE(x {{.Type}}) qm.QueryMod { return qmhelper.Where(w.field, qmhelper.LTE, x) }
 func (w {{$name}}) GT(x {{.Type}}) qm.QueryMod { return qmhelper.Where(w.field, qmhelper.GT, x) }
 func (w {{$name}}) GTE(x {{.Type}}) qm.QueryMod { return qmhelper.Where(w.field, qmhelper.GTE, x) }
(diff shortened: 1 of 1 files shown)


templates/00_struct.go.tpl
```diff
@@ -47,6 +47,15 @@ func (w {{$name}}) LT(x {{.Type}}) qm.QueryMod { return qmhelper.Where(w.field,
 func (w {{$name}}) LTE(x {{.Type}}) qm.QueryMod { return qmhelper.Where(w.field, qmhelper.LTE, x) }
 func (w {{$name}}) GT(x {{.Type}}) qm.QueryMod { return qmhelper.Where(w.field, qmhelper.GT, x) }
 func (w {{$name}}) GTE(x {{.Type}}) qm.QueryMod { return qmhelper.Where(w.field, qmhelper.GTE, x) }
+{{if or (eq .Type "int") (eq .Type "string") -}}
+func (w {{$name}}) IN(slice []{{.Type}}) qm.QueryMod {
+  values := make([]interface{}, 0, len(slice))
+  for _, value := range slice {
+    values = append(values, value)
+  }
+  return qm.WhereIn(fmt.Sprintf("%s IN ?", w.field), values...)
```
aarondl: This seems like it wouldn't work. You're passing multiple values into a single ? If it does work, I'd wonder if it's not dependent on a specific driver. Have you tested this with more than a single driver?

bugzpodder: I'm just using sqlboiler's API, see:
https://github.com/volatiletech/sqlboiler/blob/2942e962f90c880033128ab75f1b7e27badc3346/queries/query_builders_test.go#L334

aarondl:
Merged: 47826ba (PR to dev next time thanks!) And thanks for the PR

Issue #382, opened 2018-09-04
Move project to Go modules


andradei:
Are there plans to add a `go.mod` to the project? Maybe when the feature stops being not experimental?

aarondl:
Absolutely! In fact I already tried it in a local branch. However there are a few barriers and reasons to put it off.

1. Go modules is a horribly buggy feature at the moment. I say this with as much love as possible, I quite like the modules feature.
    https://github.com/golang/go/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+label%3Amodules
    I would like to see some of these bugs cleaned up before we start telling people to use Go modules with `sqlboiler`.

2. Right now there's still many Go versions (afaik <1.10.4 and <1.9.7, and yes I know 1.9 is deprecated but lots of people are still using it) in the wild that will not build the project correctly if they see semantic import versioning. This literally breaks users irreparably (outside forking or rollbacks) and is one of the biggest problems with the modules implementation as it stands. And you might say "well just add a go.mod file and don't do semantic import versioning" but that's a non-starter as you cannot opt out of semantic import versioning. If your project has a `go.mod` you must use SIV or it breaks.

3. One of the reasons to put this off is because Go modules will actually work fine with `sqlboiler` as is. For libraries that are >v2 that do not have a `go.mod` file, it gets pulled in as `v3.2.1+incompatible`. This tells Go that it's an old library (no SIV or `go.mod`) but still allows you to manage versions with go modules. This is the best of both worlds for now: No breaking slightly older Go version users, and users who are using modules get some of the benefits. The missing benefits being of course being that nobody knows what versions of it's dependencies it needs. But that's the world we lived in pre-modules anyway and it's been working for quite some time, we can afford to move a little slower so we don't break things.

I think because of 1, I'd like to hold off for a while, given that 3 allows people using modules with at least a small amount of benefit. One day we will definitely migrate to SIV + `go.mod` but I don't think today is the day.

Leaving this opened for other people to see/contribute to the conversation.

ccpost:
To add to (3) above a little bit: I'm currently on v2, and would *really* love to upgrade to v3, but keep having to put it off since it'd be a sweeping change across the entire codebase.  It seems with the Go modules "incompatible" way of doing this, there isn't any way to leverage semantic import versioning in Go modules such that I could have both versions of the sqlboiler models co-exist in the same codebase and do an incremental rather than wholesale upgrade.

[...]

PR #578, opened 2019-07-18
feat(module): enable Go module support and vendor dependencies

uroshercog:
Hello,

This pull request adds Go module support for the project. This is required for building and packaging the tool for installation using Nix. 

I've seen the issue #382, but I'm not sure if that issue is still relevant.

Best.


aarondl:
Still relevant. Have you seen the list of bugs in go modules? The feature barely works. The indexing servers and proxy servers are in complete infancy. Now is not the time. Trust me I'm the first one that wants to move there and will do so when the time is right :)

uroshercog:
Your answer raises a couple of questions. 
1. What do you mean with "the feature barely works"? Which issues are you implying to?
2. How are indexing and proxy servers related to adding some sort of dependency management to this project? 

Can you then add some other dependency management? Because we need locked dependencies (with hashes).

Thanks

aarondl:
@uroshercog There are still a lot of problems (see the issue list linked in the other issue) with basic dependency resolution, there are many edge cases that are not correctly handled. The go mod why also doesn't produce correct results frequently. Replace is half-baked doesn't work locally and there's even an issue where it's not documented.

It just needs a little bit more time to cook in my opinion. I won't be adding any other dependency management until Go modules hits. You can always fork and add it if necessary. sqlboiler is very much in a maintenance mode and releases are infrequent at this point.

Issue #546, opened 2019-06-08
Escaping parameters in an order-by clause


emwalker:
This is a request for information and a possible feature request.  I would like to be able to add an order-by clause with a parameter using the query builder DSL.  E.g.,

```golang
mods = append(
  mods,
  qm.OrderBy("(topics.name ilike ?) desc, topics.name", searchString),
)
```

Looking at the code on `master`, this doesn't seem to be supported yet.  Another possibility would be to be able to escape the string before it's passed to `OrderBy`:
```golang
mods = append(
  mods,
  qm.OrderBy(fmt.Sprintf("(topics.name ilike %s) desc, topics.name", qm.Escape(searchString)),
)
```

For this particular case, I would prefer to use the query builder rather than construct a raw query, as it makes it much easier to pick and choose what is to be added to the query.

Is there a recommended way of handling this situation?

aarondl:
Hmm. Without proper support built into the `queries` package this is not possible (sorry for that). It's very unlikely that `qm.Escape` will ever come to fruition because of how damned complicated escaping is for each of these databases (hence us relying on the Go mechanic of prepared statement sending to the DB server for actual escaping).

I would accept a PR to modify OrderBy to additionally accept a variadic argument list of args just like where does so they can properly be appended.

emwalker:
@aarondl I've made a pull request for this [here](https://github.com/volatiletech/sqlboiler/pull/593).  I assume changes will be needed.

ImVexed:
Was the previous merge request reverted? As of v4.5.0 `OrderBy` doesn't seem to accept parameters now :( https://github.com/volatiletech/sqlboiler/blob/v4.5.0/queries/qm/query_mods.go#L469

aarondl:
It seems like @emwalker may have forgotten to update the query mod, despite doing all the work to enable the arguments and I didn't pay close enough attention. Seems like `Having` and `GroupBy` could both be updated to take a variadic list of arguments now that we have the piping to allow them to function.

emwalker:
My apologies.  I ended up not using this functionality, so I didn't notice that the query mod needed to be updated as well.

Issue #589, opened 2019-08-10
OrderBy does not escape column name in PostgreSQL


parnic:
This may be a duplicate of #546, happy to add these comments there if so.

`OrderBy(columnName)` fails to escape the column name properly in a PostgreSQL database.

Table of sports games with a column holding the timestamp of the game (nullable) in a column named `when`. Code:
```
	games, err := models.Games(
		qm.WhereIn(fmt.Sprintf("%s in ?", models.GameColumns.ID), idList...),
		qm.OrderBy(models.GameColumns.When)).AllG()
```
generates query (example):
```
SELECT * FROM "games" WHERE ("id" IN ($1,$2,$3,$4,$5,$6,$7,$8,$9,$10,$11,$12)) ORDER BY when;
[1271 1272 1273 1274 1275 1276 1277 1278 1279 1280 1359 1369]
```
but I expected to see:
```
SELECT * FROM "games" WHERE ("id" IN ($1,$2,$3,$4,$5,$6,$7,$8,$9,$10,$11,$12)) ORDER BY "when";
[1271 1272 1273 1274 1275 1276 1277 1278 1279 1280 1359 1369]
```

### What version of SQLBoiler are you using (`sqlboiler --version`)?
SQLBoiler v3.4.0

### Please provide a relevant database schema so we can replicate your issue (Provided you are comfortable sharing this)
Schema of my "games" table (excluded some foreign keys for simplicity):
```
CREATE TYPE "games_status" AS ENUM ('not_started','in_progress','final','forfeit');
CREATE TYPE "games_game_type" AS ENUM ('exhibition','regular_season','playoff');
CREATE TABLE IF NOT EXISTS "games" (
  "id" serial,
  "facility_id" int DEFAULT NULL,
  "when" timestamp DEFAULT NULL,
  "created_at" timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
  "updated_at" timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
  "external_id" int DEFAULT NULL,
  "status" "games_status" NOT NULL DEFAULT 'not_started',
  "division_id" int NOT NULL,
  "game_type" "games_game_type" NOT NULL DEFAULT 'regular_season',
  "overtime" boolean NOT NULL DEFAULT '0',
  PRIMARY KEY ("id")
);
```

aarondl:
Let's ensure that the PR for #546 alleviates this pain.

parnic:
[...]

PR #593, opened 2019-08-18
Teach qm.OrderBy to accept parameters

emwalker:
This PR teaches qm.OrderBy to accept parameters.  See discussion [here](https://github.com/volatiletech/sqlboiler/issues/546#issuecomment-511080287).


main.go
```diff
@@ -18,7 +18,7 @@ import (
 
 //go:generate go-bindata -nometadata -pkg templatebin -o templatebin/bindata.go templates templates/singleton templates_test templates_test/singleton
 
-const sqlBoilerVersion = "3.4.0"
+const sqlBoilerVersion = "3.4.1"
```
aarondl: Version changes are done at release time and since this adds a feature it'd be a minor version not a patch :) Please remove

queries/query_builders.go
```diff
@@ -180,35 +180,41 @@ func buildUpdateQuery(q *Query) (*bytes.Buffer, []interface{}) {
...
+	}
+
+	var resp string
+	if q.dialect.UseIndexPlaceholders {
+		resp, _ = convertQuestionMarks(modBuf.String(), argsLen+1)
+	} else {
+		resp = modBuf.String()
+	}
+
+	fmt.Fprintf(buf, resp)
```
aarondl: Although I realize the original code had `fmt.Fprintf` I'm wondering why now. Should these not be `io.WriteString`?

queries/query.go
```diff
@@ -95,6 +95,58 @@ type with struct {
 	args   []interface{}
 }
 
+type orderBy struct {
+	clause string
+	args   []interface{}
+}
+
+type modifiers interface {
```
aarondl: I think this interface is overkill. If you look at the way it's being used it's only saving us from sending 2 arguments into the `writeParameterizedModifiers` function (as clauses() is duplicate code in all implementations). Unless there's further reasoning I think these types should be removed and simply pass keyword/delim into the function.

aarondl:
I've also noted that https://github.com/volatiletech/sqlboiler/issues/589 is talking about escaping the parameters in order by. While we're doing this is it possible that we can add this functionality as well?

emwalker:
> I've also noted that #589 is talking about escaping the parameters in order by. While we're doing this is it possible that we can add this functionality as well?

Since order-by clauses can be arbitrary expressions, there would need to be some way of parsing an expression in order to identify what to escape and what to leave unescaped:
```sql
order by lower(trim(name))
```

Or in the case of this PR:
```sql
order by name like ?
```
In these examples, only "name" would be escaped. Are there helpers in this project for parsing an expression and traversing the parse tree?

main.go
emwalker: Done.

queries/query_builders.go
emwalker: I'm still learning Go, so I don't yet know what is idiomatic.  But `io.WriteString` seems to work here.  Note that there's a lot of `fmt.Fprintf` statements elsewhere in the file.  Should these be picked up as well?

queries/query.go
emwalker: I've dropped the interface.

What I've replaced it with might also be a little weird.  Now there is a `orderByClauses` helper and a `havingClauses` helper in `query_builders.go` and a one-off struct called `clause`, a slice of which is passed as the third argument.  This was my attempt to avoid duplicating most of the code in `writeParameterizedModifiers`.  But you might have in mind something else.

aarondl:
This is good as is, thanks for your work on it! Will merge soon (tm) but clauseType is getting a rename and may be further proliferated throughout the query struct if it can be so it won't be immediate since I need a bit of time to deal with it.

gaiottino:
When will this make it into a release?

aarondl:
@gaiottino Releases are done based on "feelings" rather than a schedule unfortunately. Looking at the changelog its getting to be about that time though.

PR #611, opened 2019-09-26
Conditionally disable parsing of templates provided by the driver

kurt-stolle:
For the case when templates for a language other than Go are generated, it is useful to have a configuration parameter that stops templates provided by the driver from being parsed. 

For example, I want to generate a very simple data structure in TypeScript for the front-end of my webapp. The template is as follows (adapted from Go template):
```
// 00_class.ts.tpl

export default class {{$alias.UpSingular}}Model {
{{range $column := .Table.Columns}}
    public {{$column.Name}};
{{- end }}
}
```

Without the proposed changes, I cannot complete the generation of the class because a driver template throws the following error when running `sqlboiler psql`: 
```
Error: unable to generate output: failed to execute template: templates\17_upsert.go.tpl: template: templates\17_upsert.go.tpl:39:14: executing "templates\\17_upsert.go.tpl" at <{{template "timestamp_upsert_helper" .}}>: template "timestamp_upsert_helper" not defined
```

For my use case it makes sense to use the psql driver, but having it also output go files in a typescript project is a pointless effort.

### DIFF
--- a/README.md
+++ b/README.md
@@ -323,6 +323,7 @@ not to pass them through the command line or environment variables:
 | no-tests            | false     |
 | no-auto-timestamps  | false     |
 | no-rows-affected    | false     |
+| no-driver-templates | false     |
 
 Example:
 
--- a/boilingcore/boilingcore.go
+++ b/boilingcore/boilingcore.go
@@ -133,23 +133,24 @@ func New(config *Config) (*State, error) {
 // state given.
 func (s *State) Run() error {
 	data := &templateData{
-		Tables:           s.Tables,
-		Aliases:          s.Config.Aliases,
-		DriverName:       s.Config.DriverName,
-		PkgName:          s.Config.PkgName,
-		AddGlobal:        s.Config.AddGlobal,
-		AddPanic:         s.Config.AddPanic,
-		NoContext:        s.Config.NoContext,
-		NoHooks:          s.Config.NoHooks,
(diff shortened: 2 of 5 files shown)


aarondl:
I considered asking for a list of templates as a sort of blacklist instead of a complete on/off, but I think if you want to customize what templates are used then it's acceptable to have to recompile the driver and the bindata.

As such, I think this is the first use case where I'm seeing non-Go being used? Would you mind detailing your use-case just a little so I can learn how you're using sqlboiler? Would be much appreciated.

Thanks for the PR, merged into dev.

kurt-stolle:
I am using sqlboiler for an e-commerce project. The server/api is written in Go, and I use sqlboiler here in the way you would expect for a Go project. In essence, the application is about handling orders, payments and shipping details over HTTP requests. 

The client/website scripts are written in TypeScript. Communications between client and server often involve sending a JSON-structure derived from a model generated by SQLBoiler. In order to effortlessly parse and verify data received from the server, and compose data structures going to the server, I have created a sqlboiler template that outputs TypeScript classes. Doing so has greatly increased the efficiency of the development process.

Hope this helps.

aarondl:
@kurt-stolle Wonderful to hear that someone's making good use of sqlboiler's ability to generate more than just Go code. This PR will be closed shortly when I push the changes I'm working on. Thanks again!

PR #613, opened 2019-10-05
Sample fix for issue 404

ericlagergren:
I noticed somebody else ran into this snag, so I decided it might be a good idea to re-look at the problem.

Unfortunately, there's not much I can do on my end. There are trade offs between what other packages do:

```
type T { x *big.Int }
func (x T) Add(y T) T { ... }
````

and what my package does:

```
type T { x big.Int }
func (z *T) Add(x, y *T) *T { ... }
```

One of these is the inability to unmarshal into a nil pointer.

So, this is just an example of one way to fix the problem. I'm sure there are others. But if this is continuing to affect folks like in https://github.com/ericlagergren/decimal/issues/141, it might be good to re-open the conversation.

### DIFF
--- a/types/decimal.go
+++ b/types/decimal.go
@@ -28,6 +28,13 @@ type NullDecimal struct {
 	*decimal.Big
 }
 
+func (d *NullDecimal) UnmarshalJSON(data []byte) error {
+	if d.Big == nil {
+		d.Big = new(decimal.Big)
+	}
+	return d.Big.UnmarshalJSON(data)
+}
+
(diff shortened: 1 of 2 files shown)


aarondl:
I appreciate this. Not sure why I didn't think of this as a basic workaround. Thanks. I've taken the code and rewritten the test (panicing tests are fine imho) and also applied the same logic to Decimal not just NullDecimal. I'll credit you with the fix in the changelog however :)

ericlagergren:
Happy to help!

PR #615, opened 2019-10-09
Make port configurable via env by parse string to int

letientai299:
Before this commit, following bash script won't work

```
  PSQL_DBNAME="postgres" \
    PSQL_HOST="localhost" \
    PSQL_PORT="11432" \
    PSQL_USER="postgres" \
    PSQL_PASS='' \
    PSQL_SCHEMA="$schema" \
    PSQL_SSLMODE="disable" \
    sqlboiler --wipe --no-tests -o "internal/model/db/$schema" -p "$schema" \
    --templates ./templates/sqlboiler/ psql
```

`PSQL_PORT` is passed correctly, but failed to parse due to its string type. sqlboiler use default port 5432, which is undesired result.

This commit fix that.

### DIFF
--- a/drivers/config.go
+++ b/drivers/config.go
@@ -47,7 +47,11 @@ func (c Config) MustInt(key string) int {
 	case float64:
 		integer = int(t)
 	case string:
-		integer, _ = strconv.Atoi(t)
+		var err error
+		integer, err = strconv.Atoi(t)
+		if err != nil {
+			panic(errors.Errorf("fail to parse %v to int: %v", t, err))
+		}
 	default:
 		panic(errors.Errorf("found key %s in config, but it was not an int (%T)", key, i))
 	}
@@ -104,7 +108,11 @@ func (c Config) Int(key string) (int, bool) {
 	case float64:
 		integer = int(t)
 	case string:
-		integer, _ = strconv.Atoi(t)
+		var err error
+		integer, err = strconv.Atoi(t)
+		if err != nil {
+			return 0, false
+		}
 	default:
 		return 0, false
 	}
--- a/drivers/config.go
+++ b/drivers/config.go
@@ -5,6 +5,7 @@ package drivers
 
 import (
 	"os"
+	"strconv"
 	"strings"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


drivers/config.go
```diff
@@ -45,6 +46,8 @@ func (c Config) MustInt(key string) int {
 		integer = t
 	case float64:
 		integer = int(t)
+	case string:
+		integer, _ = strconv.Atoi(t)
```
aarondl: Not keen on ignoring the errors here. If this error occurs then the key was set by a user that expects it to parse correctly and become an integer. If that doesn't work because there's a type/some other problem we should have the program fail I think? That's in essence the entire point of MustInt, it -HAS- to be an int or it should fail.

drivers/config.go
```diff
@@ -100,6 +103,8 @@ func (c Config) Int(key string) (int, bool) {
...
```
aarondl: Ignoring the error here is worse than above because we have a way to signal failure in this method (the bool return). If the parsing fails, then your port will inexplicably be 0 without any failure notifications.

letientai299: Fixed in 6722d57ab31fffadaf70e219e0deb7d4ac4406c8

drivers/config.go
letientai299: Thanks, my bad. I've fixed this issue in 6722d57ab31fffadaf70e219e0deb7d4ac4406c8

letientai299:
@aarondl I've fixed the review comment, but I don't know how to properly resolve the 
"Change requested" as show in Github PR.

letientai299:
Hi @aarondl, any thought?

aarondl:
The "Change Requested" is a code review status and cannot be changed by the contributor, only a member of the repository who has the ability to perform code reviews :)

This is good enough, although I will change the `%v` to a `%s` where it is appropriate.

Merged to dev: d32ac34

aarondl:
Note the PR is closed but was accepted and merged, in our CONTRIBUTING.md it says specifically to open PRs against dev branch and so it doesn't show the merge correctly.

PR #617, opened 2019-10-11
Some code improvements

sosiska:
* strings.ReplaceAll instead of strings.Replace with -1.
* Change strings.ToLower comparison to strings.EqualFold.
* Simplify some functions.
* Rewrite switch statements with only one case as if.
* Some assignments simplified by using assignment operators.
* Put missing space between `//` and comment text.

### DIFF
--- a/boilingcore/boilingcore.go
+++ b/boilingcore/boilingcore.go
@@ -589,13 +589,13 @@ func mergeTemplates(dst, src map[string]templateLoader) {
 // normalizeSlashes takes a path that was made on linux or windows and converts it
 // to a native path.
 func normalizeSlashes(path string) string {
-	path = strings.Replace(path, `/`, string(os.PathSeparator), -1)
-	path = strings.Replace(path, `\`, string(os.PathSeparator), -1)
+	path = strings.ReplaceAll(path, `/`, string(os.PathSeparator))
+	path = strings.ReplaceAll(path, `\`, string(os.PathSeparator))
 	return path
 }
 
 // denormalizeSlashes takes any backslashes and converts them to linux style slashes
 func denormalizeSlashes(path string) string {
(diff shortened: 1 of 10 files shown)


aarondl:
Thanks for this. Was this done mechanically or manually?

Also despite using Go since v1 I had no idea EqualFold existed, has that existed since the beginning as I couldn't find it in the release notes anywhere.

sosiska:
@aarondl it was linter go-critic

lucaslsl:
@aarondl `strings.ReplaceAll` was added on `Go 1.12` if i'm not mistaken, so this breaks anything below that. I figured the hard way by naively not locking the cli and having my CI builds fail.

aarondl:
Thanks for the heads up. This is fine by me. Supporting 1 major version behind is what I strive for. Sorry to hear you got broken!

Any particular reason you're still using an older Go?

lucaslsl:
No really, in fact I just upgraded to `1.13`. It's a project that hadnt be touched in a while.

touyu:
I'm using GAE/SE go1.11.
GAE/SE becomes 2nd Gen from go1.12,  and there is a big destructive change, so it is difficult to raise Go version for me.
I want you to consider supporting go1.11.

touyu:
@aarondl
I'm using GAE/SE go1.11.
GAE/SE becomes 2nd Gen from go1.12, and there is a big destructive change, so it is difficult to raise Go version for me.
I want you to consider supporting go1.11.

aarondl:
I apologize touyu but I have no interest in maintaining older versions. Go 1.11 is over a year old now. Just like your Go version is pinned to something older you can take the same approach for sqlboiler, we tag all of our releases so it should be easy to use it at an older version.

PR #621, opened 2019-10-15
Add Dockerfile

dahu33:
This is part of #620

This pull adds a Dockerfile that builds a docker image for the 3 drivers available. 

If needed I can split this into 3 Dockerfiles to create one docker image per driver.

PR #623, opened 2019-10-16
Dockerfile and hooks to automatically build images Docker Hub

### LINKED ISSUE #620: sqlboiler docker image?
Would be nice to publish "official" docker images on hub.docker.com. Happy to help if needed.

dahu33:
This PR fixes #620 and supersede #621.

Some notes:

The `Dockerfile` uses multi-stage builds to create a sqlboiler docker image with all the drivers included. It can also be used to create docker images with only a specific driver included.

The `hooks` folder contains hooks to automatically build images on hub.docker.com (see: https://docs.docker.com/docker-hub/builds/advanced/).

The `build` hook will build 4 images:
- the main image with all the drivers included
- mssql image
- mysql image
- psql image

The `push` hook will tag/push the images to the repository. For example for the git tag `v3.5.0`:
- main image tags:  `latest`, `3`, `3.5`, `3.5.0`
- mssql image tags: `mssql`, `3-mssql`, `3.5-mssql`, `3.5.0-mssql`
- mysql image tags: `mysql`, `3-mysql`, `3.5-mysql`, `3.5.0-mysql`
- psql image tags: `psql`, `3-psql`, `3.5-psql`, `3.5.0-psql`

The only configuration required on docker hub to automatically build images as described above is the following (note: the docker tag field is ignored by the hook scripts):

hub docker com

The result can be seen here (if the link doesn't work, just search for sqlboiler image under the curvegrid organization): https://hub.docker.com/r/curvegrid/sqlboiler/tags?page=1&ordering=-name

@aarondl would you mind reviewing and setting up automatic builds on Volatiletech's docker hub account? 

Thanks!


.dockerignore
```diff
@@ -0,0 +1,2 @@
+Dockerfile
+hooks
```
aarondl: Git complains about not having newlines at EOF, I suggest you change your editor settings if you primarily work with git.

Dockerfile
```diff
@@ -0,0 +1,50 @@
+# builder
+FROM golang:1.13-alpine AS builder
+
+RUN apk add git
+
+WORKDIR /go/src/github.com/volatiletech/sqlboiler
+
+COPY . .
+RUN go get -v -t ./...
+RUN go build -ldflags "-w -s" . && \
```
aarondl: You could consider using `-trimpath` in your build statements

hooks/push
```diff
@@ -0,0 +1,60 @@
...
+
+    # single driver images
+    for driver in $DRIVERS; do
+        tagAndPush "sqlboiler-${driver}" "${DOCKER_REPO}:${version}-${driver}"
+    done
+}
+
+# if not a prerelease
+if [[ "${prere}" == "" ]]; then
+    tagAndPushVersion "${major}"
```
aarondl: Just asking the question: Given that the entire idea is to lock users to a specific version of the tool else they will likely get different generated outputs and trash against each other in git, is this tagging scheme something we want to support?

It seems much like the Go modules decision of "always try to get the latest X" versus "get the latest required". If you go with the former you can pull changes you didn't expect/want because you didn't get a chance to look at them and make a decision to upgrade. If you do the latter you have to remember to decide to upgrade.

For sqlboiler, what makes the most sense?

hooks/push
```diff
@@ -0,0 +1,60 @@
+#!/bin/bash
+set -ex
+
+LASTEST_MAJOR="3"
```
aarondl: This could avoid being hardcoded (since we are guaranteed to forget this exists) by doing something like:

```bash
git tag --list | tr --delete 'v' | sort --general-numeric-sort | tail --lines 1 | \
   grep --extended-regex --only-matching '^[0-9]+'
```

.dockerignore
dahu33: Fixed both my editor and this file :)

hooks/push
dahu33: Good point!

hooks/push
dahu33: Personally, I would leave this decision to the user. It's fairly common practice in docker hub to tag each major version and it's up to the user to decide what's the best for their environment. Some user could for example want to have their CI always using the latest v3 to make sure there isn't any incompatibility, and I'm sure there is a lot of use-cases that I cannot think about.

dahu33:
@aarondl thank you very much for your review and sorry for the delay... I've addressed your feedback, please have a look.

stephenafamo:
Since we now use GitHub actions, this PR would have to be modified considerably before it can be merged. And since the last update was over two years ago, I'll close this for now.

@dahu33, if you're still up for it, perhaps you can make the changes and create a new PR.

PR #624, opened 2019-10-19
Add support for go 1.13 errors

jwilner:
Switch to using github.com/friendsofgo/errors, a fork of github.com/pkg/errors, in order to provide error types which are compatible with _both_ github.com/pkg/errors and golang.org/pkg/errors

When used with go 1.13 or higher, this will use the builtin errors package with its new features; when used with older versions of go, it will use golang.org/x/xerrors, the backport of these features. 

See: 
- https://golang.org/pkg/errors/
- https://blog.golang.org/go1.13-errors

Relevant discussion:
- https://mycodesmells.com/post/migrating-pkg-errors-to-go-113-errors

### DIFF
--- a/boilingcore/boilingcore.go
+++ b/boilingcore/boilingcore.go
@@ -11,7 +11,7 @@ import (
 	"sort"
 	"strings"
 
-	"github.com/pkg/errors"
+	"github.com/friendsofgo/errors"
 	"github.com/volatiletech/sqlboiler/drivers"
 	"github.com/volatiletech/sqlboiler/importers"
 	"github.com/volatiletech/sqlboiler/strmangle"
--- a/boilingcore/output.go
+++ b/boilingcore/output.go
@@ -13,7 +13,7 @@ import (
 	"strings"
 	"text/template"
 
-	"github.com/pkg/errors"
+	"github.com/friendsofgo/errors"
 	"github.com/volatiletech/sqlboiler/importers"
 )
 
--- a/boilingcore/templates.go
+++ b/boilingcore/templates.go
@@ -11,7 +11,7 @@ import (
 	"strings"
 	"text/template"
 
-	"github.com/pkg/errors"
+	"github.com/friendsofgo/errors"
 	"github.com/volatiletech/sqlboiler/drivers"
 	"github.com/volatiletech/sqlboiler/strmangle"
 	"github.com/volatiletech/sqlboiler/templatebin"
--- a/drivers/binary_driver.go
+++ b/drivers/binary_driver.go
@@ -7,7 +7,7 @@ import (
 	"os"
 	"os/exec"
 
-	"github.com/pkg/errors"
+	"github.com/friendsofgo/errors"
 	"github.com/volatiletech/sqlboiler/importers"
 )
 
--- a/drivers/config.go
+++ b/drivers/config.go
@@ -7,7 +7,7 @@ import (
 	"os"
 	"strings"
 
-	"github.com/pkg/errors"
+	"github.com/friendsofgo/errors"
 )
 
 // Config is a map with helper functions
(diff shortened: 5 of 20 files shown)


aarondl:
I hate this change for so many reasons. Why couldn't they have just finished the stack trace printing support proposals before implementing this half solution. Now we have to accept a PR to move to this new library that migrates us halfway there, only to one day (hopefully) do another PR to remove it entirely and re-write all the error statements to the somewhat appalling `fmt.Errorf(`%w`)` (why a fmt package is required for wrapping an error is beyond me).

Before I can accept this however I have to ask the question: For current users of `sqlboiler` who may be using pkg/errors to `.Cause()` does it still work with the old package? If not this is a breaking change (requires downstream code to migrate to `friendsofgo/errors`) and I cannot accept it until we roll v4.

aarondl:
Sorry about my wording by the way, I'm certainly not against the PR and I do appreciate you having worked on it. I simply dislike the situation and how both the Go team and Dave Cheney has handled things in relation to this library. As well as Go 1.13s error handling being both somewhat odd and lacking features the community was asking for from the beginning (stack printing).

jwilner:
Haha no worries, I hear you. I 100% agree it's an awkward situation. Thanks
for considering the PR and, by the way, for a brilliant library. I've
brought SQLboiler to a couple of teams and projects at this point and it's
always popular and a real boon to productivity.

The returned errors are compatible with both patterns -- you can use `Is`,
`As`, `Cause`, etc as expected. It should be completely backwards
compatible with Go versions before 1.13, too -- build tags have the lib
fall back to xerrors for earlier versions.

aarondl:
Merged. Thanks for this.

zikaeroh:
Unfortunately people are going to break with this due to friendsofgo/errors#2, since I guess this alternative errors repo decided to force push and break sums... Fun times.

aarondl:
We already had it reported to us. How frustrating. "friendsofgo" my ass. Pardon the language.

We'll move away at the earliest convenience or sooner if this library remains problematic. At least it proves go mod sums work as intended.

Has anyone verified the contents of the tag to ensure nothing malicious is in there? That should be a priority. No left pad please.

zikaeroh:
Diff from `pkg/errors` here (commits from v0.9.1 to master are readme-only updates): https://github.com/pkg/errors/compare/master...friendsofgo:master

I see nothing suspicious, though I really do not like how this particular lib decided to create compatibility by writing the same file twice, one with a build tag for 1.13 and one without.

jwilner:
That's unfortunate about the force push, can't very well defend that. Hopefully they'll respond quickly.

In general, I stand by this library, which I researched beforehand. They need to use a `xerrors` for versions earlier than 1.13, so I think the build tag approach seems pretty reasonable. I specifically like this package versus other ones because it's not increasing the surface area of the pkg/errors API; other alternatives try to add functionality. More generally, I haven't seen one of these libraries that is markedly more popular than the others, although of course I can't say I've seen them all.

zikaeroh:
Sure, if the code is fine, then that's all good. So long as differing libraries respect the interfaces, it doesn't matter too much. Anyone can make a mistake by force pushing in this post-module world...

Seems like the repo author has tagged a new version, so we'll see if that makes this all go away. (I'm away from my PC so can't try it out just yet, but it'll likely be fine.)

aperezg:
@aarondl sorry for the inconvenience that occurred while using our library, I understand your point of view completely. There are not any malicious code in the library we only give an alternative to continue using the very extending `pkg/errors` library with the new errors. The problem was, as @jwilner comments a push force into the tag, we know that this kind of thing are completely forbidden, but our library in this moment had a very few time of life and we create the review 0.9.1 leaving a silly change behind, so we pushed force the tag almost at the same time, and we thought that would not have collateral damage.

So, very sorry for that, seriously, we will never do it again, and thank you for the opportunity.
Any questions, suggestions or whatever you need let me know.

PR #625, opened 2019-10-22
Expand support for whereHelperXXX.IN helper method

### LINKED ISSUE #609: Expand support for whereHelperXXX.IN helper method
Release v3.5.0 added support for wherein helpers for `int` and `string` types (via PR #575).  In the relevant template, this is implemented via the `qm.WhereIn` function, which takes some number of `interface{}` arguments as the values to match against.  My understanding is that any type can be passed so long as it is either one of the base types supported by `database/sql` or it implements the `driver.Valuer` interface.  At generation-time I realize it may not be simple to determine if a field's type implements the `driver.Valuer` interface, but would it be possible to at least have sqlboiler's template generate `whereHelperXXX.IN` methods for any of the base types that `database/sql` supports?  I believe the base types supported by `database/sql` are documented here:

https://godoc.org/database/sql#Rows.Scan

I ran into this issue when I read the release notes for v3.5.0 and tried to use the new `whereHelperXXX.IN` method for my `int64` field only to discover that it is only generated for fields with type `int`.

I'm happy to create a PR for this once it's been accepted and any details ironed out.

nwidger:
Expand support for `whereHelperXXX.IN` helper methods by generating a
`whereHelperXXX.IN` method for any type `XXX` where `isPrimitive XXX`
returns true.  Currently, `isPrimitive` returns true for the following
types:

    int int8 int16 int32 int64
    uint uint8 uint16 uint32 uint64
    float32 float64
    byte rune string

Previously, `whereHelperXXX.IN` helper methods were only generated for
fields of type `string`.

Fixes #609


aarondl:
Thanks for this @nwidger! Simplest PR I ever reviewed. Sorry it took me so long to get to it. I do my open source in bursts :)

nwidger:
Thank you @aarondl sqlboiler is a fantastic tool!  Any idea when this will be going into a new release?

aarondl:
@nwidger The typical flow is I just sort of let it sit in dev while people on the edge test it, and then when there's enough changes I bundle them up. I could consider doing another release when the docker stuff lands so I don't have to publish as many of those.

nwidger:
@aarondl I figured your process was something along those lines.  Sounds good, I'll keep my eye out for the next release.  Thanks again!

PR #643, opened 2019-11-30
Add usage/access example to Eager Loading in readme.

tooolbox:
From [looking at the Readme](https://github.com/volatiletech/sqlboiler#relationships), it was unclear how to actually *get* the related structs I had asked for using `qm.Load`.  It was only when I did some searching and newly looked at [a small taste](https://github.com/volatiletech/sqlboiler#a-small-taste) that I realized where they were.

This PR improves the documentation under the Relations/Eager Loading section so it becomes clear where the related structs are loaded to.


aarondl:
Thanks for the PR!

PR #646, opened 2019-12-11
Ignore tags values on a struct fields during parsing

bogdanprodanj:
Added a possibility to provide a list of table columns so they can be ignored during parsing, i.e. all the tags except `boil` will have "-" as a tag key. This can be used when models generated by `sqlboiler` is used on different layers of web application and some fields should be omitted explicitly (like `created_at`, `updated_at`, etc.) without the need for the creation of additional high-level models and converters.


README.md
```diff
@@ -324,6 +324,7 @@ not to pass them through the command line or environment variables:
 | no-auto-timestamps  | false     |
 | no-rows-affected    | false     |
 | no-driver-templates | false     |
+| ignore-fields       | []        |
```
aarondl: The naming of this is ambiguous. It must somehow refer to struct tags.

templates/00_struct.go.tpl
```diff
@@ -1,10 +1,14 @@
...
+{{- $orig_tbl_name := .Table.Name -}}
 
 // {{$alias.UpSingular}} is an object representing the database table.
 type {{$alias.UpSingular}} struct {
 	{{- range $column := .Table.Columns -}}
 	{{- $colAlias := $alias.Column $column.Name -}}
-	{{if eq $.StructTagCasing "title" -}}
+	{{- $orig_col_name := $column.Name -}}
+	{{if ignore $orig_tbl_name $orig_col_name $.IgnoreFields -}}
+	{{$colAlias}} {{$column.Type}} `{{generateIgnoreTags $.Tags}}boil:"{{$column.Name}}" json:"-" toml:"-" yaml:"-"`
```
aarondl: I don't see a function in this PR named generateIgnoreTags. How does this work?

bogdanprodanj: It was already defined, I just reused it:
https://github.com/volatiletech/sqlboiler/blob/5a51e7228cba48f4f40e542504540fcd8b14b26b/templates/00_struct.go.tpl#L17
https://github.com/volatiletech/sqlboiler/blob/5a51e7228cba48f4f40e542504540fcd8b14b26b/strmangle/strmangle.go#L631

README.md
bogdanprodanj: I can rename it to `tag-ignore` and put a better description of usage.

aarondl:
Merged into dev: 0223fea, next time please create the change against dev branch as state in contributor guidelines. Thanks for the PR! Sorry it took so long to get to.

PR #656, opened 2020-01-17
use (FALSE) instead of empty (column IN())

jackdoe:
when you pass empty slice to WhereIn, it generates `WHERE IN ()` which is invalid SQL, so it is difficult to simply run code
like:

for _, u := range model.Users(qm.WhereIn("id IN ?",uids...)).AllP(db) {
   ...
}
this patch generates `WHERE (FALSE)` instead of `WHERE column IN ()` so it is still chainable and there is no need to constantly check `if len(uids) > 0`

### DIFF
--- a/queries/query_builders.go
+++ b/queries/query_builders.go
@@ -371,10 +371,10 @@ ManualParen:
 			// for _, u := range model.Users(qm.WhereIn("id IN ?",uids...)).AllP(db) {
 			//    ...
 			// }
-			// instead when we see empty IN we produce FALSE so it can still be chained
+			// instead when we see empty IN we produce 1=9 so it can still be chained
 			// with other queries
 			if ln == 0 {
-				buf.WriteString("(FALSE)")
+				buf.WriteString("(1=0)")
 			} else {
 				matches := rgxInClause.FindStringSubmatch(where.clause)
 				// If we can't find any matches attempt a simple replace with 1 group.
--- a/queries/query_builders.go
+++ b/queries/query_builders.go
@@ -375,62 +375,63 @@ ManualParen:
 			// with other queries
 			if ln == 0 {
 				buf.WriteString("(1=0)")
-			} else {
-				matches := rgxInClause.FindStringSubmatch(where.clause)
-				// If we can't find any matches attempt a simple replace with 1 group.
-				// Clauses that fit this criteria will not be able to contain ? in their
-				// column name side, however if this case is being hit then the regexp
-				// probably needs adjustment, or the user is passing in invalid clauses.
-				if matches == nil {
-					clause, count := convertInQuestionMarks(q.dialect.UseIndexPlaceholders, where.clause, startAt, 1, ln)
-					if !manualParens {
-						buf.WriteByte('(')
-					}
-					buf.WriteString(clause)
-					if !manualParens {
-						buf.WriteByte(')')
-					}
-					args = append(args, where.args...)
-					startAt += count
-					break
-				}
-
-				leftSide := strings.TrimSpace(matches[1])
-				rightSide := strings.TrimSpace(matches[2])
-				// If matches are found, we have to parse the left side (column side)
-				// of the clause to determine how many columns they are using.
-				// This number determines the groupAt for the convert function.
-				cols := strings.Split(leftSide, ",")
-				cols = strmangle.IdentQuoteSlice(q.dialect.LQ, q.dialect.RQ, cols)
-				groupAt := len(cols)
-
-				var leftClause string
-				var leftCount int
-				if q.dialect.UseIndexPlaceholders {
-					leftClause, leftCount = convertQuestionMarks(strings.Join(cols, ","), startAt)
-				} else {
-					// Count the number of cols that are question marks, so we know
-					// how much to offset convertInQuestionMarks by
-					for _, v := range cols {
-						if v == "?" {
-							leftCount++
-						}
-					}
-					leftClause = strings.Join(cols, ",")
-				}
-				rightClause, rightCount := convertInQuestionMarks(q.dialect.UseIndexPlaceholders, rightSide, startAt+leftCount, groupAt, ln-leftCount)
+				break
+			}
+			matches := rgxInClause.FindStringSubmatch(where.clause)
+			// If we can't find any matches attempt a simple replace with 1 group.
+			// Clauses that fit this criteria will not be able to contain ? in their
+			// column name side, however if this case is being hit then the regexp
+			// probably needs adjustment, or the user is passing in invalid clauses.
+			if matches == nil {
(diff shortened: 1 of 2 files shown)


aarondl:
@jackdoe It's a bit of a contentious pull request. The reason being that that is indeed invalid sql and in the event that you're not chaining anything then there's actually no reason whatsoever to send this query to the database server because it can never return anything.

The reasoning for not doing anything about it was basically that it's always wrong to send this query. And in the case when you're chaining, then it's incorrect to have it as a condition since it's value is predetermined.

Thoughts?

jackdoe:
@aarondl thanks for your reply!

I agree that if it is only 1 where clause there is no need to send the query, but in many cases you have chained ORs that will produce results and the empty IN is just one of them, so it makes it really nice to not think about it. e.g. 2 criteria `'(filtersA IN (a,b,c)) OR (filterB in (d,e,f))'` , or the very common `where ... AND id NOT IN (modified_ids)`

I will give few examples how this is handled in other query builders:

* sqlx (https://github.com/jmoiron/sqlx)
It will generate error "empty in condition" ~same as sqlboiler just with more readable error

* JOOQ (https://github.com/jOOQ/jOOQ)
Empty Collection is transformed into falseCondition(), so it behaves the same way as the `(false)` I propose.

* GORM (https://github.com/jinzhu/gorm)
In GORM you will get `'WHERE x IN (NULL)'` for empty slice, which i believe is worse, considering that is not what the user intended at all.

```
package main

import (
	"github.com/jinzhu/gorm"
	_ "github.com/jinzhu/gorm/dialects/sqlite"
)

type Product struct {
	gorm.Model

	Code  string
	Price uint
}

func main() {
	db, err := gorm.Open("sqlite3", "test.db")
	if err != nil {
		panic(err)
	}
	defer db.Close()

	db.LogMode(true)
	db.AutoMigrate(&Product{})

	products := []Product{}

	_ = db.Where("id IN (?)", []string{}).Find(&products)
}
```
prints 

```
[2020-01-21 08:44:05]  [0.07ms]  SELECT * FROM "products"  WHERE "products"."deleted_at" IS NULL AND ((id IN (NULL)))
```

* knex (https://github.com/knex/knex)
Knex makes `1=0` condition (same as false) http://knexjs.org/#Builder-whereIn

* active record (https://github.com/rails/rails/tree/master/activerecord)
same as knex, it makes `1=0`

In general it is more common than not in the query builders to handle the empty IN() clause as it produces annoying runtime errors due to developers just forgetting to check boundary conditions, and the cost of running an empty (in some cases) query is negligible.

From all the options I think `(FALSE)` or `(1=0)` are the best substitutions for empty `IN()` as it does not do anything unexpected, considering the alternatives:

* in (NULL)
as in it will not find unwanted results like in GORM,

* removing the clause altogether
you can delete your whole table (e.g. `delete from table where in()` -> `delete from table`), 

* throwing error
runtime errors from inexperienced (with sqlboiler) developers that are coming from other orms/query builders

Thanks for taking your time to review it!

-b

aarondl:
I really appreciate the research you did into the other ORM-likes out there to determine the best course of action for this package.

I think that it comes down to one of:
* Throw a better error at runtime
* Use `false/0=1`

We could actually do a check to see if the where clause is present but empty, and refuse to do the query silently in some fashion too. But it's a bit more magic and it's just for this one case so it doesn't feel right.

Before we merge one thing to do is figure out why AR does 1=0? What's the history behind that because it may be a better check.

I'm still hesitating on this because I really do not like the idea of empty queries going to the database. It feels like a fundamentally wrong thing to do. But it's a somewhat indefensible perspective because currently users who have a potential `nil` slice for `WhereIn` are subject to runtime nil errors. Perhaps in an ideal world, a linter could check to see if there's a possibility of an empty slice and then based on that give a warning to do the if. Maybe the magical solution of ignoring the query completely isn't the worst idea?

jackdoe:
When I decided to use FALSE I checked sqlite, mysql and postgres, but it turns out `FALSE` is quite new in sqlite3, so i pushed a change to use `1=0` instead of false in case someone uses sqlboiler with older version of sqlite. (added in https://www.sqlite.org/releaselog/3_23_0.html)

I agree with you that ideally query wont be sent, but I think it is also weird that query wont be executed when someone expects that query will execute (even being empty), and given the complexity needed to check for other clauses and evaluate _if_ it is actually empty query is not worth it.

aarondl:
I'm not concerned with old versions of sqlite3, there's no backwards compatibility guarantees. But MSSQL (SQLServer) needs to be checked for compatibility. I'm happy with false so long as everything supports it within a few years of recent versions which that sqlite3 one does.

I don't think anyone would expect any query to execute if it were a no-op. I'm being a bit stubborn here because this is the sort of programming I dislike in general. Where we do the wrong thing here, there, everywhere and suddenly our performance is dying from a death-by-a-thousand cuts sort of thing. It's the sort of thing that leads to bloatware and programs in the 70s doing things better than we do them today - it all gets fairly philosophical. The only reason I'm considering this change is it seems like a lesser evil when compared with "my query works fine in dev, but when I use it in prod (and my list is empty and I don't know it) I get errors coming back".

jackdoe:
I understand your point, but also consider as a developer I dont want to think about it, and the worse thing I want is runtime error that could've easily be handled from the underlying library, especially considering most of the similar libraries handle it (consider how many people coming from gorm or activerecord will expect this behavior)

queries/query_builders.go
```diff
@@ -367,60 +367,70 @@ ManualParen:
...
-			if matches == nil {
-				clause, count := convertInQuestionMarks(q.dialect.UseIndexPlaceholders, where.clause, startAt, 1, ln)
+			// WHERE IN () is invalid sql, so it is difficult to simply run code like:
+			// for _, u := range model.Users(qm.WhereIn("id IN ?",uids...)).AllP(db) {
+			//    ...
+			// }
+			// instead when we see empty IN we produce 1=0 so it can still be chained
+			// with other queries
+			if ln == 0 {
+				buf.WriteString("(1=0)")
```
aarondl: Sorry for taking so long to get to this. I've decided to merge this but this should really have been an early `break` to avoid indenting the remainder of the code. If you change that I'll merge it :)

jackdoe: @aarondl sorry for the delay, i missed the message; i pushed the early break

jackdoe:
just ping @aarondl , not sure if you saw the early break thing

thanks for taking time to look into it

-b

aarondl:
I tend to batch my time in open source. Been relatively busy these days in my spare time so I'm only doing a push on sqlboiler today since the last time we exchanged messages (26 days ago!) I realize that's not ideal but it's how I've resolved to manage my time in Open Source. Usually I try to do a day or two every 2 weeks but sometimes things slip!

Thank you for your patience and for the PR.

Issue #227, opened 2017-11-25
Question passing an Int slice into WhereIn


fi0:
how do I pass a slice into whereIn? 

example
`qm.WhereIn("id in ?", 1, 2, 3)`

these won't work
```
ids := []int{1, 2, 3}
qm.WhereIn("id in ?", ids)
```
```
ids := []int{1, 2, 3}
qm.WhereIn("id in ?", ids...)
```

aarondl:
The bottom one should be working, can you turn on debug mode and show what kind of query or error you're getting? Need more information to determine what the problem is :)

payaaam:
I have had problems with this too. The solution is to convert your slice to `[]interface{}`.

*Example*
```go
ids := []int{1, 2, 3}
convertedIDs := make([]interface{}, len(ids))
for index, num := range ids {
	convertedIDs[index] =  num
}

qm.WhereIn("id IN ?", convertedIDs...)
```

fi0:
thank you it works

aarondl:
I actually think we should try to fix this. It shouldn't really behave like this and we may be able to do something about it.

ceshihao:
This problem is from golang/go#16235
It seems to be fixed in go1.9 by supporting custom import types.

tonyhb:
@aarondl I'm pretty sure we can fix this with a quick bit of reflection.

Right now we always expect that a user splats their variadic arg.  That's the root cause of the issue. 

For argument's sake, lets allow people to call WhereIn with a slice (`qm.WhereIn("id IN ?", ids)` where ids is a slice).

Then, we inspect our variadic argument to see if `len` is 1.  If so, we _may_ have a slice passed.  We can then fix WhereIn as follows:

```
func WhereIn(clause string, args ...interface{}) QueryMod {
	if len(args) == 1 && reflect.TypeOf(args[0]).Kind() == reflect.Slice {
		val := reflect.ValueOf(args[0])
		slice := make([]interface{}, val.Len())
		for i := 0; i < val.Len(); i++ {
			slice[i] = val.Index(i)
		}
		return WhereIn(clause, slice...)
	}

	fmt.Printf("continue: %vv", args)
}
```

We use reflect to see if the only item in `arg` is a slice, and if so we handle the conversion of slice to interfaces for people automatically.  Type safe, no impact, better UX.

I'll make a PR now

tonyhb:
Here's a quick demo: https://play.golang.org/p/0rrgz5cwFtj

[...]

PR #673, opened 2020-02-21
Add ability to use WhereIn with any slice type

tonyhb:
Fixes https://github.com/volatiletech/sqlboiler/issues/227 by using reflect to automatically convert any concrete slice type to a slice of interfaces where passed.

See https://github.com/volatiletech/sqlboiler/issues/227#issuecomment-589784513 for more info.

In summary, this PR allows us to:

```go
ids := []int{1, 2, 3, 4}
qm := WhereIn("id in ?", ids)
```

Verses before when we needed to:

```go
ids := []int{1, 2, 3, 4}

slice := make([]interface{}, len(ids))
for index, num := range ids {
        slice[index] = num
}

qm := WhereIn("id in ?", slice...)
```

### DIFF
--- a/queries/qm/query_mods.go
+++ b/queries/qm/query_mods.go
@@ -1,6 +1,7 @@
 package qm
 
 import (
+	"reflect"
 	"strings"
 
 	"github.com/volatiletech/sqlboiler/queries"
@@ -237,6 +238,18 @@ func (qm whereInQueryMod) Apply(q *queries.Query) {
 // WhereIn allows you to specify a "x IN (set)" clause for your where statement
 // Example clauses: "column in ?", "(column1,column2) in ?"
 func WhereIn(clause string, args ...interface{}) QueryMod {
+	// Allow people to use WhereIn with any type slice.  If we only have one
+	// element within `args` and it's a slice, convert the slice to slice of interfaces
+	// so we can splat it.
+	if len(args) == 1 && reflect.TypeOf(args[0]).Kind() == reflect.Slice {
+		val := reflect.ValueOf(args[0])
+		slice := make([]interface{}, val.Len())
+		for i := 0; i < val.Len(); i++ {
+			slice[i] = val.Index(i)
+		}
+		return WhereIn(clause, slice...)
+	}
+
 	return whereInQueryMod{
 		clause: clause,
 		args:   args,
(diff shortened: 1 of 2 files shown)


tonyhb:
Le derp, the test is broken and I will fix :)

tonyhb:
Note that, yes, this does now mean we have reflection as a dependency.  However, reflection is *only* used if the len of args == 1, and only then if the first element is of type slice.  Splatting your variadic parameter avoids all reflection and should not incur any speed penalties :)

aarondl:
This is an interesting change. On the one hand it solves a common problem that people have with the library (that expanding the variadic []int they have breaks because it wants []interface{} and Go is too restrictive). On the other hand it introduces a sort of weirdness around the types involved. Ordinarily if you saw a function signature like this you'd say: "Oh, I can do `1,2,3` or I can do `[]int{1,2,3}...` as an argument. But now the frictionless way is to do: `[]int{1,2,3}`.

I'm not sure what to make of it. It solves the pain, but only if you know about this little trick that sort of flies in the face of the language's normal conventions and would rely on documentation/closing issues pointing to said documentation to solve.

The other question is: How necessary is this given that this common use case is eliminated completely by `models.UserWhere.ID.IN([]int{1,2,3}...)`?

tonyhb:
That's a good question and perhaps I need to put up a PR for generic types.  By default, sqlboiler doesn't generate an `IN` helper for custom types (I use UUIDs from `github.com/google/uuid` for primary keys).

This means that I always need to convert a slice of UUIDs to the slice of interfaces in order to query via an IN right now.

tonyhb:
Side note:  I appreciate your thoughtful replies to everybody's PRs!

aarondl:
Currently in our code we also use UUIDs but we rely on postgres to do the validation largely for us. The UUIDs inside the objects themselves are strings so we're able to use the string helpers so we haven't run into this.

I'm not sure if you're comfortable with that approach because technically it could be any old string and type safety etc. However I do think that we will not go ahead with this change as it was just a bit too clever :)

I'm opened to some more interesting WhereIn generators for different types. Not sure what we could do exactly but maybe there's some possibilities there?

I'm going to close this PR, but I do appreciate the effort you put in to open it and the discussion it prompted. Thanks for that!

PR #683, opened 2020-02-25
Update query.go

ORESoftware:
for debugging, I want to print the raw sql generated by `qry := model.NewQuery(args...)`

for some reason however, it's returning an empty string :(

### DIFF
--- a/queries/query.go
+++ b/queries/query.go
@@ -115,6 +115,10 @@ func (q *Query) Exec(exec boil.Executor) (sql.Result, error) {
 	return exec.Exec(qs, args...)
 }
 
+func (q *Query) GetSQL() (string){
+	return q.rawSQL.sql;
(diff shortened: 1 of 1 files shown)


ORESoftware:
alternatively, this might be better:

```
func (q *Query) GetSQL() (string, []interface{}){
	return BuildQuery(q)
}
```

as long as it doesn't mutate q

jarleriksen:
Have you tried to turn on `boil.DebugMode = true`?

aarondl:
As mentioned by @jarleriksen (thanks again for fielding so many questions as I've been busy) `boil.DebugMode` will print out the queries that are being run and their arguments. There's also new ways recently that have made their way into sqlboiler to allow you to pass context values that enables per-operation debug logging.

Issue #510, opened 2019-04-02
Does eager loading do joins, or separate queries?


e-nikolov:
When I enable the debug logging I see a separate query being printed for each of the eagerly loaded relationships, but i was under the impression that qm.Load() would automatically do joins where possible.

aarondl:
It currently does separate queries. There's room for improvements if it were able to use joins in some way, but I think it'd require a complete rewrite of the entire eager loading system. Sorry for this! I wish it would use joins where possible too.

requaos:
This truly is unfortunate, for places where I have observed this impact
performance I use qm.SQL() to write the query myself as a workaround. I
hope this helps.

razor-1:
I was looking for this also. It's really too bad, because everything works great with `Load`. But issuing separate queries for every related table is a nonstarter from a performance perspective, at least for us.  I tried playing around with InnerJoin and building up the query that way, but then the relationship doesn't seem to get set up correctly so that it can be easily accessed via `.R`.

aarondl:
@razor-1 Relationships are never set by InnerJoin as it's not meant to be relationship aware. That's simply there for you to be able to pull in extra data in the case you need to do where statements on it or perhaps do custom binding to get specific columns back.

This issue is somewhat unfortunate though.

tonyhb:
Could we potentially start a backlog of issues for refactors like this using labels?  I'm wondering if that may inspire people to help out.

I do agree that questions should be closed out!

PR #684, opened 2020-03-03
Eager Loading via JOIN

razor-1:
Opening this draft PR to have a discussion on whether this kind of approach is acceptable to solve for #510. The basic idea is to add some additional generated types for each table, and then leverage them to create the necessary queries to do an eager loading using JOIN.

This isn't going to solve every use case - it only goes one level deep, for example, and doesn't recursively join children that themselves could benefit from being joined.  But it will be better than not having it, I think.

Tests are needed of course, and probably some better comments, but before going further down this road I wanted to get some feedback on the overall approach.


aarondl:
@razor-1 I truly appreciate the attempt. But I think that if we're to replace the existing code we want to ensure that it's going to do the job properly. There are a number of pain-points that require the current system to be overhauled, namely:

* Joins should be used where possible
* Query mods should work in all situations

The approach of enabling joins other than `inner` by way of a fully nullable type I think could be circumvented somehow by doing two scans on null-y joins:

```
	rows, err := db.Query(`select null from users;`)
	if err != nil {
		panic(err)
	}

	for rows.Next() {
		var id string
		var idCheck interface{}
		if err := rows.Scan(&idCheck); err != nil {
			panic(err)
		}

		if idCheck == nil {
			fmt.Println("ID was nil!")
			continue
		}

		if err := rows.Scan(&id); err != nil {
			panic(err)
		}

		fmt.Println("found ID: ", id)
	}
```

If you change `null` to `id` in the code above in the query, it correctly outputs the IDs for the users in the table. So you can imagine if we have a join that we know is going to be potentially nullable we could `make([]interface{}, len(columns))` (which is roughly the same allocation as we're doing by creating a big nullable struct). And use that to check if all the columns (or maybe just important ones?) are null or not.

I haven't fully considered the pros and cons of each of these mind you. Perhaps you could share your thoughts on this feedback and we could go from there.

razor-1:
Thank you for the feedback. Let me try to find some time to experiment with this approach for the null issue, if it works it would be superior.

My intention was not to replace `Load` any time soon, if ever, but to offer a joined alternative for common use cases. I'll also look to make sure nothing in the existing `Load` flow was changed, so that this can stand on its own for now.

razor-1:
@aarondl latest commit hopefully moves in the right direction - using a preliminary scan if there's a possibility of null joins. Also moved to using inner join when it's a non-null join. Thanks.

razor-1:
I see that there is an accepted PR for outer joins now. If this is headed in the right direction at all then I could refactor to bring up to current with the head of dev and incorporate that...

templates/00_struct.go.tpl
```diff
@@ -23,6 +23,31 @@ type {{$alias.UpSingular}} struct {
 	{{end -}}
 }
 
+
+{{if .Table.FKeys -}}
```
aarondl, line 27: I'm particularly curious what this is all for. Can we not use the .R struct to hold all of this? Or is this some avoidance of an issue where you could have a row that has 3 entities in it, 1 of which is null?

templates/singleton/boil_types.go.tpl
```diff
@@ -75,8 +160,9 @@ It only titlecases the EnumValue portion if it's snake-cased.
 // Enum values for {{if $isNamed}}{{$name}}{{else}}{{$table.Name}}.{{$col.Name}}{{end}}
 const (
 	{{- range $val := $vals -}}
+		{{- $valStripped := stripWhitespace $val -}}
```
aarondl, line 163: Why's this necessary? It hasn't been in the past. Should this make it in as a separate fix?

templates/singleton/boil_table_names.go.tpl
```diff
@@ -7,3 +7,24 @@ var TableNames = struct {
 	{{titleCase $table.Name}}: "{{$table.Name}}",
 	{{end -}}
 }
+
+var TypeNameToTableName = map[string]string {
```
aarondl, line 11: Are these useful for people? Or for the new mechanism?

strmangle/strmangle.go
```diff
@@ -18,8 +18,9 @@ var (
 	smartQuoteRgx = regexp.MustCompile(`^(?i)"?[a-z_][_a-z0-9\-]*"?(\."?[_a-z][_a-z0-9]*"?)*(\.\*)?$`)
 
 	rgxEnum            = regexp.MustCompile(`^enum(\.[a-z0-9_]+)?\((,?'[^']+')+\)$`)
-	rgxEnumIsOK        = regexp.MustCompile(`^(?i)[a-z][a-z0-9_]*$`)
+	rgxEnumIsOK        = regexp.MustCompile(`^(?i)[a-z][a-z0-9_\s]*$`)
```
aarondl, line 21: This is definitely looking like a fix that has nothing to do with this PR that we should split out

queries/reflect.go
```diff
@@ -303,10 +487,66 @@ Rows:
...
+// both string slices passed to this function will contain entries like "table.column"
+func findNullTables(allColumns, nullColumns []string) []string {
+	nullTables := make([]string, 0)
+
+	allColumnsMap := tableColumnMap(allColumns)
+	nullColumnsMap := tableColumnMap(nullColumns)
+
+	for table, columns := range nullColumnsMap {
+		// if all the columns in the null map are the same as all the columns in the table, then it's a null table
+		if len(columns) == len(allColumnsMap[table]) {
```
aarondl, line 507: Might be nicer to do an early `continue` here by inverting the condition

queries/reflect.go
```diff
@@ -158,10 +174,95 @@ func (q *Query) Bind(ctx context.Context, exec boil.Executor, obj interface{}) e
 		return errors.Wrap(err, "error from rows in bind")
 	}
 
-	if len(q.load) != 0 {
+	if len(q.load) != 0 && !q.loadJoin {
 		return eagerLoad(ctx, exec, q.load, q.loadMods, obj, bkind)
 	}
 
+	if q.loadJoin {
```
aarondl, line 181: I wish this could be a bit more generic and wasn't a big special case everywhere. If we were to do this properly I don't think we'd have to make a large distinction, it would only be a difference in how we look up the data.

drivers/column.go
```diff
@@ -11,6 +11,7 @@ import (
 type Column struct {
 	Name      string `json:"name" toml:"name"`
 	Type      string `json:"type" toml:"type"`
+	NullType  string `json:"null_type" toml:"null_type"`
```
aarondl, line 14: What's the purpose of this given `Nullable` + `Type`?

razor-1:
OK. I was going for more of an iterative approach where we start with something and then add more functionality as we go. But it's your project; if you want to have it all at once that's your call. I don't think I'm in a position to do it all at once. I will try to go through and comment on the above though, and some of them you're right are fixes that should be broken out, and I'll try to do that. Thanks for looking at it, I will close it down once addressing the comments.

templates/singleton/boil_types.go.tpl
razor-1: because if the enum value has a space in it (which is allowed in mysql at least, perhaps other dbs) then the generated code is invalid.

strmangle/strmangle.go
razor-1: it is. I'll create a separate PR for the enum space fix.

templates/00_struct.go.tpl
razor-1: you're right, I think we could just add the `boil` tags to the relevant items in the .R struct and use it directly. While that would be a bit of a change to the existing struct, and I took a perhaps too-conservative approach of not wanting to change it, in hindsight it would probably be cleaner.

templates/singleton/boil_table_names.go.tpl
razor-1: I think it's generally useful, but also specifically used when building the join query mod.

queries/reflect.go
razor-1: agreed!

drivers/column.go
razor-1: no purpose - just an unintentional artifact of the previous method of using the all-null struct, and this was not cleaned up completely.

queries/reflect.go
razor-1: Yes, I understand. If we did a wholesale change to using joined eager joins everywhere, trying to do both together probably wouldn't be needed. I personally think there is some risk in trying to change it all at once, and that having both in parallel side by side may be useful - to people who depend on the current behavior, or just to bake the new approach over time - but it's all good. Thanks again.

wsantos:
@aarondl Have we made progress on this? how do you guys normally do this kind of query? Thanks in advance.

aarondl:
No progress has been made on this. If I want the performance I go to raw queries, if I don't care I use the eager load and ignore that it's being inefficient.

wsantos:
@aarondl do you mind showing a simple example ? do I need to create a custom structure, and the bind will work correctly?

wsantos:
So just to make it clear, I'd like to know how to do custom query and bind the model along with its relation, I'm not being able to make it work so far, I've found your video here (https://www.youtube.com/watch?v=M9bgMOLQLs8) but still not sure how to properly do it.

istrau2:
@aarondl This is problematic for us as well. We've just completed a v1 and in testing we see eager loads being done by multiple queries. Obviously not an issue for a `.One` but for an `.All` it can be quite expensive.

Obviously, in the long run it would be great if eager loading was done via join. In the short term, we'd like to `.Bind()` into the `.R` associations. This is to avoid racking up a bunch of tech debt by creating a bunch of entities that don't really need to be there.

Can you give any clarification regarding binding into associations proper (rather than artificial objects)? What do you do in your codebase?

Thanks

aarondl:
Hi @wsantos and @istrau2.

See the documentation for the `Bind()` method it has everything you need in it. There are also examples in the README of how to do inner join queries with Bind.

Bind itself cannot directly bind into the R structs. That's part of the eager loading stuff.

You can however do `myStruct.R = myStruct.R.NewStruct()` (if I recall correctly) to instantiate it and then bind to the fields within. It's onerous but you could make some reflect based stuff yourself in order to do that.

PR #686, opened 2020-03-04
Support outer joins

### LINKED ISSUE #153: Question about using LEFT OUTER JOIN in query building
So... this might be a fairly easy question to answer, but I couldn't find a way of making it work. 

I'm trying to do a simple query that needs a `LEFT OUTER JOIN` with another table, I'm trying to use the `models.NewQuery` method for building it and found that the only join type that's available in `qm` is `InnerJoin`, I looked around and found that the queries module has `joinKind` constants that do include other types of joins, but couldn't find how to use it with query building.

Example of what I imagine:
```
// Use query building
err := models.NewQuery(db,
  Select("pilots.id", "pilots.name", "jets.id", "jets.pilot_id", "jets.age", "jets.name", "jets.color"),
  From("pilots"),
  LeftOuterJoin("jets on jets.pilot_id = pilots.id"),
).Bind(&paj)
```

Is there any way of doing this right now, or do I have to use Raw querying?

Thanks!

tzachshabtay:
Resolves #153.

This adds support for left/right/full joins.

### DIFF
--- a/queries/_fixtures/16.sql
+++ b/queries/_fixtures/16.sql
@@ -1,1 +1,1 @@
-WITH cte_0 AS (SELECT * FROM other_t0), cte_1 AS (SELECT * FROM other_t1 WHERE thing=$1 AND stuff=$2) SELECT * FROM "t";
+SELECT "cats".* FROM "cats" LEFT JOIN dogs d on d.cat_id = cats.id;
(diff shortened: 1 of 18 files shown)


aarondl:
For those adding emojis to this issue please be mindful that this does not just magically work out of the box and it is a half solution. This just gives you a query mod that will break under normal circumstances. You have to use custom data structures to make it work (much like an inner join where you want to return data).

This LGTM, thank you for the PR @tzachshabtay!

PR #701, opened 2020-04-03
Generate types for tables only

### LINKED ISSUE #699: Until views are supported they should be automatically ignored

v3.6.1

PostgreSQL 10.4

`sqlboiler psql`

```sql
CREATE VIEW dummy_view AS
SELECT 'dummy' AS name
```

"Error: primary key missing in tables (dummy_view)" is thrown. Why are "View" objects included in the generation process if they are not supported?

chochihim:
I find that mysql and mssql already have the condition `table_type = 'BASE TABLE'`. I just add it in psql for consistent behaviour. Resolve #699


aarondl:
Merged (not yet pushed), thank you for the PR!

PR #702, opened 2020-04-03
bump  sqlBoilerVersion

Dot-H:
# What

Update `sqlBoilerVersion` in `main.go`.

# Why

This was causing `sqlboiler --version` to return `3.6.1` instead of `3.7.0`.


aarondl:
Thanks. Will fix this by creating a new release.

PR #705, opened 2020-04-03
support enums that contain whitespace

razor-1:
If the value of an enum contained whitespace, it would be omitted from the generated code. This adds support for such enums.


aarondl:
Merged (not pushed yet), thanks for the PR!

PR #709, opened 2020-04-08
Fix tests generator for PostgreSQL partitioned tables

troyanov:
This PR should fix an issue with failing tests like this:
```
devices_session_test.go:705: Unable to upsert DevicesSession: models: unable to upsert devices_session: pq: insert or update on table "devices_session_default" violates foreign key constraint "fk_device_id"
```

So here are two examples (regular table and partitioned one):

```sql
-- table without partition
CREATE TABLE devices_session (
    device_id integer NOT NULL,
    payload_size int NOT NULL,
    timestamp timestamp(3) without time zone NOT NULL,
    PRIMARY KEY(device_id, timestamp),
    CONSTRAINT fk_device_id FOREIGN KEY (device_id) REFERENCES devices(id)
);

-- partitioned table
CREATE TABLE devices_session (
    device_id integer NOT NULL,
    payload_size int NOT NULL,
    timestamp timestamp(3) without time zone NOT NULL,
    PRIMARY KEY(device_id, timestamp),
    CONSTRAINT fk_device_id FOREIGN KEY (device_id) REFERENCES devices(id)
) PARTITION BY RANGE (device_id);
CREATE TABLE devices_session_default PARTITION OF devices_session DEFAULT;
```

After creation they look a bit different in output provided by `pg_dump`.
```
-- table without partition
ALTER TABLE ONLY public.devices_session
    ADD CONSTRAINT fk_device_id FOREIGN KEY (device_id) REFERENCES public.devices(id);

-- partitioned table
ALTER TABLE public.devices_session
    ADD CONSTRAINT fk_device_id FOREIGN KEY (device_id) REFERENCES public.devices(id);
```
You might already notice that, the `ONLY` keyword is missing.

>If ONLY is specified before the table name,
only that table is altered. If ONLY is not specified,
the table and all its descendant tables (if any) are altered.
Optionally, * can be specified after the table name to explicitly
indicate that descendant tables are included.

So in order to capture all ALTER TABLE statements I've changed `rgxPGFkey` expression...


aarondl:
Ah thanks for this @troyanov. Clearly an issue that was created when we originally were staring at pg dumps output without close inspection. Merged.

PR #715, opened 2020-04-11
Add template to generate order clause

takisawa:
I have added a template that generates the code below.
```
var Book = struct {
  IDAsc string
  IDDesc string
  NameAsc string
  NameDesc string
}{
  IDAsc: "\"books\".\"id\" ASC",
  IDDesc: "\"books\".\"id\" DESC",
  NameAsc: "\"books\".\"name\" ASC",
  NameDesc: "\"books\".\"name\" DESC",
}
```
This generated code is used as follows.
```
qm.OrderBy(BookOrder.NameDesc),
qm.OrderBy(BookOrder.IDAsc),
```

Thanks.

### DIFF
--- a/templates/22_order.go.tpl
+++ b/templates/22_order.go.tpl
@@ -1,0 +1,15 @@
+{{- $alias := .Aliases.Table .Table.Name -}}
+
+var {{$alias.UpSingular}}Order = struct {
+	{{range $column := .Table.Columns -}}
+	{{- $colAlias := $alias.Column $column.Name -}}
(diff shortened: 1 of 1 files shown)


aarondl:
I think that this is a large amount of generated code for such a feature.

Could we maybe consider `qm.Asc`/`qm.Desc` that may be used like so?

```
qm.OrderBy(qm.Asc(models.BookOrderColumns.ID)),
```

takisawa:
Thank you for your reply.
I understand it.

I will close this PR.

Issue #525, opened 2019-05-09
BeforeSelectHook


KopiasCsaba:
Hi!

I think it would be great, to have a BeforeSelectHook, where we could modify the incoming query.

It would be great for things like enforce deleted_at is null.

What do you think?

Thanks!

aarondl:
We haven't done this in the past because there was previously no way to disable hook processing for a given query. Now there is so we could consider something like this but I still prefer writing a small helper function to apply the necessary filters despite being easy to miss.

frederikhors:
@aarondl any news on this?

aarondl:
@frederikhors Nope. The label "help wanted" implies I'm not going to be taking care of this, and no one else seems to be interested in doing it either! :)

namco1992:
Hi @aarondl, I created [this issue](https://github.com/volatiletech/sqlboiler/issues/262) long time back and we end up maintaining our own templates. It seems @frederikhors was facing the same problem here. Would you like to shed light on how would you like to resolve this?

I can think of two solutions:
1. Add support for `BeforeSelectHook` as this issue mentioned. However, I'm not sure how we can mutate the query in the hooks since the hook func signature is `func(context.Context, boil.ContextExecutor, *Model) error`.
2. Add a flag to enable the soft delete, it includes adding soft delete func and update selects to exclude soft-deleted records, which is what we are doing in our templates.

I would like to work on this if you would like to share some thoughts. Sqlboiler is widely used in our projects since I first made this decision, and I feel it's not ideal to maintain our own templates as it will get out of sync and even broken in the future. Therefore I propose to help and make it compatible with our needs and probably other people's needs too. Thanks.

aarondl:
@namco1992 I think that without proper template support you could not do any kind of soft deletion. `1` is definitely out of the picture due to this (given the signature). The queries are also sort of write-once in a way.

I think the only way to accomplish this is exactly what you've mentioned, a soft delete "magic field" much like we have `created_at timestamp` and `updated_at timestamp`, you would need `deleted bool`.

There are quite a few places where selects happen however, so we would need to be cognizant of all of them:
* Regular selects (`models.Users()`)
* Relationship helpers (`models.User.Videos()`)
* Eager loading (`models.Users(qm.Load("Videos"))`)

It's also not 100% clear what we're supposed to do for APIs such as `DeleteAll()`. Assuming that in a soft-delete world we would want this to soft-delete and not ACTUALLY delete, how do we then delete for real?

[...]

PR #716, opened 2020-04-12
Add soft delete

### LINKED ISSUE #262: ssl-mode value is not compatible with mysql 5.6 and how should I implement the soft delete
Hi,
Thank you for sharing this amazing tool! We are trying to use sqlboiler in our projects and it works great, but I still have some minor issues which I hope you can help with.

1. In the generated code `main_test.go`, the `makeOptionFile` method provides the credentials and params for the db connection, `fmt.Fprintf(tmp, "ssl-mode=%s\n", m.sslMode(m.sslmode))` will try to add the `ssl-mode` based on the settings. However, the `DISABLED` and `PREFERRED` are only supported in mysql 5.7 and they will be rejected in mysql 5.6. You can refer to the doc [here](https://dev.mysql.com/doc/refman/5.6/en/encrypted-connection-options.html#option_general_ssl-mode).

It's not a big deal but I need to manually change the file every time I re-generate the models. According to the doc, "No explicit --ssl-mode option is equivalent to no explicit --ssl option." is applicable to both mysql 5.6 and 5.7. **So I think the better way is to omit the `ssl-mode` when it's disabled and be compatible with both versions.**

2. Shall we have a `--soft-delete` arg when we generate the code? It's common that we update the `deleted_at` in the record instead of hard delete the record. I think it's doable to utilize the `BeforeDeleteHooks` to update the record and return an error to stop the real delete, but I don't think it's a good practice. What do you think? How should I implement the soft delete besides playing with hooks? Thanks.

namco1992:
This PR is for review only.

It tries to resolve #262 and the detailed discussion happens in #525.

Based on the discussion I had with @aarondl:

1. Add a new flag `--add-soft-deletes` to support this functionality. Perhaps in v4 we could flip it to `--no-soft-deletes`;
2. Only the tables have `deleted_at` column and the go type `null.Time` will be considered supporting soft deletes;
3. `Delete` and `DeleteAll` methods accept `hardDelete bool` to indicate if it's a hard deletion;
4. All the other retrievals will have the statement `deleted_at is null` automatically in order to omit the soft deleted records;
5. Test suites for soft deletes will be generated if soft deletion is enabled.

However, I encountered an issue with the generated tests. In the tests all the non-default columns are randomised generated, including the `deleted_at` field. Then it leads to the failures of the assertions since the `deleted_at` is not null. Maybe we can make an exception for this but I'm not sure if it's the ideal way. @aarondl would you like to share some thoughts?


frederikhors:
Only a clarification: this does not resolve #525.

namco1992:
@frederikhors sorry for hijacking the original issue, updated the PR, thanks.

boilingcore/config.go
```diff
@@ -22,6 +22,7 @@ type Config struct {
 	Debug             bool     `toml:"debug,omitempty" json:"debug,omitempty"`
 	AddGlobal         bool     `toml:"add_global,omitempty" json:"add_global,omitempty"`
 	AddPanic          bool     `toml:"add_panic,omitempty" json:"add_panic,omitempty"`
+	SoftDeletes       bool     `toml:"soft_deletes,omitempty" json:"soft_deletes,omitempty"`
```
aarondl, line 25: Pedantry: Can we name this `AddSoftDelete` to follow the sort of boolean "Add" or "No" thing we have going on here?

templates/18_delete.go.tpl
```diff
@@ -1,10 +1,12 @@
 {{- $alias := .Aliases.Table .Table.Name -}}
-{{- $schemaTable := .Table.Name | .SchemaTable}}
+{{- $schemaTable := .Table.Name | .SchemaTable -}}
+{{- $canSoftDelete := .Table.CanSoftDelete -}}
+{{- $soft := and .SoftDeletes $canSoftDelete }}
```
aarondl, line 4: This variable has been created but further down the same `and` is happening.

templates/18_delete.go.tpl
```diff
@@ -122,12 +150,20 @@ func (q {{$alias.DownSingular}}Query) DeleteAllP({{if .NoContext}}exec boil.Exec
 {{end -}}
 
 // DeleteAll deletes all matching rows.
-func (q {{$alias.DownSingular}}Query) DeleteAll({{if .NoContext}}exec boil.Executor{{else}}ctx context.Context, exec boil.ContextExecutor{{end}}) {{if .NoRowsAffected}}error{{else}}(int64, error){{end -}} {
+func (q {{$alias.DownSingular}}Query) DeleteAll({{if .NoContext}}exec boil.Executor{{else}}ctx context.Context, exec boil.ContextExecutor{{end}}{{if $soft}}, soft bool{{end}}) {{if .NoRowsAffected}}error{{else}}(int64, error){{end -}} {
```
aarondl, line 153: I think that maybe the false value here should be for soft deleting (false is the default)? So maybe we turn this into the `hardDelete` or `realDelete`?

aarondl:
If you can fix this PR up soon, I'd like to put it in the next release (which I also want to do in the next few days). Do you think you can find the time to complete this? If not that's okay, let me know either way.

namco1992:
I have some work stuff to rush out this week, I can finish the fix by the end of this week, is it ok for you?

namco1992:
Hi @aarondl just a reminder that I've done the changes as you requested, thanks!

templates/18_delete.go.tpl
```diff
@@ -211,6 +245,33 @@ func (o {{$alias.UpSingular}}Slice) DeleteAll({{if .NoContext}}exec boil.Executo
 	}
 	{{- end}}
 
+	{{if $soft -}}
```
namco1992, line 248: I missed the `slice.DeleteAll` in the previous PR and added it in, take note and review again pls, thanks!

randomize/randomize.go
```diff
@@ -60,6 +60,7 @@ func Struct(s *Seed, str interface{}, colTypes map[string]string, canBeNull bool
 	copyBlacklist := make([]string, len(blacklist))
 	copy(copyBlacklist, blacklist)
 	blacklist = copyBlacklist
+	blacklist = append(blacklist, "deleted_at")
```
aarondl, line 63: This should go in the calls to randomize, not in the randomize function itself I think.

namco1992: I understand that adding in the calls to randomize is neater. However, I found that there are plenty of places (~50) that need to be modified and the logic needs to be duplicated everywhere. Therefore I ended up changing the `randomize.Struct` func. The only potential issue I can think of is that the `deleted_at` column serves some other purposes and can't be null, but I'm not sure if it's a valid concern and I decided to go to this conventional approach.

aarondl: Okay. I'll probably merge as-is and then attempt to do some fancy regex replaces to put this in the correct spot.

nii236:
From the OP:

> All the other retrievals will have the statement deleted_at is null automatically in order to omit the soft deleted records

This is not suitable for our purposes, which is the ability to run SQLBoiler retrievals and show the soft-deleted records for undeletion.

Is there a way we can have soft-deletions, but also optionally retrieve records with and without soft-deleted records?

PR #727, opened 2020-04-27
Bind: add support for Structs with pointers

gmazzamuto:
This makes it possible, for example, to bind to Structs generated by proto3 which contain pointers to Structs.

Here is an example:
```protobuf
// proto3
message Book {
  int64 id = 1;
  string author = 2;
  string title = 3;
  NullableString subtitle = 4;
}

message NullableString {
  string string = 1;
  bool valid = 2;
}
```

Which results in the following Go structs:
```go
type Book struct {
    ID       int64
    Author   string
    Title    string
    Subtitle *NullableString
}

type NullableString struct {
    String_  string
    Valid    bool
    ...
}
```
To make `Bind()` work with this example, one has to implement `Scan()` for `NullableString`:
```go
func (s *NullableString) Scan(value interface{}) error {
	if value == nil {
		s.String_, s.Valid = "", false
		return nil
	}
	s.Valid = true
	err := convert.ConvertAssign(&s.String_, value)
	return err
}
```
```

### DIFF
--- a/queries/reflect.go
+++ b/queries/reflect.go
@@ -71,11 +71,11 @@ func (q *Query) BindG(ctx context.Context, obj interface{}) error {
 // If you neglect checking the rows.Err silent failures may occur in your
 // application.
 //
-// Valid types to bind to are: *Struct, []*Struct, and []Struct. Keep in mind
-// if you use []Struct that Bind will be doing copy-by-value as a method
-// of keeping heap memory usage low which means if your Struct contains
-// reference types/pointers you will see incorrect results, do not use
-// []Struct with a Struct with reference types.
+// Valid types to bind to are: *Struct, []*Struct, and []Struct. If your Struct
+// contains reference types / pointers, these must implement the Scan interface.
+// See the StructField struct in queries/reflect_test.go for an example. If any
+// of the pointers in the struct are uninitialized, they will be initialized
+// automatically.
 //
 // Bind rules:
 //   - Struct tags control bind, in the form of: `boil:"name,bind"`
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: queries/reflect_test.go [hunk 1]
```diff
@@ -12,6 +12,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/volatiletech/null/convert"
```
aarondl, line 15: This should be pointing to null/v8

queries/reflect.go
```diff
@@ -71,11 +71,11 @@ func (q *Query) BindG(ctx context.Context, obj interface{}) error {
...
-// Valid types to bind to are: *Struct, []*Struct, and []Struct. Keep in mind
-// if you use []Struct that Bind will be doing copy-by-value as a method
-// of keeping heap memory usage low which means if your Struct contains
-// reference types/pointers you will see incorrect results, do not use
-// []Struct with a Struct with reference types.
+// Valid types to bind to are: *Struct, []*Struct, and []Struct. If your Struct
+// contains reference types / pointers, these must implement the Scan interface.
+// See the StructField struct in queries/reflect_test.go for an example. If any
+// of the pointers in the struct are uninitialized, they will be initialized
+// automatically.
```
aarondl, line 78: This has a strange repercussion for the decimal type that sqlboiler supports. Currently the struct looks like this: https://github.com/volatiletech/sqlboiler/blob/master/types/decimal.go#L27-L29

The `NullDecimal` looks the same but behaves differently. If the pointer comes back from the database as `nil` inside the struct it must remain nil and should not be initialized to anything because then you've lost your nullability.

I did mention that this would be okay so long as it did not break the decimal type. I hope there's no actual problems here.

gmazzamuto:
Thanks for your feedback!

I fixed the import. (just so that you know, I noticed that `README.md` probably needs that kind of fix in many places, after you switched to modules)

I understand your concerns about `NullDecimal`. Actually I verified that all of your tests passed, including [this one](https://github.com/volatiletech/sqlboiler/blob/ece78e1128702fd8ab6aa6b8658240c5b1cbfa62/types/decimal_test.go#L139). If you can provide a snippet of code that fails, I'll see if I can make work :)

Thanks for your review.

aarondl:
Hi @gmazzamuto. Are you still interested in getting this change in? I realize it's been left for a very long time. My apologies. I haven't tested the NullDecimal yet.

gmazzamuto:
Hi Aaron, thanks for your message. Actually at the moment I'm not using my forked repo, that is to say my code does not rely on this change. I still think it might be of some interest in some cases, i.e. when using protobuf, but you can probably leave it aside for the moment.

vhakulinen:
What work is there left to do to get this merged?

aarondl:
@vhakulinen just have to test the NullDecimal stuff. Just need someone who cares enough about the change to do the tests and see it through :)

stephenafamo:

PR #728, opened 2020-04-28
Fix soft delete issues

psucodervn:
Fix some issues relate to soft delete.


aarondl:
Thanks very much for this PR!

Issue #329, opened 2018-07-10
Mysql model generation takes very long and exit with error


nqsang90:
If you're having a generation problem please answer these questions before submitting your issue. Thanks!

### What version of SQLBoiler are you using (`sqlboiler --version`)?
SQLBoiler v2.7.3

### If this happened at generation time what was the full SQLBoiler command you used to generate your models? (if not applicable leave blank)
`sqlboiler mysql`

### What is the output of the command above with the `-d` flag added to it? (Provided you are comfortable sharing this, it contains a blueprint of your schema)
Part of the exception is. It seems the connection takes so long and closes unexpectedly.
`packets.go:36: unexpected EOF (Invalid Connection)`
Error comes from here:
`github.com/volatiletech/sqlboiler/bdb/interface.go:62`

### Please provide a relevant database schema so we can replicate your issue (Provided you are comfortable sharing this)
Problem is this query takes very long to run
```
SELECT
	c.column_name,
	c.column_type,
	IF(c.data_type = 'enum', c.column_type, c.data_type),
	IF(extra = 'auto_increment','auto_increment', c.column_default),
	c.is_nullable = 'YES',
		EXISTS (
			SELECT c.column_name
			FROM information_schema.table_constraints tc
			INNER JOIN information_schema.key_column_usage kcu
				ON tc.constraint_name = kcu.constraint_name AND tc.table_name = kcu.table_name AND tc.table_schema = kcu.table_schema
			WHERE c.column_name = kcu.column_name AND tc.table_name = c.table_name AND
				(tc.constraint_type = 'PRIMARY KEY' OR tc.constraint_type = 'UNIQUE') AND
				(SELECT count(*) FROM information_schema.key_column_usage WHERE table_schema = kcu.table_schema AND TABLE_NAME = tc.table_name AND CONSTRAINT_NAME = tc.constraint_name) = 1
		) AS is_unique
	FROM information_schema.columns AS c
	WHERE TABLE_NAME = 'table_name_here' AND table_schema = 'schema_name_here' AND c.extra NOT LIKE '%VIRTUAL%';
```

Explain the query:
```
id	select_type	table	type	possible_keys	key	key_len	ref	rows	Extra
1	PRIMARY	c	ALL	NULL	TABLE_SCHEMA,TABLE_NAME	NULL	NULL	NULL	Using where; Open_frm_only; Scanned 0 databases
[...]

PR #732, opened 2020-05-02
Fix for the extremely slow model generation with some MySQL databases.

oderwat:
If one tries to generate the models on a MySQL server with a lot of databases and tables the generation is getting very slow (#329).

This seems to result from MySQL not optimizing queries on views very well.

The solution I have chosen and validated is to replace the table and schema in the subquery with the actual names instead of references to the parent queries.

The result for my system is insane and generation went down from over 9 hours to 4 seconds.

This fix is kinda ugly but straight forward and I think the speed advantage is worth it. Especially as the code is never being seen by anyone.


aarondl:
This fix is indeed ugly but the result is incredible. I would have never thought to try something like this. Job well done! Thank you very much for the PR. Though it was made against master not dev as is outlined in CONTRIBUTING.md so I will close the PR but it is merged into dev here: 7241ce1

oderwat:
I am sorry! Next time PRs will be against dev!

PR #733, opened 2020-05-03
Mysql MariaDB (fix 1)

oderwat:
First: Sorry but this is an incremental PR to be added after the slow MySQL fix PR. I need need to work with the code and creating a PR on master for this makes no sense for me as MySQL is unusable on my system without the fix for the generation speed.

It addresses the problem that MariaDB stores default values in a different way in the information schema. The MySQL driver test fails for MariaDB and this is part of the reason why this happens.

There are other reasons too but the different storage of the default values seems to be a real problem with sqlboiler as it would use the wrong values (`'abc'` instead of `abc`).

I actually wonder that MySQL seems to use an empty string for the NULL value. But I added that too.

There is also `CURRENT_TIMESTAMP` vs `current_timestamp()` and a different encoding for 0 bytes (`\u000` vs `\\0`). But those are just making the test fail because of the tester expecting the fixed format MySQL is using. I think it does not really matter for sqlboiler functionality.

### DIFF
--- a/drivers/sqlboiler-mysql/driver/mysql.go
+++ b/drivers/sqlboiler-mysql/driver/mysql.go
@@ -185,7 +185,7 @@ func (m *MySQLDriver) TableNames(schema string, whitelist, blacklist []string) (
 // converts the SQL types to Go types, for example: "varchar" to "string"
 func (m *MySQLDriver) Columns(schema, tableName string, whitelist, blacklist []string) ([]drivers.Column, error) {
 	var columns []drivers.Column
-	args := []interface{}{tableName, schema}
+	args := []interface{}{tableName, tableName, schema, schema, schema, schema, tableName, tableName, schema}
 
 	query := `
 	select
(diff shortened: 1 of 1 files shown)


aarondl:
It has never been my intention to directly support MariaDB so this PR is a bit awkward in this regard. Though because it should have no downsides for any other database driver I think it's okay if we allow this in. The differences are beginning to stack up though, and I wonder when we're going to need a completely new driver for mariadb. They simply aren't as compatible as people think they are.

Thanks for the PR! Merged here: 956d421

oderwat:
Yes, I was wondering the same but then it is just cloning most of the code it seems.  Thank you for accepting that PR. Not sure about having a driver for MariaDB. It would clone huge amounts of code and had to be maintained for every new feature.

I personally think it should not be split. Figuring out the server type and adjust some parts in the generation is pretty trivial. The way the tests work would need to be modified a bit, as it currently fails for a MariaDB server. I will continue the evaluation and maybe find an elegant solution for the problem.

aarondl:
Yeah. We'll have to continue to evaluate.

The tests in my opinion are sort of a mistake and I don't think it's worth it for anyone to spend much time on them. For a very basic database they work fine, but for many things they fail and it would be very difficult to make them work for everything. They don't really serve a purpose.

oderwat:
I am not sure if it was clear but here I meant the driver testing itself. I think it is very helpful to have it. At least it provides a way to validate that some "small change" does not mess up everything.

aarondl:
My mistake. I was thinking about the generated tests :p

Issue #222, opened 2017-11-14
json tag for relations


polderudo:
Is there any way to generate the Relations (R) with a json tag set? Right now they are generated like `R *test2R `boil:"-" json:"-" toml:"-" yaml:"-"``
Would be nice if building an API and returning the data 1:1 from the database.

aarondl:
The way the decision was made is that you should never have an API that returns something called `R` with a bunch of sub-relationships on it. The intention is to force you to create some sort of a useful return struct for your data going to your API. So while it might be convenient for the programmer, it's actually a bad thing for the consumers of the API in my opinion. So I doubt I'd accept this since I think it helps permeate bad APIs. I'm never completely closed on any issue, feel free to debate but I will close for now.

polderudo:
I think it depends, on what you try to achive. It doesn't make sense to publish the fields of a table at all, if you think like this. The user should have to create his own structs even for the tabledata. 
I'd like to have the relations, because i need to show related data on the page. Some 1-1 (only lookups), some 1-n (master details). Load() comes in very handy for this (by the way the main reason why i'd like to switch to SQLBoiler from sqlx).
Right now i have to create structs for nothing and add them to some intermediate lists, just to be able to send them to the client. This means i also have to create logic to get the data together.
I have about 35 tables that need to be displayed this way, which means 35 intermediate structs.
Right now the db defines the logic and thus includes all relations. The server is more like a dumb thing which just gathers/recives the data and sends it to the client/db.

Why not add a option like --include-relations?
One can then decide on his own if relations should be included or not. 
Or even better add some cunfiguration per table->relation

speatzle:
We are encountering the same problem with sqlboiler as @polderudo did a few years back.

We now have to make intermediary tables for the entire database, which is what we wanted to avoid by using sqlboiler in the first place. And all these structs that we have to make look exactly like the ones sqlboiler generates with the exception of the relationships because our clients rely on the the names and structure being the same as the db to Generate forms and replies. Maintaining all these structs is a huge nightmare that we don't want to life through.

I hope that you can consider giving the green light for a PR that adds a --include-relations flag like @polderudo suggested. Please let us know what you decide.

aarondl:
@speatzle I guess my lack of urgency or awareness of this issue comes from the fact that I have generated structs on both sides. I use OpenAPI/gRPC in any API that I'm creating and so I don't have the issue that you two are pointing out (my issue turns into one of converting and mapping between two structs of different shape).

[...]

PR #734, opened 2020-05-04
Added relation-tag flag & config option

speatzle:
This implements a flag & config option to allow you to set the json,yaml,toml,... Tags for the Relationships. Discussed in #222 .


aarondl:
I'm not sure why but this PR came with a lot of the bindata regenerated with what looks like an old or different version so it changed a lot of files it should not have. I will regenerate it properly in the merge commit. The version is outlined here: [CONTRIBUTING.md](https://github.com/volatiletech/sqlboiler/blob/master/CONTRIBUTING.md#developer-getting-started)

Thank you for this, I hope it solves some problems for people :)

PR #745, opened 2020-05-15
Add a way to reset queries so they can be reused

### LINKED ISSUE #647: Can't reuse queries.

SQLBoiler v3.6.1

Postgresql 11.5

```go
myQuery := models.Things(qm.Where("name=?", someName))

// Use myQuery once, it works.
exists, err := myQuery.Exists(
    context.Background(), db)
    if err != nil {
        return fmt.Errorf("error looking for thing: %w", err)
    }

// Use myQuery again, doesn't work.
thing, err := myQuery.One(context.Background(), db)
if err != nil {
    return fmt.Errorf("error getting existing Thing: %w", err)
}
```

I wanted to keep the code DRY by having a `myQuery` variable that I could reuse. But apparently that doesn't work. The code runs without a problem, but `thing` turns out to be the **zero-valued** `models.Thing`.

Is it an implementation idiosyncrasy of `database/sql` or of `sqlboiler`?

I know that **not reusing** `myQuery` fixes the problem, so it might not be a problem at all. But this was a little tricky to figure out at first. Seems like `myQuery` is consumed when it is run by the first _Finisher_, in this case, `Exists()`.

stephenafamo:
Also add a flag that generates models that resets the query after each finisher.

Fixes #647 

When the models are generated with the flag `--add-reset`, Queries started with `models.Pilot()` can be reused. For example:

```go
query := models.Users()

users, err := query.Count(req.Context(), s.DB)
fmt.Println(users, err)

count, err := query.Count(req.Context(), s.DB)
fmt.Println(count, err)
```

Previously that will throw an error.

It also adds the type `QueryReset` which is the primary way or resetting queries. For example:

```go
query := models.NewQuery(
    qm.From("users"),
)
resetter := q.AddReset()

resetter.Save() // saves what the query was at this point
_ = q.Bind(ctx, db, &obj) // selects all
resetter.Reset() // resets the query to the saved point

qm.Apply(query, qmhelper.Where("active", qmhelper.EQ, true)
resetter.Save() // New save point

_ = q.Bind(ctx, db, &obj) // selects all active users
resetter.Reset() // resets the query to the saved point

queries.SetCount(query)
 = q.Bind(ctx, db, &obj) // counts
```

This method is also completely backward compatible.


stephenafamo:
Can I get some feedback on this PR?

queries/query_reset.go
```diff
@@ -0,0 +1,25 @@
...
+	q     *Query
+	undo  []undoFunc
+	saved Query
+}
+
+type undoFunc = func(*QueryReset)
+
+// Save removes the effect of a finisher
+func (q *QueryReset) Save() {
+	q.saved = *q.q
```
aarondl, line 14: This is insufficient because it does not do a deep copy. If slices/maps are modified in the query it could affect a saved query.

stephenafamo:
Thanks for the feedback. I could make some changes to the PR, but I think it's better if I get a better sense of how you'd prefer the reset to work.

I think this feature is useful, but I'd like to understand what you have in mind so I see if it is something I can work on.

aarondl:
My issue with the approach lies in the fact that users will not try that first. They'll first try what andradei tried:

```
myQuery := models.Things(qm.Where("name=?", someName))

// Use myQuery once, it works.
exists, err := myQuery.Exists(
    context.Background(), db)
    if err != nil {
        return fmt.Errorf("error looking for thing: %w", err)
    }

// Use myQuery again, doesn't work.
thing, err := myQuery.One(context.Background(), db)
if err != nil {
    return fmt.Errorf("error getting existing Thing: %w", err)
}```

And only after they encounter errors/problems will they seek a resolution and finally find the resetter and learn how that works and use it.

Perhaps that's better than "they encounter errors/problems and seek a solution only to find one doesn't exist" but does there exist a design where we could simply have the obvious thing work without any additional syntax?

Basically what I'd *like* to happen (whether or not it's possible is another question) is for andadei's example to function as he expected it would.

I haven't been in the code in-depth for a very long time so I actually do not know if that's possible or what we would have to give up in order to have it (for example query caching inside the query object maybe) so that investigation would have to be done before we could make a decision.

stephenafamo:
Okay. I'll do some research on this later, and see if I can think of a clean way to implement this.

PR #750, opened 2020-05-19
Add "alias" struct tag casing

Darkclainer:
We have number of options for struct tag casing, but we have option that
respect our choice in alias section of configurations.


Darkclainer:
May be more proper way will be set this implementation for "title" casing.

aarondl:
Thanks for the PR! Be more careful with branches next time. This brought a commit into dev from master that was technically incorrect. :)

PR #751, opened 2020-05-19
Fixate external tools version

Darkclainer:
Root file tools.go should import external tools that are used in project.
If tool is runned through `go run` then its version specified in go.mod
will be used.
File tools.go have special build tag because it's not used in build only
as hint to go mod system.


aarondl:
This means that everyone running in prod will have to download/investigate the go.mod's of all these extra dependencies so we can have version pinning, yes?

Darkclainer:
I didn't quite get what do you mean so I will try to elaborate about impact of this PR.

The developing of sqlboiler mean using some tools written in Go. Everyone who makes a PR can have different version of mentioned tools. In theory this can be troublesome because different versions can have minor distinctions or there can be a case when a author of a tool messes up with semver and breaks backward compatibility. Moreover developing became harder: you need to manually install tools (and actually you need to guess correct version!). 

This PR fixes versions of tools used by sqlboiler in go generate stage. Yes, versions will be managed my go.mod like any other dependency. In fact in the file `tools.go` we say that some tools is dependency for sqlboiler. 
The development of sqlboiler will became easier for newcomers, they will be able to regereate code simply by running:

```sh
go generate ./...
```

Go get and go buid will manage tools for them. Another good point is that the code changes in generated files will become more determenistic.

This PR introduce minor penalty in execution of `go generate`, but thanks to go cache system it's not very big.

aarondl:
Yes, but look at the size of the go.mod. We've introduced a ton of extra dependencies.

Currently the way it works is it's more painful for developers because we don't have a solid manifest of dev tools, but using this approach penalizes all users of sqlboiler forever. Every time the `go` tool interacts with a codebase that includes sqlboiler, it will now have to evaluate all these other dependencies in whatever way - only to inevitably throw them in the garbage at the `build` stage. It makes for longer download times from a clean cache, and it makes for many more open source licenses to shift through to see if you're complying with said licenses etc.

Rust/NPM etc. are already in dependency hell and Go is already starting to get quite bad because people are too nonchalant with their dependencies, I do not think we should be part of that problem and should instead seek to minimize dependencies.

Until such time as there exists an NPM `dev-dependencies` equivalent for Go modules I think we should find another way of managing development dependencies and their versions. I'm opened to discussing what that system should be, but I think it's fair to say it should not impact the regular users of sqlboiler in any way.

PR #757, opened 2020-05-25
Add support for custom build tags

chrisseto:
If a user desires for unit tests to run without accessing external
  resources (a database) but still wishes to make use of the
  autogenerated tests, they may specify test-build-tags to add go build
  tags, https://golang.org/pkg/go/build/#hdr-Build_Constraints, to their
  test files. For example a user may set test-build-tags to
  ["integration"] and continue to run unit tests without a database.
  When a database is available, the test binary may be built with `-tags
  integration` to run the autogenerated sqlboiler tests.

### DIFF
--- a/boilingcore/config.go
+++ b/boilingcore/config.go
@@ -34,6 +34,8 @@ type Config struct {
 	StructTagCasing   string   `toml:"struct_tag_casing,omitempty" json:"struct_tag_casing,omitempty"`
 	RelationTag       string   `toml:"relation_tag,omitempty" json:"relation_tag,omitempty"`
 	TagIgnore         []string `toml:"tag_ignore,omitempty" json:"tag_ignore,omitempty"`
+	BuildTags         []string `toml:"build_tags,omitempty" json:"build_tags,omitempty"`
+	TestBuildTags     []string `toml:"test_build_tags,omitempty" json:"test_build_tags,omitempty"`
 
 	Imports importers.Collection `toml:"imports,omitempty" json:"imports,omitempty"`
 
--- a/boilingcore/output.go
+++ b/boilingcore/output.go
@@ -116,6 +116,13 @@ func executeTemplates(e executeTemplateData) error {
 		imps = importers.AddTypeImports(imps, e.state.Config.Imports.BasedOnType, colTypes)
 	}
 
+	var buildTags []string
+	if e.isTest {
+		buildTags = e.state.Config.TestBuildTags
+	} else {
+		buildTags = e.state.Config.BuildTags
+	}
+
 	for dir, dirExts := range e.dirExtensions {
 		for ext, tplNames := range dirExts {
 			out := templateByteBuffer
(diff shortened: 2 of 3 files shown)


Darkclainer:
I am not a maintainer, just my humble opinion.

We can add in all test function check for `-short` flag that testing package offer.  I like this solution more, because it helps to manage testing dependencies without managing additional tags. 

https://golang.org/pkg/testing/#Short

aarondl:
I tend to agree with @Darkclainer. There are other ways we can do this. The autogenerated test's bring very very little value to a solution anyway and we shouldn't be running around trying to keep them running because they create false negatives (errors that appear while running the tests will not be present while running actual sqlboiler application code).

The addition of build tags for the sqlboiler generated code itself shouldn't be useful because there doesn't seem to ever be use case to build an application's code without the sqlboiler generated code to support it.

I will close this PR as I don't have good reason to accept it unless it can be debated further. Please let me know if there are use cases I'm not thinking of where this would be useful.

Thanks for the PR regardless!

chrisseto:
I may have explained my use case poorly or may be missing another obvious solution. Apologies if either of those are they case.

I honestly see huge value from having the autogenerated tests, they've caught a few would be bugs for me (schema weirdness on my application's behalf).

For the project I'm using sqlboiler in, I'm running all of my unit tests as part of my docker build process via `go test ./...`. The difficulty is that a database won't be available in that environment. Doing so has helped force a TDD pattern and ensure that unit tests stay fast and focused (I don't mean to preach just want to motivate this a touch more).

Integration tests are run as a later portion of the build process and are gated via `// +build integration`.

This PR allows sqlboiler to slot quite nicely into that build process. I'm happy to maintain continue using my own fork if you disagree. I assumed that others may encounter the same issues as myself 🙂 

> The addition of build tags for the sqlboiler generated code itself shouldn't be useful because there doesn't seem to ever be use case to build an application's code without the sqlboiler generated code to support it.

Agreed, I just needed/wanted tags for tests and figured I add them everywhere while I was poking around 🤷. I'm happy to remove them if you would consider that to be a blocker.

> We can add in all test function check for -short flag that testing package offer. I like this solution more, because it helps to manage testing dependencies without managing additional tags.

I disagree with the use of `-short` as that flag should be concerned with runtime rather than test dependancies. Tags, on the other hand, are explicitly for build constraints and dependancies.

aarondl:
@chrisseto So the main use case here is that you'd like to always have the autogenerated tests in your code, but not run them on `go test ./...`, you only want to run them conditionally if you enable them via the tags.

It's kind of a hard sell to add a feature for this because there's already a way to do this without build tags, which is to explicitly list the packages you'd like to test. This gives you the exact same functionality without additional magic. The main difference is opt-out vs opt-in. Your use case maybe makes opt-out onerous though?

```
go test $(go list ./... | grep -v 'yourmodule.com/somepaths/v3/models')
```

PR #762, opened 2020-05-29
Ignore generated columns for psql

chochihim:
psql's [generated columns](https://www.postgresql.org/docs/12/ddl-generated-columns.html) cannot be set. So this PR add condition to filter those columns.
mysql driver already ignores virtual columns (which is generated column in psql) in #266


aarondl:
Thanks for this! Apologies in accepting it so late.

Issue #755, opened 2020-05-23
How to get a table name from the generated model?


righ:
### What version of SQLBoiler are you using (`sqlboiler --version`)?
SQLBoiler v3.4.0

### What is your database and version (eg. Postgresql 10)
MySQL 5.7.29

### Question
How do I get a real table name from the generated model?
I could not find the feature and It seems newer version also does not provide it. (probably)

I think it would work if I add the following code to templates. Is this the right thing to do? 

```go
func (_ {{$alias.UpSingular}}) GetTableName() string { return "{{.Table.Name}}" }
```

Also, I want it to be supported as standard. if I make the PR here, is there a possibility that you accept it?

Darkclainer:
I think what you want is presented in `singleton/boil_table_names.go.tpl` template. 

This correspond to `boil_table_names.go` generated files. There is var `TableNames` where all table from your schema are present.

righ:
Thanks for your reply.

`TableNames` surely have all names as you said, but it is related to not singular name and it can not be retrieved from the model. Let me know if you know the way to get a name from specified model using the `TableNames`.

For example, if I have a `jets` table and a `Jet` model.
I want a name `"jets"` in the following way:

```go
jet := Jet{ID: 1}
tableName := jet.GetTableName()
// use the table name
fmt.Println("ID:", jet.ID, "table:", tableName)
```

Darkclainer:
Hm, I have misunderstood what you want. I don't think sqlboiler have a builtin capability to do what you want.

Also I think your idea to include additional method in the template is perfectly OK, I have done this too for validating user's request (if a requested column exists for speicified table/model?). But be aware that adding your custom template is not very "neat" solution, because you will need to copy all original templates (remeber that you need to bother with versions) somewhere, specify them and your additional template folder with `--templates`. I have done this through building special docker container, than moved to my own fork of sqlboiler (because I have needed changes beyond templates). 

I don't know if your request will be accepted or not, but I think we can do best, if we will provide an easier way to specify additional templates (without removing original ones). This is not very easy to implement, but I think it can be worth it.

righ:
In fact, I don't want to handle it with custom template for that reason.
Because our project is working in go modules mode, it is hard to get the standard template paths for each  environments. And I do not want to make a directory copy.
So I agree with your suggestion. I'll make the option when I have much time.

[...]

PR #768, opened 2020-06-06
Add --extra-templates option

righ:
#755

This is a change to add the `--extra-templates` option.
It works the same way as `--templates`, but we don't need to specify original templates.

The following is a log of the test run.

Preparing:

```bash
# build a binary at my sqlboiler repository dir.
root@517e8bbfecb1:/sqlboiler# go build -o sqlboiler .
# copy
root@517e8bbfecb1:/sqlboiler# cp sqlboiler /root/
root@517e8bbfecb1:/sqlboiler# cd /root/

root@517e8bbfecb1:~# mkdir templates

root@517e8bbfecb1:~# cat << EOL > templates/99_tablename.go.tpl
> {{- \$alias := .Aliases.Table .Table.Name -}}
>
> // GetTableName returns the real table name.
> func (_ {{\$alias.UpSingular}}) GetTableName() string { return "{{.Table.Name}}" }
> EOL

root@517e8bbfecb1:~# cat templates/99_tablename.go.tpl

// GetTableName returns the real table name.
func (_ {{$alias.UpSingular}}) GetTableName() string { return "{{.Table.Name}}" }
```

Execution:

```bash
root@517e8bbfecb1:~# ./sqlboiler
Error: must provide a driver name

SQL Boiler generates a Go ORM from template files, tailored to your database schema.
Complete documentation is available at http://github.com/volatiletech/sqlboiler

Usage:
  sqlboiler [flags] <driver>

Examples:
sqlboiler psql

Flags:
      --add-global-variants        Enable generation for global variants
      --add-panic-variants         Enable generation for panic variants
      --add-soft-deletes           Enable soft deletion by updating deleted_at timestamp
  -c, --config string              Filename of config file to override default lookup
  -d, --debug                      Debug mode prints stack traces on error
      --extra-templates strings    The additional templates directory, that does not override the bindata's template folders in sqlboiler
  -h, --help                       help for sqlboiler
      --no-auto-timestamps         Disable automatic timestamps for created_at/updated_at
      --no-back-referencing        Disable back referencing in the loaded relationship structs
      --no-context                 Disable context.Context usage in the generated code
      --no-driver-templates        Disable parsing of templates defined by the database driver
      --no-hooks                   Disable hooks feature for your models
      --no-rows-affected           Disable rows affected in the generated API
      --no-tests                   Disable generated go test files
  -o, --output string              The name of the folder to output to (default "models")
  -p, --pkgname string             The name you wish to assign to your generated package (default "models")
  -r, --relation-tag string        Relationship struct tag name (default "-")
      --struct-tag-casing string   Decides the casing for go structure tag names. camel, title or snake (default snake) (default "snake")
  -t, --tag strings                Struct tags to be included on your models in addition to json, yaml, toml
      --tag-ignore strings         List of column names that should have tags values set to '-' (ignored during parsing)
      --templates strings          A templates directory, overrides the bindata'd template folders in sqlboiler
      --version                    Print the version
      --wipe                       Delete the output folder (rm -rf) before generation to ensure sanity

root@517e8bbfecb1:~# ./sqlboiler mysql --output models_mysql --pkgname models_mysql --wipe --extra-templates ./templates/

root@517e8bbfecb1:~# tail models_mysql/jets.go  -n20
        var exists bool
        sql := "select exists(select 1 from `jets` where `id`=? limit 1)"

        if boil.IsDebug(ctx) {
                writer := boil.DebugWriterFrom(ctx)
                fmt.Fprintln(writer, sql)
                fmt.Fprintln(writer, iD)
        }
        row := exec.QueryRowContext(ctx, sql, iD)

        err := row.Scan(&exists)
        if err != nil {
                return false, errors.Wrap(err, "models_mysql: unable to check if jets exists")
        }

        return exists, nil
}

// GetTableName returns the real table name.
func (_ Jet) GetTableName() string { return "jets" }
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -418,6 +418,7 @@ Flags:
       --add-soft-deletes           Enable soft deletion by updating deleted_at timestamp
   -c, --config string              Filename of config file to override default lookup
   -d, --debug                      Debug mode prints stack traces on error
+      --extra-templates string     The additional templates directory, that does not override the bindata's template folders in sqlboiler
   -h, --help                       help for sqlboiler
       --no-auto-timestamps         Disable automatic timestamps for created_at/updated_at
       --no-back-referencing        Disable back referencing in the loaded relationship structs
@@ -735,12 +736,21 @@ output_dir/
 
 **Note**: Because the `--templates` flag overrides the internal bindata of `sqlboiler`, if you still
 wish to generate the default templates it's recommended that you include the path to sqlboiler's templates
-as well.
+as well. 
+Either you can use `--extra-templates` to specify additional (to those included in internal bindata) templates directories.
 
 ```toml
 templates = [
   "/path/to/sqlboiler/templates",
   "/path/to/sqlboiler/templates_test",
+  "/path/to/your_project/more_templates"
+]
+```
+
+Or
+
+```toml
+extra_templates = [
   "/path/to/your_project/more_templates"
 ]
 ```
--- a/boilingcore/boilingcore.go
+++ b/boilingcore/boilingcore.go
@@ -251,6 +251,23 @@ func (s *State) initTemplates() ([]lazyTemplate, error) {
 			}
 		}
 	}
+	if len(s.Config.ExtraTemplateDirs) != 0 {
+		for _, dir := range s.Config.ExtraTemplateDirs {
+			abs, err := filepath.Abs(dir)
+			if err != nil {
+				return nil, errors.Wrap(err, "could not find abs dir of templates directory")
+			}
+
+			base := filepath.Base(abs)
+			root := filepath.Dir(abs)
+			tpls, err := findTemplates(root, base)
+			if err != nil {
+				return nil, err
+			}
+
+			mergeTemplates(templates, tpls)
+		}
+	}
 
 	if !s.Config.NoDriverTemplates {
 		driverTemplates, err := s.Driver.Templates()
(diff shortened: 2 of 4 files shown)


Darkclainer:
Don't you mind if I will amend README.md as additional commit in this PR?

righ:
@Darkclainer Of cource I do not. Thanks!

Darkclainer:
I have found that PR should be opened against dev branch: <https://github.com/volatiletech/sqlboiler/blob/master/CONTRIBUTING.md>

righ:
Okay, thanks for letting me know.
I changed the base and rebased this branch into `dev`.

And I fixed your commit slightly to match the current code.

aarondl:
Hello @righ. I appreciate the PR! I don't think it's a good idea to add a second option for templates though. I quite like the explicitness of the current system and I think it's also important to ensure that sqlboiler upgrades don't change your main templates, since your overriding/extra templates can depend on things declared in those and there are many unexported things we could currently change that may break your templates.

I could possibly be convinced but as I see it there's no difference between:

```
--extra-templates my_dir
```
and
```
--templates my_sqlboiler_templates --templates my_dir
```

The second isn't really onerous enough for me to want to accept this. Besides if you're doing it from a Go module path then you can use: `$GOPATH/pkg/mod/github.com/volatiletech/sqlboiler@version` as a way of maintaining compatibility which is also nice.

Going to close for now.

jaakkotuosa:
In case someone else is checking this issue, I would like to mention a more complete example of including the default templates:
```--templates $GOPATH/pkg/mod/github.com/volatiletech/sqlboiler/v4@v4.1.2/templates --templates $GOPATH/pkg/mod/github.com/volatiletech/sqlboiler/v4@v4.1.2/templates_test```

Issue #273, opened 2018-04-09
MySQL/MariaDB ssl-mode (5.6 incompatibility)


aarondl:
Originally from #262:

> In the generated code main_test.go, the makeOptionFile method provides the credentials and params for the db connection, fmt.Fprintf(tmp, "ssl-mode=%s\n", m.sslMode(m.sslmode)) will try to add the ssl-mode based on the settings. However, the DISABLED and PREFERRED are only supported in mysql 5.7 and they will be rejected in mysql 5.6. You can refer to the doc here.

I've also noticed that MariaDB Server and MySQL Client are incompatible in this way.

From 5.6 Docs:

> This option is available only for client programs, not the server. It specifies the security state of the connection to the server:
> If this option is not specified, the default is to establish an unencrypted connection. This is like the --ssl=0 option or its synonyms (--skip-ssl, --disable-ssl).
> If this option is specified, the only permitted value is REQUIRED (establish an encrypted connection if the server supports encrypted connections). The connection attempt fails if an encrypted connection cannot be established.

From 5.7 Docs:

> This option is available only for client programs, not the server. It specifies the security state of the connection to the server. These option values are permitted:
> PREFERRED: Establish an encrypted connection if the server supports encrypted connections, falling back to an unencrypted connection if an encrypted connection cannot be established. This is the default if --ssl-mode is not specified.
> REQUIRED: Establish an encrypted connection if the server supports encrypted connections. The connection attempt fails if an encrypted connection cannot be established.
> VERIFY_CA: Like REQUIRED, but additionally verify the server Certificate Authority (CA) certificate against the configured CA certificates. The connection attempt fails if no valid matching CA certificates are found.
> VERIFY_IDENTITY: Like VERIFY_CA, but additionally perform host name identify verification by checking the server's Common Name identity in the certificate that the server sends to the client. The client verifies the Common Name against the host name the client uses for connecting to the server, and the connection fails if there is a mismatch. For encrypted connections, this option helps prevent man-in-the-middle attacks. This is like the legacy --ssl-verify-server-cert option.
> DISABLED: Establish an unencrypted connection. This is like the legacy --ssl=0 option or its synonyms (--skip-ssl, --disable-ssl).

Currently SQLBoiler is using 5.7 configuration options:
https://github.com/volatiletech/sqlboiler/blob/master/templates_test/main_test/mysql_main.tpl#L72

[...]

PR #769, opened 2020-06-07
Increased control over ssl-mode in MySQL driver

tooolbox:
This works around #273 by improving control over the generated `ssl-mode` flag.

Previously, there were three options:
- Specify `true` in your config file, which resulted in `ssl-mode=REQUIRED`
- Specify `false` in your config file, which resulted in `ssl-mode=DISABLED`
- Specify anything else in your config file, which resulted in `ssl-mode=PREFERRED`
- Leave your config file blank, which Viper would default to `true` and therefore `ssl-mode=REQUIRED`

Now the driver allows directly specifying `REQUIRED`, `DISABLED`, or other previously unavailable options such as `VERIFY_CA` or `VERIFY_IDENTITY`.  The legacy values `true` and `false` still work the same as they did before.

The value `PREFERRED` is specially treated by omitting the `ssl-mode` setting entirely.  This is fine because `PREFERRED` is the default, so the behavior is equivalent.  As a side effect, using `PREFERRED` is a workaround for MariaDB compatibility until https://jira.mariadb.org/browse/MDEV-22129 is handled.

-------------------

This seemed like the best way to go about this because:
- It's backwards compatible.
- It increases the range of options.
- It gives an option for MariaDB users to run tests without editing the generated files.

### DIFF
--- a/drivers/sqlboiler-mysql/driver/override/templates_test/singleton/mysql_main_test.go.tpl
+++ b/drivers/sqlboiler-mysql/driver/override/templates_test/singleton/mysql_main_test.go.tpl
@@ -108,8 +108,10 @@ func (m *mysqlTester) sslMode(mode string) string {
 		return "REQUIRED"
 	case "false":
 		return "DISABLED"
+	case "PREFERRED":
+		return ""
 	default:
-		return "PREFERRED"
+		return mode
 	}
 }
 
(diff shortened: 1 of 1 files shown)


tooolbox:
Sorry, this is slightly premature.  Will reopen when I've checked a couple things.

tooolbox:
Okay, much better.  Bindata included, tested, etc.

Issue #775, opened 2020-06-13
Update updated_at on soft delete (deleted_at)


Darkclainer:
If you're having a generation problem please answer these questions before submitting your issue. Thanks!

### What version of SQLBoiler are you using (`sqlboiler --version`)?

SQLBoiler v4.1.2

### What is your database and version (eg. Postgresql 10)

Postgresql 9.6

### If this happened at runtime what code produced the issue? (if not applicable leave blank)

```go
object, err := pmodel.FindObject(ctx, db, objectID)
if err != nil {
    return err
}
_, err = object.Delete(ctx, pdb.D, false)
```

### Please provide a relevant database schema so we can replicate your issue (Provided you are comfortable sharing this)

```sql
CREATE TABLE IF NOT EXISTS objects (
     id SERIAL PRIMARY KEY,
     updated_at TIMESTAMP WITHOUT TIME ZONE,
     deleted_at TIMESTAMP
)
```

### Further information. What did you do, what did you expect?

Field updated_at updated with deleted_at simultaneously.

I don't know if this is a bug or not from your perspective, but if it is, I'm glad to provide PR for this issue.

aarondl:
It's a good question. If you update `updated_at` you've technically lost information. Since if you see `deleted_at` purely as metadata then having `deleted_at` set as `now()` and `updated_at` set as whenever an actual update was made to actual columns (like `name`, or `age` or whatever) you retain the ability to differentiate. Unfortunately then if you want the behavior where deleting it counts as updating then you need to do: a max(deleted_at, updated_at) to find the last time _any_ change was made.

I'm leaning towards doing no changes. What do you think about the issue?

Darkclainer:
Actually I think this behaviour can be OK, or can be considered a bug.

For me this is a bug, because semantically (for me) 'delete' is a subset of 'update' in sense of working with data/ But it is a single opinion.

There is another side of this issue, that come to me recently: if people are considering current behaviour as well defined, then if we change it, it can break their code. So either way this should be implemented as option (and enabled in new major version?).

Maybe the best way to deal with it is just describe this behaviour in README.md.

aarondl:
@Darkclainer I agree with everything you've said. I will document this in the README.

PR #776, opened 2020-06-13
Update updated at on soft delete

Darkclainer:
See #775


aarondl:
Closing as we've closed the related issue. Thanks for opening it in the first place!

Issue #576, opened 2019-07-13
ForeignFN conflicts with ColumnName


bugzpodder:
I am seeing a conflict in all my foreign keys for these two function.
I can see that if you have user_id as a foreign key, you would get UserId as field and User() as foreign function?

ForeignFn: https://github.com/volatiletech/sqlboiler/blob/7d3d954a4be10dc06b9464fb3e6d74836274c467/boilingcore/text_helpers.go#L40

ColumnName:
https://github.com/volatiletech/sqlboiler/blob/4dea9a5f77e3ec6090ad455df3d408e47e077700/boilingcore/aliases.go#L78

I found the following logic help:
```
       if foreignFn == fkColumnName {
		if strings.HasPrefix(foreignFn, foreignTableName) {
			foreignFn = foreignTableName
		} else {
			foreignFn += foreignTableName
		}
	}
```
So if you have CreatedBy to the User table, it would generate foreignn fn CreatedByUser()
and if you have UserIdentifier on the User table, foreign fn would be User()

I know there are aliases in the config file, but when every table have a User field called created_by, it seems to be troublesome to have set same alias across 100 tables.  If aliases work from the destination table or there is a global list of aliases that would be great as well.

aarondl:
This sort of logic is already present. If there is no short-form available and it would cause a conflict then it uses the foreign key name + the other table name. Take this example in the documentation of `txtNameToMany`:

```
// industries - industries
// industry_id  mapped_industry_id
// fk == table = industry.Industries
// fk != table = industry.MappedIndustryIndustry
```

Not pretty but it should avoid any problems. Is it perhaps that you're using the uppercase fork that you've made (the PR) and you're running into equality troubles or something?

bugzpodder:
Let's just talk about the lower case scenario.  I was rambling a bit, but I think the issue is this:
1) Completely different names.  Consider you have books table and it has author column, pointing to users.ID

then author column would both be field (Author) and foreign function Author().
Similarly if you have a "created_by" column pointing to users.ID
or perhaps a non standard suffix, so instead of user_id, you had user_name, then field (UserName) and function UserName() would conflict.

This happens when trimSuffixes does nothing, ie when your columns doesn't end with _id.

After my change, you would have:
Author, AuthorUser()
CreatedBy, CreatedByUser()
UserName, User()

[...]

PR #780, opened 2020-06-20
Add some aliases to fix conflicting struct field name and method name

yuzuy:
Fix for #576 
Add some aliases for foreignFn to fix conflicting struct field name and method name when using the same name as the foreign table to the column name or using column name has the suffix "_by".

For examples:
```sql
create table users (
	id int primary key,
	name varchar(255)
);

create table videos (
	id int primary key,
	name varchar(255),
	created_by int,
	constraint fk_videos_users_created_by foreign key (created_by) references users (id)
);
```
foreignFn -> CreatedByUser

```
- created_by int
+ user int
```
foreignFn -> VideoUser

In detail, refer https://github.com/volatiletech/sqlboiler/issues/576#issuecomment-526716185


boilingcore/text_helpers.go
```diff
@@ -38,12 +38,23 @@ import (
...
+	fkColumnTrimmedSuffixes := strmangle.Singular(trimSuffixes(fk.Column))
+	fkeyNotTableName := fkColumnTrimmedSuffixes != strmangle.Singular(fk.ForeignTable)
+	singularForeignTable := strmangle.Singular(fk.ForeignTable)
+
+	if fkColumnTrimmedSuffixes == singularForeignTable {
+		foreignFn = strmangle.TitleCase(strmangle.Singular(fk.Table) + "_" + fkColumnTrimmedSuffixes)
+		if fk.Column != singularForeignTable {
+			foreignFn = strmangle.TitleCase(fkColumnTrimmedSuffixes)
+		}
+	} else if fkColumnTrimmedSuffixes == fk.Column && strings.HasSuffix(fk.Column, "_by") {
```
aarondl, line 50: This is not exactly what was spec'd out. A `_by` suffix doesn't mean anything in particular and is just an example, we should not be giving special behaviors to columns with this suffix.

A PR that fixes this correctly would first start with the code that bugzpodder suggested. This should correct the two conflicts in bold in the table. Then some additional code should be added to avoid the `UserUser` name (it should end up as `VideoUser` as discussed below the table).

yuzuy:
@aarondl I'm sorry I'm late for fixing.
About go-bindata/different, I saw CONTRIBUTING.md, but I could not understand which version should use, so I used the version that got by "go get -u github.com/kevinburke/go-bindata/...".
If I was wrong, please tell me the version.

PR #781, opened 2020-06-25
Additional fix for MySQL soft delete code generation

mfzy602:
First of all, I'd like to thank you for your hard work.
I've just begun working on a project using SQLBoiler but its ease of use made me love it already.

This PR addresses an issue that was supposed to have been fixed in v4.1.2.
Even after upgrading to v4.1.2, applications using RDBMS such as MySQL are still affected by the issue since DeleteAll code generated for model slices remains to suffer from the very problem (partially) addressed by <https://github.com/volatiletech/sqlboiler/commit/e7b6e07c3d0de5d1ead51192db29fc9880016e35>.

Since the fix mentioned above apparently didn't have an accompanying test or something, I didn't try to guarantee that this fix works. (TBH, I have no idea how.)

### Preconditions

- Soft deletes are enabled via configuration (`add-soft-deletes`).
- There's at least one table that is [capable of soft deletion](https://github.com/volatiletech/sqlboiler/blob/50beb2f5634a449d07cdb2fe7ee2638f08b6be65/drivers/table.go#L68-L75).
- Using an RDBMS whose database driver doesn't `UseIndexPlaceholders`.

### Expected behavior

Generated code should not produce SQL containing placeholders like `$1`.

### Actual behavior

Generated test fails with message like following:
``` 
--- FAIL: TestSliceSoftDeleteAll (0.00s)
    --- FAIL: TestSliceSoftDeleteAll/Products (0.00s)
        product_test.go:120: db: unable to delete all from product slice: Error 1054: Unknown column '$1' in 'where clause'
        product_test.go:131: want zero records, got: 1
```

### DIFF
--- a/templates/18_delete.go.tpl
+++ b/templates/18_delete.go.tpl
@@ -273,7 +273,7 @@ func (o {{$alias.UpSingular}}Slice) DeleteAll({{if .NoContext}}exec boil.Executo
 		}
 		wl := []string{"deleted_at"}
 		sql = fmt.Sprintf("UPDATE {{$schemaTable}} SET %s WHERE " +
-			strmangle.WhereClauseRepeated(string(dialect.LQ), string(dialect.RQ), {{if .Dialect.UseIndexPlaceholders}}2{{else}}1{{end}}, {{$alias.DownSingular}}PrimaryKeyColumns, len(o)),
+			strmangle.WhereClauseRepeated(string(dialect.LQ), string(dialect.RQ), {{if .Dialect.UseIndexPlaceholders}}2{{else}}0{{end}}, {{$alias.DownSingular}}PrimaryKeyColumns, len(o)),
 			strmangle.SetParamNames("{{.LQ}}", "{{.RQ}}", {{if .Dialect.UseIndexPlaceholders}}1{{else}}0{{end}}, wl),
 		)
 		args = append([]interface{}{currTime}, args...)


aarondl:
> Since the fix mentioned above apparently didn't have an accompanying test or something, I didn't try to guarantee that this fix works. (TBH, I have no idea how.)

I just looked in the generated code to ensure that it was doing the right thing which is using a `?` in the sprintf as a placeholder, not $1.

The line you changed changes the start index that's given to `strmangle.WhereClause` which (although not documented in the comments for some reason), which returns a `?` for 0, and the same for successive arguments. Whereas anything else would give `$start, $start+1, $start+2` for things like postgres.

I'm not comfortable merging this until you're happy that you've been able to validate it :)

mfzy602:
Ah, my bad. That's not what I meant. I think I should've phrased more carefully.

> The line you changed changes ...

I perfectly knew what I was doing here because I've read through generated code to strmangle one that you just pointed out and have figured out how it works. (Anyway, thanks for the explanation. I feel reassured :) )

As for the validation, I've incorporated this fix into my codebase beforehand and confirmed that it indeed worked and generated tests got to pass. I just wanted to let you know that I chose not to include anything to guarantee the validity of this fix for CI (since I couldn't refer to the *prior* fix for how).

aarondl:
Thanks for this PR!

PR #784, opened 2020-07-02
Make binding cache accesses allocation-free

zikaeroh:
This is a followup to #706, which removed the cache key creation for the `reflect.Type`, but left the column key as is.

This PR optimizes the binding cache further, eliminating all allocations from the fast path by using the column key's byte slice as the map key (as string conversions in map accesses do not allocate).

Comparing `boilbench` shows a minor memory improvement and no significant change to time:

```
name                      old time/op    new time/op    delta
BoilDelete/boil-4            661ns ± 1%     664ns ± 0%  +0.52%  (p=0.033 n=8+9)
BoilInsert/boil-4           2.21µs ± 2%    2.20µs ± 1%    ~     (p=0.448 n=9+9)
BoilRawBind/boil-4          9.18µs ± 0%    9.17µs ± 1%    ~     (p=0.604 n=9+10)
BoilSelectAll/boil-4        9.06µs ± 1%    9.02µs ± 0%  -0.52%  (p=0.001 n=9+9)
BoilSelectSubset/boil-4     9.72µs ± 1%    9.66µs ± 1%  -0.62%  (p=0.001 n=9+10)
BoilSelectComplex/boil-4    12.3µs ± 1%    12.3µs ± 1%    ~     (p=0.812 n=10+10)
BoilUpdate/boil-4           1.84µs ± 2%    1.83µs ± 1%    ~     (p=0.855 n=10+9)

name                      old alloc/op   new alloc/op   delta
BoilDelete/boil-4             168B ± 0%      168B ± 0%    ~     (all equal)
BoilInsert/boil-4             968B ± 0%      968B ± 0%    ~     (all equal)
BoilRawBind/boil-4          4.30kB ± 0%    4.30kB ± 0%    ~     (all equal)
BoilSelectAll/boil-4        2.93kB ± 0%    2.86kB ± 0%  -2.20%  (p=0.000 n=10+10)
BoilSelectSubset/boil-4     3.07kB ± 0%    3.01kB ± 0%  -2.13%  (p=0.000 n=10+10)
BoilSelectComplex/boil-4    3.99kB ± 0%    3.93kB ± 0%  -1.65%  (p=0.000 n=10+10)
BoilUpdate/boil-4             920B ± 0%      920B ± 0%    ~     (all equal)

name                      old allocs/op  new allocs/op  delta
BoilDelete/boil-4             8.00 ± 0%      8.00 ± 0%    ~     (all equal)
BoilInsert/boil-4             21.0 ± 0%      21.0 ± 0%    ~     (all equal)
BoilRawBind/boil-4            47.0 ± 0%      47.0 ± 0%    ~     (all equal)
BoilSelectAll/boil-4          48.0 ± 0%      47.0 ± 0%  -2.08%  (p=0.000 n=10+10)
BoilSelectSubset/boil-4       52.0 ± 0%      51.0 ± 0%  -1.92%  (p=0.000 n=10+10)
BoilSelectComplex/boil-4      78.0 ± 0%      77.0 ± 0%  -1.28%  (p=0.000 n=10+10)
BoilUpdate/boil-4             16.0 ± 0%      16.0 ± 0%    ~     (all equal)
```

However, in my project, I often query models/tables with many columns, so the eliminating the column key allocation offers a more significant memory improvement:

```
name                   old time/op    new time/op    delta
HandleNop-4               381µs ± 1%     378µs ± 1%  -0.61%  (p=0.010 n=8+8)
HandleNopParallel-4       266µs ± 0%     265µs ± 1%  -0.40%  (p=0.015 n=8+8)
HandleCustomCommand-4     391µs ±10%     379µs ± 1%    ~     (p=0.277 n=9+8)
HandleMixed-4             391µs ± 5%     392µs ± 1%    ~     (p=0.780 n=9+10)

name                   old alloc/op   new alloc/op   delta
HandleNop-4              15.0kB ± 0%    14.2kB ± 0%  -5.14%  (p=0.000 n=9+9)
HandleNopParallel-4      15.1kB ± 0%    14.3kB ± 0%  -5.12%  (p=0.000 n=9+10)
HandleCustomCommand-4    15.0kB ± 0%    14.2kB ± 0%  -5.12%  (p=0.000 n=10+9)
HandleMixed-4            15.1kB ± 0%    14.3kB ± 0%  -5.13%  (p=0.000 n=10+7)

name                   old allocs/op  new allocs/op  delta
HandleNop-4                 259 ± 0%       258 ± 0%  -0.39%  (p=0.000 n=10+10)
HandleNopParallel-4         259 ± 0%       258 ± 0%  -0.39%  (p=0.000 n=10+10)
HandleCustomCommand-4       259 ± 0%       258 ± 0%  -0.39%  (p=0.000 n=10+10)
HandleMixed-4               260 ± 0%       259 ± 0%  -0.38%  (p=0.000 n=10+10)
```

That's 5% fewer bytes for a single allocation.

### DIFF
--- a/queries/reflect.go
+++ b/queries/reflect.go
@@ -17,17 +17,6 @@ import (
 	"github.com/volatiletech/strmangle"
 )
 
-type colBindingKey struct {
-	typ     reflect.Type
-	colsKey string
-}
-
-var (
-	mut         sync.RWMutex
-	bindingMaps = make(map[colBindingKey][]uint64)
-	structMaps  = make(map[reflect.Type]map[string]uint64)
-)
-
 // Identifies what kind of object we're binding to
 type bindKind int
 
@@ -235,34 +224,9 @@ func bind(rows *sql.Rows, obj interface{}, structType, sliceType reflect.Type, b
 		ptrSlice = reflect.Indirect(reflect.ValueOf(obj))
 	}
 
-	var strMapping map[string]uint64
-	var sok bool
-	var mapping []uint64
-	var ok bool
-
-	colsKey := makeColsKey(structType, cols)
-
-	mut.RLock()
-	mapping, ok = bindingMaps[colsKey]
-	if !ok {
-		if strMapping, sok = structMaps[structType]; !sok {
-			strMapping = MakeStructMapping(structType)
-		}
-	}
-	mut.RUnlock()
-
-	if !ok {
-		mapping, err = BindMapping(structType, strMapping, cols)
-		if err != nil {
-			return err
-		}
-
-		mut.Lock()
-		if !sok {
-			structMaps[structType] = strMapping
-		}
-		bindingMaps[colsKey] = mapping
-		mut.Unlock()
+	mapping, err := getMappingCache(structType).mapping(cols)
+	if err != nil {
+		return err
 	}
 
 	var oneStruct reflect.Value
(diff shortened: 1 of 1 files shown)


queries/reflect.go
```diff
@@ -441,18 +405,69 @@ func getBoilTag(field reflect.StructField) (name string, recurse bool) {
...
+	}
+}
+
+func (b *mappingCache) mapping(cols []string) ([]uint64, error) {
 	buf := strmangle.GetBuffer()
+	defer strmangle.PutBuffer(buf)
+
 	for _, s := range cols {
 		buf.WriteString(s)
+		buf.WriteByte(0)
```
aarondl, line 450: I'm concerned about this line. One of the reasons to write the string is to ensure that the order of the columns themselves is correct, not just the number of columns.

`id, username` and `username, id` as columns would now cause a cache key collision and could potentially result in the mapping returned being pointers to the wrong things.

zikaeroh: I'm confused by this; this code is identical to the original code except that I've added NUL between the columns: https://github.com/volatiletech/sqlboiler/blob/0116414cbb4d642dfef515aeec1e54ebf60317ea/queries/reflect.go#L447

I'm still writing the strings, in the same order.

I can remove the NUL to preserve the original logic, but I added it because one could technically have three columns "foo", "bar", and "foobar", and two queries asking for `["foo", "bar"]` and `["foobar"]` could collide if the column names are coming in unquoted. But that's already true now.

zikaeroh: I can remove it and look into submitting a PR with it and a regression test, if that's preferred, but I'm definitely still writing the column names in the same order.

aarondl: I'm confused by my own comment. For some reason I imagined a `-` in front of that line. My bad! This is indeed an improvement.

zikaeroh: Thanks for re-looking!

aarondl:
Excellent work. Gratefully accepted.

PR #790, opened 2020-07-07
Prevent TestToManyAdd failure due to unique constraint

sadayuki-matsuno:
fixed below.
 
```
--- FAIL: TestToManyAdd (3.45s)
    --- FAIL: TestToManyAdd/XxxToXxxs (0.02s)
        xxxs_test.go:5308: failed to insert into foreign table: db: unable to insert into xxxs: pq: duplicate key value violates unique constraint "xxx_constraint"
```

This error occurred when the table column of character was unique.
Since one character is selected from 62 characters of random character [0-9a-zA-Z], when the value of seed becomes 31 intervals or 62 intervals, an error occurred due to a unique constraint violation.

Therefore, for the TestToManyAdd test, I changed the seed value to a continuous value and I fixed it so that collisions are less likely to occur.

I hope you'll check and merge this.


sadayuki-matsuno:
Thank you for reviewing. The two points pointed out have been corrected.

> gofmt needs to be run on this code (there's spaces not tabs in the code)

I inserted the tab manually. Is there a way to gofmt .tpl file? Please let me know if there is any good way.

> There should not be an additional dependency added in go.mod as this change requires no new dependencies.

When I execute the following two commands, the file is rewritten automatically. I also manually checked out and restored it.

`./boil.sh go-generate all` 
`./boil.sh build all`

aarondl:
I'm trying to understand the fix you've provided. You're saying that the seed is being exhausted too quickly because it's being used for other random things, which causes it to cycle through the 62 characters too quickly landing on a collision eventually.

What this PR does is sort of against what the idea of the seed was. Since on the first call your seed value is 0. So then we use 0-10 say for the first `.Randomize` call. Then you reset it to 1. So the next call uses `1-11`. The point of the seed was to avoid collisions by trying to ensure that we don't repeat things.

Perhaps this is just a problem occurring because your table has a large number of columns that's eating too many of the seed's values causing it to wrap to the first character of the alphabet again too quickly? Could you tell me how many columns your table has?

I'm not really keen on merging this now that I've had a good think about it. Because this could cause other failures depending on the schema of people. For example if I had a table like this:

```create table test (a text unique, b text unique)```

Then this change would cause this table to fail.

The end result here is that the tests are simply unreliable and cannot deal with constraints properly. It's best to ignore these sorts of failures and not try to fix them. Unfortunately.

**Comment by sadayuki-matsuno** [comment]
@aarondl 
I'm sorry to reply late.
We have provided an example of a table that is currently not tested successfully due to a seed collision error.

```
CREATE TABLE parents(
        id BIGSERIAL PRIMARY KEY NOT NULL
);

CREATE TYPE enum AS ENUM (
    '1',
    '2',
    '3',
    '4'
);

CREATE TABLE children(
        parent_id bigint NOT NULL REFERENCES parents(id),
        field1 text,
        enum enum NOT NULL,
        field2 character varying(100),
        field3 character varying(100),
        field4 timestamp without time zone,
        field5 boolean DEFAULT false,
        created_at timestamp without time zone DEFAULT statement_timestamp() NOT NULL,
        updated_at timestamp without time zone DEFAULT statement_timestamp() NOT NULL,
        CONSTRAINT children_parent_id_enum_constraint UNIQUE(parent_id, enum)
);
```

This is just one sample and I'm facing test case errors due to seed conflicts several times with different numbers of columns and different data types.

sadayuki-matsuno:
What are the cases where an error occurs in my PR? The DDL case `create table test (a text unique, b text unique);` you provided does not have a Primary key, so it seems that an error will occur, but I would be grateful if you could let me know if there are any cases to consider.

aarondl:
Hi @sadayuki-matsuno there's no errors, but at this time I don't think it's a good idea to merge this PR. It changes the fundamental assumptions about how seeds are supposed to work.

The tests are a mistake. When we created them it was in naivety, if they do not pass it doesn't indicate much and they should be ignored, I think I will remove them in the next version of sqlboiler, they tend to cause more trouble than they're worth.

PR #795, opened 2020-07-09
Sorting of columns used in tempalte.

sadayuki-matsuno:
I implemented sorting of columns used in tempalte.
With this, it is possible to control the order of columns when code generation is performed using sqlboiler template.

Especially for postgres, changing the actual column order is difficult and not realistic, so doing it with sqlboiler makes code generation easier to use.

For example, if there is a Users table like the one below,
```
CREATE TABLE users(
        name text,
        code text,
        created_at timestamp without time zone DEFAULT statement_timestamp() NOT NULL,
        updated_at timestamp without time zone DEFAULT statement_timestamp() NOT NULL,
        deleted_at timestamp without time zone,
);
```
 I want to create a function called NewUsers using sqlboiler.

```
// NewUser :
func NewUser(
	id int64,
	name string,
	code string,
	created_at time.Time,
	updated_at time.Time,
	deleted_at null.Time,
) *User {
	return &User{
		ID:        id,
		Name:      name,
		Code:      code,
		CreatedAt: createdAt,
		UpdatedAt: updatedAt,
		DeletedAt: deletedAt,
	}
}
```

In this situation, when I add a new column called mail_address, I want that column to be inserted under code. 
That's why I made this PR.

The setting of order is as follows.

```
[orders]
  id = -1
  created_at = 3
  updated_at = 4
  deleted_at = 5
```
Columns for which no order is specified defaults to 0, and this number is used to sort the columns in ascending order.

### DIFF
--- a/boilingcore/boilingcore.go
+++ b/boilingcore/boilingcore.go
@@ -137,6 +137,7 @@ func (s *State) Run() error {
 	data := &templateData{
 		Tables:            s.Tables,
 		Aliases:           s.Config.Aliases,
+		Orders:            s.Config.Orders,
 		DriverName:        s.Config.DriverName,
 		PkgName:           s.Config.PkgName,
 		AddGlobal:         s.Config.AddGlobal,
@@ -191,6 +192,7 @@ func (s *State) Run() error {
 		}
 
 		data.Table = table
+		data.Table.Columns = data.Orders.sortColumns(data.Table.Columns)
 
 		// Generate the regular templates
 		if err := generateOutput(s, regularDirExtMap, data); err != nil {
--- a/boilingcore/config.go
+++ b/boilingcore/config.go
@@ -39,6 +39,7 @@ type Config struct {
 
 	Aliases      Aliases       `toml:"aliases,omitempty" json:"aliases,omitempty"`
 	TypeReplaces []TypeReplace `toml:"type_replaces,omitempty" json:"type_replaces,omitempty"`
+	Orders       Orders        `toml:"orders,omitempty" json:"orders,omitempty"`
 
 	Version string `toml:"version" json:"version"`
 }
@@ -256,4 +257,22 @@ func columnFromInterface(i interface{}) (col drivers.Column) {
 	}
 
 	return col
+}
+
+// ConvertOrders is necessary because viper
+// All unspecified columns default to 0
+//   [orders]
+//     id = -1
+//     created_at = 3
+//     updated_at = 4
+//     deleted_at = 5
+//
+func ConvertOrders(i interface{}) (o Orders) {
+	if i == nil {
+		return o
+	}
+	columns := cast.ToStringMapInt(i)
+	return Orders{
+		Columns: columns,
+	}
 }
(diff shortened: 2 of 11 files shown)


aarondl:
The columns are currently sorted in an order that ensures idempotent generation. If you generate, and generate again, the exact same output will always come out.

Could you elaborate on why you might want to control the generated struct's column order other than a general cleanliness of "this is how it looks in my head"?

sadayuki-matsuno:
When using the struct model generated by sqloboiler, the order of the fields does not matter if the value is entered in each field of the struct as follows.

```
user := &User{
		ID:        id,
		Name:      name,
		Code:      code,
		CreatedAt: createdAt,
		UpdatedAt: updatedAt,
		DeletedAt: deletedAt,
}
```

However, I want to generate a NewXXX fuction with sqlboiler and use the model through it instead of directly entering the value. 
This is because the compiler will tell you when the number of fields increases or when the type changes.

```
// NewUser :
func NewUser(
	id int64,
	name string,
	code string,
	created_at time.Time,
	updated_at time.Time,
	deleted_at null.Time,
) *User {
	return &User{
		ID:        id,
		Name:      name,
		Code:      code,
		CreatedAt: createdAt,
		UpdatedAt: updatedAt,
		DeletedAt: deletedAt,
	}
}
```

When actually using this NewXXX, it is difficult to understand what is added to which field, because it only packs the value in the argument.

```
	user := NewUser(
		0,
		"name",
		"code",
		time.Now(),
		time.Now(),
		null.TimeFrom(time.Now()),
	)
```

So, if there is some regularity in the order of the fields that are generated to some extent, it will be easier to understand where and which value is contained, and it will be easier to implement.

I found that sqlboiler can be used to automatically generate most sources from db column information.
I don't think this column order control is necessary for those who use the current sqlboiler functions, but I think it is necessary for those who want to use sqlboiler to generate other code.

sadayuki-matsuno:
@aarondl 
I'm sorry to reply late.
This is intentional. As shown in the sample, I think that the order of the columns I want to sort is common to all tables.
(ex. For example, putting id at the top, puting created_at and updated_at at the bottom, and so on.)

If necessary, it may be possible to implement so that the same definition can be made for individual tables, and give priority to the definition of individual tables.

aarondl:
You can achieve this same thing by using the syntax:

```
user.User{
	0,
	"name",
	"code",
	time.Now(),
	time.Now(),
	null.TimeFrom(time.Now()),
}
```

It may be slightly nicer to use this approach, but sqlboiler has so many features this doesn't feel like one more we need to add.

Specifically because sqlboiler itself isn't going to benefit from the `New` function because it's a bit clunky. A lot of values in a normal database object will have defaults and the way sqlboiler is designed is to take into account Go zero values and use those as placeholders to fill default values back into your struct. It feels clunky to have a constructor that you have to remember to put in the zero values for those types like `time.Time{}` so you get the proper behavior.

At this time I don't think we want to accept this because all of this functionality is possible today (adding the New function or creating the struct with unnamed properties) and because of our idempotent generation this will always function correctly. The desire to have a slightly nicer ordering on the values isn't a priority since as far as I know nobody is doing either of these two things today and there's been no other requests for it. I want to avoid adding features to sqlboiler where possible.

PR #797, opened 2020-07-13
fix the foreign key name

sobhanattar:
Based on the naming in line `558`, it seems that the key name in line `576` 
must be changed to `videos_user_id_fkey` from `videos_author_id_fkey`


sobhanattar:
I missed the comment

Issue #678, opened 2020-02-23
Unclear how to store passwords outside of version control


ORESoftware:
I see an example like so:

```
[psql]
  dbname = "dbname"
  host   = "localhost"
  port   = 5432
  user   = "dbusername"
  pass   = "dbpassword"
  schema = "myschema"
  blacklist = ["migrations", "other"]
```

what I want to do is use an env variable for the password:

```
[psql]
  dbname = "dbname"
  host   = "localhost"
  port   = 5432
  user   = "dbusername"
  pass   = "$dbpassword"   // inject the password from env var
  schema = "myschema"
  blacklist = ["migrations", "other"]
```

is this possible?

aarondl:
ENV vars are supported. Although it is poorly documented. `PSQL_*` works, so to remove the password from the configuration file and use env vars you would do: `PSQL_PASS=password`

ORESoftware:
can we add docs for those relevant env vars and how to capture them in the .toml file?

aarondl:
I rewrote the section on configuration in the README. It shows how to use the config vars (though does not provide an exhaustive list).

PR #802, opened 2020-07-16
Prevent blacklist and whitelist set via ENVs from being converted int…

Amandeepsinghghai:
Hi,
First of all thanks for the project. Really appreciate it.

As for the PR, I came across this issue when I tried to generate the models inside a container. I replicated the issue by setting the values in `~/.zshenv`(for bash it would be `~/.bash_profile`)  -> `export PSQL_BLACKLIST=schema_migrations` and ran sqlboiler which generated the `schema_migrations` model.

My current workaround is having the sqlboiler.toml file as 
```
[psql]
  blacklist = ["schema_migrations"]
```
and **NOT** setting the PSQL_BLACKLIST env value

**NOTE**: If I set both the env and the sqlboiler.toml file values, the envs take precedence and the `schema_migrations` model gets generated

Let me know how I can help to get this fix onto master.

Might help to get all envs working with this issue: https://github.com/volatiletech/sqlboiler/issues/678

### DIFF
--- a/main.go
+++ b/main.go
@@ -204,8 +204,10 @@ func preRun(cmd *cobra.Command, args []string) error {
 
 	keys := allKeys(driverName)
 	for _, key := range keys {
-		prefixedKey := fmt.Sprintf("%s.%s", driverName, key)
-		cmdConfig.DriverConfig[key] = viper.Get(prefixedKey)
+		if key != "blacklist" && key != "whitelist" {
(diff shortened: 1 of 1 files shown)


aarondl:
Why in the world do I have the feeling I've seen this exact issue and that it had been corrected at some point? Either way thank you for the PR :)

Issue #565, opened 2019-07-02
[Feature] Table and Column Comments


mickeyreiss:
It would be helpful to pull comments from tables and column, and if they are present, write them as comments in the generated go code. This will help with documentation and follow the design goal: the schema is the source of truth.

If this is interesting, I can help with the mysql implementation.

aarondl:
Would appreciate this feature very much.

sjansen:
Assuming the above is acceptable, I also have plans for how to add support for PostgreSQL table comments.

vladvelici:
Is anyone still working on this? I'd be happy to send an updated patch in the next few days.

aarondl:
@vladvelici I don't believe anyone has done anything with it that I'm aware of.

vladvelici:
I've sent a PR. There are some tests failing, seems like for mysql. I can look at that if you think I'm on the right path for postgres.

https://github.com/volatiletech/sqlboiler/pull/807

arp242:
I did a bit of work on adding table/struct comments too, the problem I ran in to is that the `drivers.Constructor` interface defines a `TableNames` method:

    TableNames(schema string, whitelist, blacklist []string) ([]string, error)

Modifying this to also return the columns isn't possible with this function signature.

Are these interfaces considered stable? I can't add a new method to this interface without breaking users of this interface.

Otherwise I can add an optional `TableInfo` or `TableCapabilities` interface, which is a little bit uglier but maintains compatibility.

This also applies to `ViewConstructor.ViewNames`.

arp242:
Here's a basic patch btw: https://github.com/schoolyear/sqlboiler/commit/fea0e6b76ab8b5bbc7601b92cf7c69c2d0a8c5d5

It's compatible, but has some duplication in `TableNames()` and `Tables()`, but it's not too bad and a bit hard to avoid.

I'll wait for some feedback on what patch you want to take, but as soon as that's clear I'll add a test, maybe cleanup a few things, and submit a PR.

stephenafamo:
This looks alright. Would you be able to extend this to also include column comments?

PR #807, opened 2020-07-26
Add comments to structs for postgresql (#565)

vladvelici:
I used work from @sjansen's [pull request](https://github.com/volatiletech/sqlboiler/pull/622) and tried to respect @aarondl's comments on that PR as well.

I tried to add mysql here as well but after spending about one hour I couldn't get a local database working and I quit on it. Is it acceptable to add this for postgres only at first, maybe someone else can have a go with mysql? It should be an easy addition.

The comments, if available in the db, are added above the field in the struct. Multi-line comments are supported by prefixing each line with "// ". Splitting is done by a new template function `splitLines` that simply splits the string by `"\n"` and returns `nil` for empty strings (otherwise empty comment lines were added in case of no comments as the split function would return a slice with one element, the empty string).


aarondl:
Good enough. Rather have it for postgres than nothing at all. Hopefully someone will add it for the others one day. Thanks for the PR. Will have to figure out why the tests are failing before this gets released.

PR #808, opened 2020-07-27
add uniq_keys info

wych42:
In my scenario of using unique key, it's not only for constraints but also for query. My customized templates would generate functions named `Find${TableName}By${UniqKeys}` for each uniq key, but uniq key info is not retrieved in `drivers/table.go`, so I created this pull request.

The usage would be like this:
```
{{ range $uniqKey := .Table.UniqKeys -}}
{{- $ukColDefs := sqlColDefinitions $.Table.Columns $uniqKey.Columns -}}
{{- $ukColNames := $ukColDefs.Names | stringMap (aliasCols $alias) | stringMap $.StringFuncs.camelCase | stringMap $.StringFuncs.replaceReserved -}}
{{- $ukArgs := joinSlices " " $ukColNames $ukColDefs.Types | join ", " -}}
{{- $fnName := $ukColNames | stringMap $.StringFuncs.titleCase | join "" -}}

// Find{{$alias.UpSingular}}By{{$fnName}} retrieves a single record by uniq key {{$ukColNames}} with an executor.
func Find{{$alias.UpSingular}}By{{$fnName}}({{if $.NoContext}}exec boil.Executor{{else}}ctx context.Context, exec boil.ContextExecutor{{end}}, {{$ukArgs}}, selectCols ...string) (*{{$alias.UpSingular}}, error) {
	{{$alias.DownSingular}}Obj := &{{$alias.UpSingular}}{}

	sel := "*"
	if len(selectCols) > 0 {
		sel = strings.Join(strmangle.IdentQuoteSlice(dialect.LQ, dialect.RQ, selectCols), ",")
	}
	query := fmt.Sprintf(
		"select %s from {{$.Table.Name | $.SchemaTable}} where {{if $.Dialect.UseIndexPlaceholders}}{{whereClause $.LQ $.RQ 1 $uniqKey.Columns}}{{else}}{{whereClause $.LQ $.RQ 0 $uniqKey.Columns}}{{end}}{{if and $.AddSoftDeletes $canSoftDelete}} and {{"deleted_at" | $.Quotes}} is null{{end}}", sel,
	)

	q := queries.Raw(query, {{$ukColNames | join ", "}})

	err := q.Bind({{if not $.NoContext}}ctx{{else}}nil{{end}}, exec, {{$alias.DownSingular}}Obj)
	if err != nil {
		if errors.Cause(err) == sql.ErrNoRows {
			return nil, sql.ErrNoRows
		}
		return nil, errors.Wrap(err, "{{$.PkgName}}: unable to select from {{$.Table.Name}}")
	}

	return {{$alias.DownSingular}}Obj, nil
}
{{ end -}}
```


aarondl:
I'm concerned about adding something like this to sqlboiler only for third-party template flexibility. This represents additional calls to the drivers as well as other drivers needing to adopt this: https://github.com/volatiletech/sqlboiler-sqlite3 https://github.com/glerchundi/sqlboiler-crdb

wych42:
>  only for third-party template flexibility. 

It's something missing from the beginning, since unique keys are one of the essential parts of table schema and should be included.

> This represents additional calls to the drivers as well as other drivers needing to adopt this

This may be a break change for third-party drivers, but they can depends on the current release(without this change) or adopt this change. I would like to help when avaliable.

PR #811, opened 2020-07-28
mysql driver support comment tinyint_as_bool

uffy:
It is very expensive to change the column type to fit tinyint_as_int feature if a database is not designed for it when it created.

Is use comment of a column a better way to implement this feature?

```SQL
CREATE TABLE `test` (
  `i_am_tinyint` tinyint(3) UNSIGNED NOT NULL DEFAULT '0',
  `i_am_bool` tinyint(3) UNSIGNED NOT NULL DEFAULT '0' COMMENT 'tinyint_as_bool',
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci;
```

### DIFF
--- a/drivers/column.go
+++ b/drivers/column.go
@@ -16,6 +16,7 @@ type Column struct {
 	Nullable  bool   `json:"nullable" toml:"nullable"`
 	Unique    bool   `json:"unique" toml:"unique"`
 	Validated bool   `json:"validated" toml:"validated"`
+	Comment   string `json:"comment" toml:"comment"`
 
 	// Postgres only extension bits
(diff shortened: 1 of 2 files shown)


wych42:
If it's expensive to change column type from tinyint(3) to tinyint(1), won't it be expensive to add comment too?

aarondl:
I'm not interested in this kind of pull request. sqlboiler's meant to be a fairly opinionated piece of software (either your DB works well with it or it doesn't). We've done enough to support mysql bools and I don't think it's a good idea to go any further.

PR #820, opened 2020-08-10
Use schema table instead of table name in mysql and mssql upsert

emmanuel099:
Using the table name instead of the schema table causes some problems when, e.g., the table name is a SQL keyword.


aarondl:
Seems reasonable. Thank you for this :)

PR #822, opened 2020-08-11
TableNames now honors config alias overrides

### LINKED ISSUE #813: models.TableNames does not match alias definitions

SQLBoiler v4.2.0

Postgresql 12

sqlboiler psql (I have the following in my sqlboiler.toml file)
```toml
[aliases.tables.ab_envs]
up_plural = "ABEnvs"
up_singular = "ABEnv"
down_plural = "abEnvs"
down_singular = "abEnv"
```

```sql
CREATE TABLE IF NOT EXISTS ab_envs (id serial PRIMARY KEY, name varchar(25) NOT NULL UNIQUE);
```
I ran sqlboiler, no errors, and everything except the models.TablesNames use the alias as defined.
models.TableNames contains AbEnvs instead of ABEnvs.  I would expect model.TableNames to be consistent with the alias definitions.

mgrubb:
Fixes #813

### DIFF
--- a/templates/singleton/boil_table_names.go.tpl
+++ b/templates/singleton/boil_table_names.go.tpl
@@ -1,9 +1,9 @@
 var TableNames = struct {
-	{{range $table := .Tables -}}
-	{{titleCase $table.Name}} string
+	{{range $tblAlias := .Aliases.Tables -}}
+	{{$tblAlias.UpPlural}} string
 	{{end -}}
 }{
-	{{range $table := .Tables -}}
-	{{titleCase $table.Name}}: "{{$table.Name}}",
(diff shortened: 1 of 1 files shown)


aarondl:
You know I hadn't thought about it until just now. But this is a breaking change :( Meaning we can't add it since suddenly people's generated code will fail to compile if they just regenerate. Backwards compatibility is tough.

mgrubb:
Would a command flag to switch this change on and keep the default behavior as it is be adequate to address backward compatibility?

aarondl:
I'd rather just leave it until the next major version.

aarondl:
The next major version has no release date and isn't really being "planned". We could consider flagging this if it's that important for users.

stephenafamo:

PR #831, opened 2020-08-28
mssql, mysql: Add support for "do nothing" on conflict to Upsert

emmanuel099:
I have added the `None` column kind to avoid adding a new `updateOnConflict` boolean parameter to `Upsert`, which would cause an API break.

Please note that `None` is not absolutely necessary. An empty update list can also be build by passing an empty white list instead of `None`.
But having an dedicated column kind to intentionally set the update columns to empty has the (big) advantage, that
we still get an error if the list of update columns couldn't be build when `Infer`, `Whitelist`, `Greylist` or `Blacklist` is used.

Example: `material.Upsert(ctx, tx, boil.None(), boil.Infer());`
Corresponds to `material.Upsert(ctx, tx, false, nil, boil.Infer(), boil.Infer());` when using Postgres.


aarondl:
This seems fine. Thanks for the PR. Sorry it took so long, exceedingly busy lately.

Issue #424, opened 2018-11-12
Feature Request: make enums their own type


jstrohm-triller:
I'm new to Go and SQLBoiler but would it be possible to have enums be their own type like this?

```
type Workday 
const (
  WorkdayMonday Workday  = "monday"
  WorkdayTuesday Workday  = "tuesday"
  WorkdayWednesday Workday = "wednesday"
  WorkdayThursday Workday = "thursday"
  WorkdayFriday Workday  = "friday"
)
```

aarondl:
This is definitely something that could be implemented but it requires a larger change to be useful.

In SQLBoiler there's still a few places where it's useful to build strings with `+` and this prevents that from happening because it requires a conversion from `Workday` to `string` for it to work.

aarondl:
This could be done as part of #426 but it's a breaking change.

aarondl:
This could possibly be done now as #426 is finished. Though it is a breaking change.

Atrox:
I would love to see this in SQLBoiler. Maybe as an option in `sqlboiler.toml`?

aarondl:
Since it is a breaking change yes, an opt-in flag in config would do until 4.0 at which point that'd be the only way. This really is only made possible by 426 though since those types will be pain to deal with in any other scenario.

aarondl:
Happy to see a PR for this at some point.

tomscholz:
I'd love to help. As far as I understood is #426 resolved and the feature should be behind a feature flag. Correct?

tomscholz:
Okay, so I'm currently browsing the repo and making myself familiar with how the project is structured. The template for the boil_types.go` is located under `templates/singleton/boil_types.go`.

There is `templatebin/bindata.go` which is the only place I could find where the file name is being used, but I'm unsure what `bindata.go` exactly does. Any guidance here would be helpful :)

Cheers,
Tom

aarondl:
@tomscholz take a look at the contributing guide. Talks about a tool called go-bindata and where to get it. The `boil.sh` build script has ways to take the modified templates and turn it into that bindata.go

PR #834, opened 2020-09-03
feat: make enums their own type

tomscholz:
As discussed in #424.

I'm currently in the cleanup phase and will commit the results at the end of this week.


tomscholz:
Ahh tests! I'll look into writing some over the weekend.

aarondl:
Was waiting on the tests tom if you were going to write any.

tomscholz:
Hi Aaron,
sorry I've been so quiet. I've gone ahead and added two methods String() and IsValid() for each generated enum type.

I've also tried refactoring some of the stuff in `boil_types.go.tpl`. I haven't really worked with the go templating engine before, so I'm not exactly sure how bad the refactoring is. I'm not really happy with it, but I thought it made sense to put some of the logic that is repeated into variables. Maybe you could have a look.

As for the tests; I'm planning on writing simple tests for IsValid and String or do you think that is redundant and some aspect should be tested?

Have a nice day,
Tom :)

templates/singleton/boil_types.go.tpl
```diff
@@ -60,29 +60,81 @@ MySQL output looks like:    TableNameColNameEnumValue = "enumvalue"
...
 It only titlecases the EnumValue portion if it's snake-cased.
 */}}
+
 {{$once := onceNew}}
 {{- range $table := .Tables -}}
 	{{- range $col := $table.Columns | filterColumnsByEnum -}}
 		{{- $name := parseEnumName $col.DBType -}}
 		{{- $vals := parseEnumVals $col.DBType -}}
 		{{- $isNamed := ne (len $name) 0}}
+        {{- $enumName := "" -}}
```
aarondl, line 70: I don't really understand why this is declared here and used way further down in a completely different indentation level. Why not declare closer to where it's used?

templates/singleton/boil_types.go.tpl
```diff
@@ -60,29 +60,81 @@ MySQL output looks like:    TableNameColNameEnumValue = "enumvalue"
 
 It only titlecases the EnumValue portion if it's snake-cased.
 */}}
+
```
aarondl, line 63: The whitespace changes in these templates are a bit odd. I notice you're using spaces but these are all tabs (like Go is).

templates/singleton/boil_types.go.tpl
```diff
@@ -60,29 +60,81 @@ MySQL output looks like:    TableNameColNameEnumValue = "enumvalue"
...
+    {{ $enumName = titleCase $table.Name $col.Name}}
+{{- end -}}
+
+{{if $.AddEnumTypes}}
+type {{$enumName}} string
+{{end}}
+
+// Enum values for {{$enumName}}
 const (
-	{{- range $val := $vals -}}
```
aarondl, line 77: By unindenting this you're changing the scheme of how these templates are done. This belongs to the `if` statement above on line 76 and so it should remain indented.

The reason for this is because the templates are REAL hairy (as you probably noticed haha) and so keeping the actual template lines indented to match the nesting level helps to keep them a bit cleaner.

aarondl (changes requested):
This PR seems mostly fine. Just need to tidy up the indentation in the templates. Keep it consistent (tabs not spaces at the front, just like Go code) and respect the nesting level where possible (sometimes due to the thing you're templating it might not always be possible).

stephenafamo:
Is this still on the cards @tomscholz?
If you're busy at this time, maybe I can pull from your branch and make the required changes?

PR #837, opened 2020-09-12
Fix aliased timestamp column name in struct

### LINKED ISSUE #788: Generated code doesn't match generated table model when using aliases for automatic timestamp fields

v3.7.1

Postgresql 12.2

```
CREATE TABLE IF NOT EXISTS "user_management"."user"
(
    "id"                uuid                        PRIMARY KEY,
    "created_by"        uuid                        NOT NULL,
    "created_at"        timestamp with time zone    NOT NULL,
    "updated_by"        uuid                        NOT NULL,
    "updated_at"        timestamp with time zone    NOT NULL,
    "deleted"           bool                        NOT NULL,
    "deleted_by"        uuid                        NOT NULL,
    "deleted_at"        timestamp with time zone    NOT NULL,
    "first_name"        varchar(100)                NOT NULL,
    "last_name"         varchar(100)                NOT NULL,
);
```

In the sqlboiler.toml, I set the aliases for the automatic time fields as such:
```
[aliases.tables.user.columns]
    created_at = "CreateTime"
    updated_at = "UpdateTime"
    deleted_at = "DeleteTime"
```

and after code generation the table model looks correct:
```	
type Customer struct {
	...
        CreateTime   time.Time  `boil:"created_at" json:"created_at" toml:"created_at" yaml:"created_at"` 
	...
}
```
However in all the database interaction functions, the old time field names are used
```
func (o *User) Insert(ctx context.Context, exec boil.ContextExecutor, columns boil.Columns) error {
	if o == nil {
		return errors.New("models: no user provided for insertion")
	}

	var err error
	if !boil.TimestampsAreSkipped(ctx) {
		currTime := time.Now().In(boil.GetLocation())

		if o.CreatedAt.IsZero() {                          <<<
			o.CreatedAt = currTime                     <<<
		}
		if o.UpdatedAt.IsZero() {                          <<<
			o.UpdatedAt = currTime                     <<<
		}
	}
        ...

```

the code obviously doesn't compile

Is there a solution to this while keeping the aliases?

I use reflection to automatically map my input structs to the table model generated by sqlboiler and would like to stick to google standardised names for fields like created_time, updated_time, ...

while-loop:
closes #788 
Adds aliased autoupdate timestamp fields to the insert/update/upsert function for models.

I downloaded go-bindata v3.21.0 (as specified in [go.mod](https://github.com/volatiletech/sqlboiler/blob/dev/go.mod#L13)) and it seems like they've added new code for `AssetDebug`.
I'm not sure if you want those out of the generated files or keep it as-is.


aarondl:
Thanks for this!

PR #842, opened 2020-09-24
Add before-select hook

### LINKED ISSUE #525: BeforeSelectHook

Hades32:
This should fix #525 .

Only special thing to note: I had to introduce dummy objects for the selects that return more than one object, as there was non there yet. Also this means an asymmetry: before-select is called only once, but after select is called N-times. IMHO this is not correct and a better solution would be to have after-select receive a slice, but I didn't want to touch this


aarondl:
Hm. Seeing the implementation it's sort of not ideal. The hook is sort of an object-level "before or after X happens, this object has Y happen". This is more a "before you do any selects at all for an object of this type, do this without regard to any object in general".

I'm questioning the utility of such a method. Other than tracing, or debug printing what could we possibly use this for?

stephenafamo:
Closing this due to inactivity and merge conflicts.

If you're still up for it @Hades32 you can create a new PR. Although we're still not sure of the utility of such a hook.

parkerduckworth:
> Other than tracing, or debug printing what could we possibly use this for?

@aarondl a good usecase such a hook is for making authorization checks before querying the DB, which is what lead me to this issue/PR. It would be awesome to have `BeforeSelectHook` for this reason alone.  But let's say this hook _did_ exist just for tracing or debugging -- what is the downside to introducing it?

parkerduckworth:
> If you're still up for it @Hades32 you can create a new PR. Although we're still not sure of the utility of such a hook.

I would take care of this if @Hades32 is no longer interest

stephenafamo:
I'm still unsure about this feature, but I am willing to review a PR for this.

Issue #809, opened 2020-07-27
Order of columns matter in eager loading, it causes problems sometimes


vladvelici:
Hello.

This is somewhat related to #587 but it's not about selecting custom columns. It's about making the order of the columns in the database irrelevant for the queries by getting rid of `select *`.

Use case: It's not important most of time time, but we encounter this problem in our development when we use migrations. With postgres you can't specify the position of columns. If an up migration removes a column and a down migration re-adds it, the order of the columns is not preserved after a migrate up followed by migrate down operation (which happens often in development and reviewing pull requests).

A fix would be to generate queries that specifically select columns by name as opposed to `select *` in eager loading methods - this way even if the columns in the database are different and models weren't regenerated, the application won't crash. Is this a complicated change? Would this be fixed anyway by the implementation of #587?

Using sqlboiler v4.2.0 with postgres 9.6 if it's relevant.

aarondl:
Hi @vladvelici. I think this is a fine change to go ahead and implement. In reality `select *` probably should never have been used. And we even did workarounds for it to exist when say adding a column the sqlboiler generated code didn't know about yet via a different application (made `Bind` work so that it ignores columns it doesn't find, rather than exploding as is proper).

I think that if we were to go back to this, bind should start failing if it finds column results for which it has no place to bind to again.

I don't believe it's a very complicated change.

vladvelici:
It's great this has been fixed, I've noticed it in some of my generated code. Just wanted to say thank you @inoc603 !

I hope it's OK that I'm closing this issue now, seems fixed as far as I can tell.

PR #845, opened 2020-10-12
refactor: use column names instead of * for eager loading in m2m relationships

inoc603:
As stated in #809 , the order of columns may cause errors during eager loading. This commits only replaces `select *` in m2m relationships eager loading with actual column names, since this is the only place eager loaded data is scanned manually with generated code.

I'm using the latest version of go-bindata, so the generated code in drivers also has some changes.


aarondl:
Thanks for this!

PR #846, opened 2020-10-16
Add support to uniqueidentifier and datetimeoffset2 mssql column types

severedsea:
- Add [datetimeoffset](https://docs.microsoft.com/en-us/sql/t-sql/data-types/datetimeoffset-transact-sql?view=sql-server-ver15) column type translation to time.Time
- Fix [uniqueidentifier](https://docs.microsoft.com/en-us/sql/t-sql/data-types/uniqueidentifier-transact-sql?view=sql-server-ver15) handling by changing the column type translation to [mssql.UniqueIdentifier](https://github.com/denisenkom/go-mssqldb/blob/master/uniqueidentifier.go).


aarondl:
Unfortunately this doesn't seem okay. It introduces a new dependency and a type not every user may want.

Is there good reason for using `mssql.UniqueIdentifier`? Is this something we need? Are there problems if we don't use it?

severedsea:
We're using MS SQL uniqueidentifier type for our PK, and unfortunately, using string data type doesn't represent the data correctly. 

For instance, if I retrieve a parent record's ID and try to use it as an FK when inserting a child record (one-to-many), MS SQL will throw an error because of the FK value. 

I will run some samples later to explain the issue better.

aarondl:
My question is more: Is mssql usable without this? Is this just what your team is using and their preference?

We have support for type replaces, without this PR you can start using this type. See the README for details on that.

severedsea:
Thanks! I'll take a look at the type replaces 🙏🏻 

But currently, MS SQL's uniqueidentifier column type is unusable with string as backing data type (without unmarshalling into mssql.UniqueIdentifier)

lohmanndouglas:
image

I am trying to do an insert on a time that has a datetimeoffset field but I could not perform the operation. Then I found this PR.   

Does it mean that boiler does not handle "datetimeoffset"? How do I handle datetimeoffset using sql boiler?

severedsea:
> image
> 
> I am trying to do an insert on a time that has a datetimeoffset field but I could not perform the operation. Then I found this PR.
> 
> Does it mean that boiler does not handle "datetimeoffset"? How do I handle datetimeoffset using sql boiler?

The current state does not support/handle `datetimeoffset`, hence I added in the following swich cases in this PR: 
- https://github.com/volatiletech/sqlboiler/pull/846/commits/8096c7ec55a2442504a2b9c66f7b226d02bb7c05#diff-ef59d41f66e467478cd71c3916dba52511ef7d6cac3bdf1dede91eb75df47051R432
- https://github.com/volatiletech/sqlboiler/pull/846/commits/8096c7ec55a2442504a2b9c66f7b226d02bb7c05#diff-ef59d41f66e467478cd71c3916dba52511ef7d6cac3bdf1dede91eb75df47051R398

aarondl:
Well, as you mentioned string cannot be used at all for properly interacting with this type. It's likely that current users are either not using this type or have type replaced it away so it should be safe to merge this. It doesn't -really- add another dependency unless you use this type in your models so I think this is okay. Thanks for the PR.

lohmanndouglas:
@severedsea Could you help me figure out how make this datetimeoffset work with sqlBoiler, please? 

I have a table with a field (DatAlt) type of datetimeoffset. So I did the following:
  - create a config file . toml (attached an example)
  - generate the model using the commnad "**sqlboiler mssql -c configs/boil/*.toml --wipe**"
  - try to insert a row "batch.Insert(context.Background(), h.DB, boil.Infer())". batch is a generate model.
  
 So far a got the error:

image

Checking this I found the [datetimeoffset conversion handling](https://github.com/denisenkom/go-mssqldb/blob/master/doc/how-to-handle-date-and-time-types.md). Ok, but then would need a specific query to handle the insert and I won't use boil insert? 
I was expecting that with this PR I could handle this, but I am kinda lost in how to make it work. 

I believe that I am missing something in the usage of SqlBoiler and mssql. Can you help me?





SqlBoiler version: **4.5.0**

Table filed type:

image



Model generate file: 
```go
// Code generated by SQLBoiler 4.5.0 (https://github.com/volatiletech/sqlboiler). DO NOT EDIT.
// This file is meant to be re-generated in place and/or deleted at any time.

...

type Batch struct {
        ...
	DatAlt       null.String `boil:"dat_alt" json:"dat_alt,omitempty" toml:"dat_alt" yaml:"dat_alt,omitempty"`
        ...
	R *batchR `boil:"-" json:"-" toml:"-" yaml:"-"`
	L batchL  `boil:"-" json:"-" toml:"-" yaml:"-"`
}

...

```

config file example:

```
  output   = "models/hierarchy"
  wipe     = true
  no-tests = true
  
  [mssql]
    dbname  = "dbname"
    host    = "hosturi"
    port    = 1454
    user    = "username"
    pass    = "password"
    schema  = "schemaname"
```


I will appreciate a lot any help that you can provide

severedsea:
@lohmanndouglas Hmm... Even with the `DatAlt` having a `time.Time` type, it still throws that error? Not sure if there are updates to `go-mssqldb` driver since the last time I used. 

Sorry, couldn't help much. But back when I was working on this, that error usually comes up if the type is wrong (similar to using a string to represent MS SQL's uniqueidentifier)

PR #848, opened 2020-10-19
Fix NotIn query mods and building.

peterldowns:
This PR fixes two problems:
* The `qm.WhereNotIn`, `qm.AndNotIn`, `qm.OrNotIn` helpers were adding `Query` structs with the wrong `kind` because they used the `AppendIn` function instead of the `AppendNotIn`. This presents as a bug where building a not in query with an empty set -- it should be rendered as `(1=1)` (everything is not a member of the empty set) but instead renders as `(1=0)`, as if it were a In query (nothing is a member of the empty set.)
* Queries with `whereKindNotIn` were not rendering appropriately because the regex used to parse their `clause` was the same as the one used for `whereKindIn`. I created a separate regex instead of making the `NOT\s+` optional.

This PR includes tests for the second problem, but none for the first because there are currently no tests for the `qm` package and I wasn't sure the best way to write them. I think the best testing strategy might be to update the query building tests to call the `qm` helpers instead of manually constructing `Query` structs, or at least add some tests that work this way. But that led me to some circular import errors so I figured I'd send the PR without the tests as a starting point.

### DIFF
--- a/queries/qm/query_mods.go
+++ b/queries/qm/query_mods.go
@@ -361,7 +361,7 @@ type whereNotInQueryMod struct {
 
 // Apply implements QueryMod.Apply.
 func (qm whereNotInQueryMod) Apply(q *queries.Query) {
-	queries.AppendIn(q, qm.clause, qm.args...)
+	queries.AppendNotIn(q, qm.clause, qm.args...)
 }
 
 // WhereNotIn allows you to specify a "x NOT IN (set)" clause for your where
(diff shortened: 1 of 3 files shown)


aarondl:
Wow. This is awkward that this problem existed. Thank you very much for the PR.

peterldowns:
@aarondl do you have any thoughts on the value of updating these tests to use the `qm` helpers for constructing the `Query` structs rather than instantiating them directly?

aarondl:
I think probably the only reason that it's the way that it is is because queries and qm like to dance around circular dependency issues. I'm not sure if it's possible to convert all existing tests to use only public interfaces from both packages in order to move the test packages out and avoid the circular dep issue.

There's value for sure. Just don't know the costs to weigh it properly.

peterldowns:
Makes sense. I'll noodle on it -- maybe there's a way to do some import magic here to make it easier. Thanks again.

PR #857, opened 2020-11-05
Fix typo in "Upsert"

abeltay:
Ref: #831


aarondl:
Thanks for pointing this out! Will make the change manually as this PR is against master and not dev.

Issue #861, opened 2020-11-13
Encountered issues when executing sqlboiler with uppercased tables in psql


severedsea:
If you're having a generation problem please answer these questions before submitting your issue. Thanks!

### What version of SQLBoiler are you using (`sqlboiler --version`)?
v4.3.0

### What is your database and version (eg. Postgresql 10)
Postgresql 11

### If this happened at generation time what was the full SQLBoiler command you used to generate your models? (if not applicable leave blank)
```
sqlboiler psql
```

### What is the output of the command above with the `-d` flag added to it? (Provided you are comfortable sharing this, it contains a blueprint of your schema)
```
using driver: /Users/xxx/go/bin/sqlboiler-psql
unable to fetch table column info (UPPERCASE_TABLE): pq: relation "public.uppercase_table" does not exist
[...]

PR #862, opened 2020-11-13
Case-sensitive table name on psql regclass table lookup

severedsea:
Fixes: https://github.com/volatiletech/sqlboiler/issues/861

Ref: https://stackoverflow.com/questions/56227815/is-there-a-way-to-use-regclass-in-a-case-sensitive-way

### DIFF
--- a/drivers/sqlboiler-psql/driver/psql.go
+++ b/drivers/sqlboiler-psql/driver/psql.go
@@ -207,7 +207,7 @@ func (p *PostgresDriver) Columns(schema, tableName string, whitelist, blacklist 
 		c.domain_name,
 		c.column_default,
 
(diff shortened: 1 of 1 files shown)


aarondl:
Looking at the linked stackoverflow wouldn't we need to add quotes around the schema name as well? I'm not 100% sure, just asking the question.

```sql
-- Recommended by source:
"schema"."tablename"
-- What this PR does
schema."tablename"
```

Also this PR should have been opened against the dev branch as per CONTRIBUTING.md :) Thanks for getting started on this issue.

severedsea:
Oops, sorry about that! Will make the changes 🙇

aarondl:
Thanks for the quick turnaround. Merged and released :)

PR #866, opened 2020-11-21
fix compatibility with ANSI_QUOTES mode

alexsander-souza:
if ANSI_QUOTES is enabled, double quotes cannot be used as string delimiter.

This fixes a regression introduced in #733


alexsander-souza:
Fixes #868

aarondl:
From the documentation (mostly for my own reference):

> ANSI_QUOTES
>
> Treat " as an identifier quote character (like the ` quote character) and not as a string quote character. You can still use ` to quote identifiers with this mode enabled. With ANSI_QUOTES enabled, you cannot use double quotation marks to quote literal strings because they are interpreted as identifiers.

Also: https://dev.mysql.com/doc/refman/8.0/en/string-literals.html

It seems like single quotes were the way to go for the highest degree of compatibility anyway. Will merge this soon. Thank you for the PR.

PR #872, opened 2020-12-06
support for qm.Comment query mod

Pilatuz:
This simple query mod adds custom comment just before main query.

For example:

```{.go}
q := shop.Customers(
  qm.Comment("shop.getAllCustomers"),
  qm.Select("id", "name"),
  qm.From("customers"),
)
```

will be:

```{.sql}
-- shop.getAllCustomers
SELECT id, name FROM customers
```

these comments might be very useful for query analysis at database layer - we can easily find function/method related to particular query.

### DIFF
--- a/queries/qm/query_mods.go
+++ b/queries/qm/query_mods.go
@@ -554,6 +554,22 @@ func For(clause string) QueryMod {
 	}
 }
 
+type commentQueryMod struct {
+	comment string
+}
+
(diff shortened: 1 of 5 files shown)


aarondl:
I'm not opposed to this. Code seems reasonable, you've included tests, it's non-breaking and barely affects users who don't use it. Thank you for the great PR.

PR #874, opened 2020-12-07
exclude foreign keys on generated columns for psql

chochihim:
Generated columns are filtered out in [Columns](https://github.com/volatiletech/sqlboiler/blob/6ccc92cffb06c37bd14183ea1e103ff942aee22b/drivers/sqlboiler-psql/driver/psql.go#L267). If I set foreign keys on those generated columns, errors will be thrown in `setForeignKeyConstraints` (panic in `GetColumn`). So these columns needed to be filtered out in `ForeignKeyInfo` as well.

This meta info about whether a column is generated (pg_attribute.attgenerated) is only added after [Postgreql 12](https://www.postgresql.org/docs/12/catalog-pg-attribute.html), so I need to add a version check for the query.

### DIFF
--- a/drivers/sqlboiler-psql/driver/psql.go
+++ b/drivers/sqlboiler-psql/driver/psql.go
@@ -38,6 +38,7 @@ func Assemble(config drivers.Config) (dbinfo *drivers.DBInfo, err error) {
 type PostgresDriver struct {
 	connStr string
 	conn    *sql.DB
+	version int
 }
 
 // Templates that should be added/overridden
@@ -89,6 +90,11 @@ func (p *PostgresDriver) Assemble(config drivers.Config) (dbinfo *drivers.DBInfo
 			err = e
 		}
 	}()
+
+	p.version, err = p.getVersion()
+	if err != nil {
+		return nil, errors.Wrap(err, "sqlboiler-psql failed to get database version")
+	}
 
 	dbinfo = &drivers.DBInfo{
 		Schema: schema,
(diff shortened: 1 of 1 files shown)


drivers/sqlboiler-psql/driver/psql.go
```diff
@@ -90,6 +91,11 @@ func (p *PostgresDriver) Assemble(config drivers.Config) (dbinfo *drivers.DBInfo
 		}
 	}()
 
+	p.version, err = p.getVersion()
+	if err != nil {
+		return nil, errors.Wrap(err, "sqlboiler-psql failed to get database version")
```
aarondl, line 96: Will this succeed 100% on all postgresql versions? If not, it should not return an error and it should fall back to the old behavior.

aarondl (changes requested):
This should be targetting the dev branch as per our CONTRIBUTING.md

drivers/sqlboiler-psql/driver/psql.go
chochihim: I can find this option from [postgresql documentation since 8.2](https://www.postgresql.org/docs/8.2/runtime-config-preset.html). So technically not all versions will succeed. But all officially supported versions (>= 9.5) are supported. Do you think if it is fine to assume it just works?

If not, about the fallback, there can be different ways to achieve the old behaviour, e.g.
```go
// 1. introduce one more flag about if getVersion() succeeds:
// whenever we need to distinguish the version for a sql statement
if p.getVersionSucceeds && p.version >= 120000 {
  sql = "new behaviour"
} else {
  sql = "old behaviour"
}

// 2. just set p.version = 0 if getVersion() fails:
if p.version >= 120000 {
  sql = "new behaviour"
} else {
  sql = "old behaviour"
}
```
Which one do you prefer? Do you have any idea?

aarondl: As long as it works on officially supported its fine by me.

Issue #881, opened 2020-12-20
Incorrect enum type used when multiple schemas with the same enum name


csueiras:
There's a bug in generation for postgres regarding enums with the same name in different schemas, it appears like the first schema will be used for the enum values generated. I've isolated the bug to the query in here https://github.com/volatiletech/sqlboiler/blob/master/drivers/sqlboiler-psql/driver/psql.go#L191 specifically this sub-query:

```
select 'enum.' || c.udt_name || '(''' || string_agg(labels.label, ''',''') || ''')'
from (
	select pg_enum.enumlabel as label
	from pg_enum
	where pg_enum.enumtypid =
	(
		select typelem
		from pg_type
		where pg_type.typtype = 'b' and pg_type.typname = ('_' || c.udt_name)
		limit 1
	)
	order by pg_enum.enumsortorder
) as labels
```

The issue is that the sub-select from the `pg_type` table doesn't have any predicate for the schema, and thus the first enum returned by that query will be the one used. If I modify that query to the following, it now produces the right results:

```
select pg_enum.enumlabel as label
    from pg_enum
    where pg_enum.enumtypid =
          (
              select typelem
              from pg_type
              join pg_namespace ON pg_type.typnamespace = pg_namespace.oid
              where pg_type.typtype = 'b' and pg_type.typname = ('_' || c.udt_name) and pg_namespace.nspname=$1
              limit 1
          )
    order by pg_enum.enumsortorder
) as labels
```

### What version of SQLBoiler are you using (`sqlboiler --version`)?
v3.7.1

### What is your database and version (eg. Postgresql 10)
Postgres 10

### If this happened at generation time what was the full SQLBoiler command you used to generate your models? (if not applicable leave blank)
 sqlboiler -d -c ./sqlboiler.toml -o /tmp/models --wipe psql

### Please provide a relevant database schema so we can replicate your issue (Provided you are comfortable sharing this)
```
CREATE SCHEMA schema_a;
CREATE TYPE schema_a.my_enum AS ENUM ('VAL1','VAL2');
CREATE TABLE schema_a.my_table ( field schema_a.my_enum NOT NULL );

CREATE SCHEMA schema_b;
CREATE TYPE schema_b.my_enum AS ENUM ('OTHERVAL1','OTHERVAL2');
CREATE TABLE schema_b.my_table ( field schema_b.my_enum NOT NULL );
```

### Further information. What did you do, what did you expect?
Here's my queries and results with the fix I propose above:

```
[...]

PR #882, opened 2020-12-20
Schema predicate for enum generating query

csueiras:
Fix for #881 

There's no predicate for the pg_type sub-query to only look at the schema being used in generation, I am not exactly sure if this is 100% correct fix as you could want to use an enum from another schema but this is just to get the ball rolling and I am happy to provide test coverage for this change but I first wanted some early feedback on my suggested fix.


aarondl:
LGTM. Thanks for the PR. Though I should note in your issue you mention a pretty old version of sqlboiler and I'd encourage you to upgrade. This has landed in the dev branch which will end up being 4.4.1 and there will be no backport. Though perhaps the postgresql driver will continue functioning with no issues even against an older sqlboiler version.

PR #885, opened 2020-12-24
Respect column alias in Load{{$relAlias.Local}}

### LINKED ISSUE #883: Column alias produces models that do not compile

`v4.4.0`

`postgres 13.1`

`sqlboiler psql`

```
CREATE TABLE gpu (
    gpu_id serial PRIMARY KEY
);
CREATE TABLE gpu_thing (
    gpu_thing_id serial PRIMARY KEY,
    gpu_id integer REFERENCES gpu (gpu_id)
);
```

I added this alias for one of my columns in `sqlboiler.toml`:

```
[aliases.tables.gpu.columns]
gpu_id = "GPUID"
```

Model generation succeeds, but the result does not compile, because some of the generated code uses the alias and some doesn't:

```
models/gpu.go:409:29: object.GpuID undefined (type *Gpu has no field or method GpuID, but does have GPUID)
models/gpu.go:423:27: obj.GpuID undefined (type *Gpu has no field or method GpuID, but does have GPUID)
```

I believe I located the issue in [templates/09_relationship_to_many_eager.go.tpl](https://github.com/volatiletech/sqlboiler/blob/master/templates/09_relationship_to_many_eager.go.tpl). I can prepare a PR, but I'm not sure which version of `go-bindata` I need to match yours. I tried both the latest release and the one listed in your `go.mod` and they both create unrelated changes.

I have a minimal working example of the issue [here](https://github.com/jalan/sqlboiler_bug/), including the broken models.

jalan:
Fixes #883

Are there automated tests for this template? If so, I didn't find them yet and I'll probably find out when CI runs on this PR :smile:

I tested this change manually both with and without aliases, and it seems to work, at least for the database I tried....


jalan:
Found the relevant tests. I'll try to add one for this case.

jalan:
Added a test that compiles the templates with every table and every column aliased, which should prevent this issue from resurfacing. Note that the test doesn't include aliases for relationships, because I didn't quite understand how those work :sweat_smile:

aarondl:
Thanks for the PR!

PR #890, opened 2021-01-06
Avoid allocation in JSON.Scan with string

bouk:
`[]byte(src.(string))` is an unneeded allocation, because you can splat a string directly into an append call.

Benchmark:

```
$ go test -benchmem -bench=JSON

# Before

goos: darwin
goarch: arm64
pkg: github.com/volatiletech/sqlboiler/v4/types
BenchmarkJSON_Scan-8   	 4658832	       237.9 ns/op	    2304 B/op	       2 allocs/op
PASS
ok  	github.com/volatiletech/sqlboiler/v4/types	1.738s

# After

goos: darwin
goarch: arm64
pkg: github.com/volatiletech/sqlboiler/v4/types
BenchmarkJSON_Scan-8   	 8657862	       118.0 ns/op	    1152 B/op	       1 allocs/op
PASS
ok  	github.com/volatiletech/sqlboiler/v4/types	1.435s
```

### DIFF
--- a/types/json.go
+++ b/types/json.go
@@ -62,20 +62,16 @@ func (j JSON) Value() (driver.Value, error) {
 
 // Scan stores the src in *j.
 func (j *JSON) Scan(src interface{}) error {
-	var source []byte
-
-	switch src.(type) {
+	switch source := src.(type) {
 	case string:
-		source = []byte(src.(string))
+		*j = append((*j)[0:0], source...)
+		return nil
 	case []byte:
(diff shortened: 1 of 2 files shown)


aarondl:
Ah, great change! But the target branch should be dev as per CONTRIBUTING.md. Sorry to get to this so late :)

bouk:
fixed

aarondl:
Merged. Ty for PR.

PR #902, opened 2021-02-08
change readme to avoid eager loading for large query, refer to #510

comomac:
From production system, noticed and found some `qm.Load()` is executing large amount of queries and affecting performance, and had to rewrite some code to use RAW SQL JOIN queries.

Also checked on the Issues #510 and confirm it is not just isolated incidents.

Updated the README.md file to reflect actual situation and avoid confusions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1342,8 +1342,8 @@ languages, err := pilot.Languages().All(ctx, db)
 
 If your relationship involves a join table SQLBoiler will figure it out for you transparently.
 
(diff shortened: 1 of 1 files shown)


aarondl:
This is incorrect. Eager loading attempts to do one query per table.

If you have 3 users and you attempt to get their videos and the video's tags, you will see 3 queries, one requesting all users, one requesting all videos that have user id in (those 3 users), and one requesting all the tags that have video_id in (the videos returned from the previous query).

Issue #905, opened 2021-02-13
Relation methods and eager loading panics when using UNSIGNED BIGINT on MySQL


maku693:
In a specific situation, eager loading and relation methods panic. They are separate issues, but occurs when using the same schema, and related to each other, so I opened this as a single issue.

If we using MySQL and having schema like this:

```sql
CREATE TABLE `category` (
  `id` BIGINT UNSIGNED NOT NULL,
  PRIMARY KEY (`id`)
);

CREATE TABLE `article` (
  `id` BIGINT UNSIGNED NOT NULL,
  `category_id` BIGINT UNSIGNED DEFAULT NULL,
  PRIMARY KEY (`id`),
  FOREIGN KEY (`category_id`) REFERENCES `category` (`id`)
);
```

and following data:

```
mysql> select * from article;
+----+---------------------+
| id | category_id         |
+----+---------------------+
|  1 | 9223372036854775808 |
+----+---------------------+

mysql> select * from category;
+---------------------+
| id                  |
+---------------------+
| 9223372036854775808 |
+---------------------+
```

then eager loading panic:

```go
articlesWithCategory, _ := models.
  Articles(qm.Load(models.ArticleRels.Category)).
  All(ctx, db)
// panic: primitive type of a (string) was not the same primitive type as b (int64)
```

Also, relation method panic:

```go
article, _ := models.
  Articles(models.ArticleWhere.ID.EQ(1)).
  One(ctx, db)
_ := article.SetCategory(ctx, db, false, category)
// panic: tried to call Scan on *null.Uint64 with <nil> but got err: converting driver.Value type int64 ("-9223372036854775808") to a uint64: invalid syntax
```

This issue occurs only if

- The column with a foreign key (`article.category_id`) is `BIGINT UNSIGNED DEFAULT NULL`
- The referenced table's id (`category.id`) is `BIGINT UNSIGNED NOT NULL`
- We have large `uint64` for referenced table's id, which overflows the range of `int64`

## Details

For eager loading, the panic is from `queries.Equal()`.

https://github.com/volatiletech/sqlboiler/blob/b799bec0cfe58dfff8042220442d20afcba4dd9e/queries/reflect.go#L483

- Assume `queries.Equal()` called with `null.Uint64` (a) and `uint64` (b)
- a is `sql.Valuer`, so that `a.Value()` called, then a is converted to `string`
  - `null.Uint64.Value()` returns string when the value overflows range of int64
- b is converted to `int64`
[...]

PR #906, opened 2021-02-14
Fix panic message from queries.Assign()

maku693:
While digging into #905, I've found that the panic message is incorrect. If we in the case of `isDstScanner && !isSrcValuer`, `val` is always nil, so the message will be `tried to call Scan on *null.Uint64 with <nil> but got err: ...` for example.

I've fixed the message to contain `val`, like panic says `tried to call Scan on *null.Uint64 with -9223372036854775808 but got err: ...` for example.


aarondl:
Thanks for this quick fix :)

PR #907, opened 2021-02-14
Fix queries.Equal() not to panic when comparing null.Uint64 and uint64

maku693:
As proposed in #905, I've fixed `queries.Equal()` to format given number to string when one is a number and another is a string after pulled a primitive value from `driver.Valuer`.

This will fix part of #905. (Eager loading panics)

I believe this will not break backward compatibility, but I'm not sure if there are some corner cases.

This PR won't change the behavior of the function when passing pairs, except for a number, and a value which finally becomes a string. In most cases, comparison with string and number won't happen. 

For affected cases, if the given string cannot be parsed as a number, a panic will raise from `parseNumeric()`. But for this situation, panic will be raised in the current dev branch too, so from the point of library users, the behavior seems to be unchanged almost.

### DIFF
--- a/queries/reflect.go
+++ b/queries/reflect.go
@@ -7,6 +7,7 @@ import (
 	"database/sql/driver"
 	"fmt"
 	"reflect"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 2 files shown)


aarondl:
An acceptable workaround for the weirdness that is uint64 and `database/sql`. Thanks!

Issue #851, opened 2020-10-22
WhereIn helper not generated for nullable primitives


fdegiuli:
As the title says.

In sql, the `WHERE ... IN` clause is equally valid for nullable and non-nullable types.
The templates generate WhereIn helper methods for primitive types:
https://github.com/volatiletech/sqlboiler/blob/fad098aa2ba38220c0727620b7ea665983aec26e/templates/00_struct.go.tpl#L56
but since `isPrimitive()` (correctly) doesn't consider null types primitives, these helpers are not generated for nullable types (e.g. `null.String`)
https://github.com/volatiletech/sqlboiler/blob/a11ea0b016bdf0dda356324466de11eb31ee63e5/boilingcore/text_helpers.go#L128

This would have been a trivial change after https://github.com/volatiletech/sqlboiler/pull/575, but now it probably means adding a function to the text helpers to translate between null types and their primitive counterparts.

I'm willing to cut a PR if this is something you're interested in having.

aarondl:
Are you sure this is a problem? I'm almost positive I've used where helpers in the past with nullable primitives... You don't mention an sqlboiler version number so it's hard to tell.

fdegiuli:
Hmm I am using version 3.7.1, but from looking at the code I think it might still be an issue, unless I missed something.

I'll try to reproduce on the latest version.

bbarcklay-cpi:
I can confirm that IN/NIN whereHelpers are not created for nullable types in 4.4.

fdegiuli:
Fixed by https://github.com/volatiletech/sqlboiler/pull/1083

PR #910, opened 2021-02-16
Generates IN/NIN whereHelpers for nullable types

bbarcklay:
Enables IN/NIN whereHelper function generation for nullable types addressing issue #851. While this implements the pattern established by the other functions, I would find it more convenient to pass regular strings or ints for the IN/NIN tests.  Thoughts?


aarondl:
> While this implements the pattern established by the other functions, I would find it more convenient to pass regular strings or ints for the IN/NIN tests. Thoughts?

I think that this is the reason these don't exist in the first place. The "right" answer is a little bit hard to find. Either way you're going to end up with conversions at some point. And it's also not clear that the code here is what we want because of the trap of feeding a `null` to the `in` clause which is usually probably not what you wanted.

If you go the way of only being able to provide ints, then you limit the ability to create a simple list of these with something like:

```
models.Model(models.ModelWhere.Column.IN(otherModel1.X, otherModel2.X))
```

Of course that comes with a potential trap of generating `IN (5, null)` which you might want to restrict in some way because does anyone ever actually want that? What's the least surprising behavior here?

fdegiuli:
> If you go the way of only being able to provide ints, then you limit the ability to create a simple list of these with something like:
> 
> ```
> models.Model(models.ModelWhere.Column.IN(otherModel1.X, otherModel2.X))
> ```

My gut feeling is that it's probably okay for this to not be supported. Given that passing a NULL to a `WHERE IN` clause is largely meaningless in SQL, it seems reasonable not to accept nullable types in the helper to generate that clause.
You do miss out on some convenience, but anybody currently manually creating IN clauses wouldn't be any worse off, nor does this preclude us from adding such functionality at a later date.

> Of course that comes with a potential trap of generating `IN (5, null)` which you might want to restrict in some way because does anyone ever actually want that? What's the least surprising behavior here?

Not accepting null types would handle this. Also, `IN (5, NULL)` _is_ valid, if pointless, SQL; and I would be fairly surprised to see my ORM doing nil checks for me.

aarondl:
This will require some rebasing (bindata has been removed and there are changes in the same area of the code). I'm still largely undecided on it and lacking compelling arguments one way or the other I'm not sure what we should be doing here.

I think the biggest obstacle to overcome in saying we should just accept primitives like `Age.IN(5, 6, 7)` is that you have to do some additional mapping to figure out the null types non-null type?

Food for thought.

fdegiuli:
> I think the biggest obstacle to overcome in saying we should just accept primitives like Age.IN(5, 6, 7) is that you have to do some additional mapping to figure out the null types non-null type?

Yeah, I mentioned it above but I imagine something like this would be fairly safe and straightforward, given that primitive types aren't likely to change much:

stephenafamo:

PR #911, opened 2021-02-20
Fix change the number of columns from 255 to 65535.

kenko21:
In my project, I have a table with more than 255 columns.
I simply doubled the current 8-bit operation to 16 bits.
(255 -> 65535).

I know that 255 columns is too many, 
but I don't have the time to fix my legacy system at the moment.


aarondl:
I don't think that we want to adopt this change for a single user at the moment. Best to fork for now unless this sort of thing gathers more support.

kenko21:
Thank you for confirming this.
Maybe I should have made it an optional feature.

PR #914, opened 2021-02-21
feat(template) add column name variables with table name

sadayuki-matsuno:
I implemented variables of the column with the table name in the template.
This is a very useful variable when dealing with multiple tables such as inner join.


aarondl:
This is against the wrong branch. Please fix. Thanks for the PR. I was debating whether or not to accept this but in reality I've needed this myself more than a few times so decided to accept.

sadayuki-matsuno:
thanks you.
I changed the branch to merge.

aarondl:
Thanks again

PR #915, opened 2021-02-21
fix(psql) fix pq: got xxxx parameters but PostgreSQL only supports 65535 parameters

### LINKED ISSUE #850: Eager loading queries can exceed Postgres max number of parameters

SQLBoiler v4.2.0

Postgresql 10

```go
func main() {
	usages, err := dbmodels.IsrcUsages(
		dbmodels.IsrcUsageWhere.UsageReportID.EQ(30),
		qm.Load(
			dbmodels.IsrcUsageRels.Isrc,
		),
	).All(ctx, dbClient)
	if err != nil {
		panic(errors.Wrap(err, "get isrc usages failed"))
	}
	logger.Info(len(usages))
}
```

```sql
CREATE TABLE IF NOT EXISTS isrc (
    id serial PRIMARY KEY,
    isrc text UNIQUE NOT NULL
);

CREATE TABLE IF NOT EXISTS isrc_usage (
    id serial PRIMARY KEY,
    usage_report_id int NOT NULL REFERENCES usage_report (id) ON DELETE CASCADE,
    isrc_id int NOT NULL REFERENCES isrc (id) ON DELETE CASCADE,
    UNIQUE (usage_report_id, isrc_id)
);

```

In this example, if there are 10 ISRC usages for the given usage_report_id, then eager loading the ISRCs generates queries like:

```sql
SELECT * FROM "isrc_usage" WHERE ("isrc_usage"."usage_report_id" = $1);

SELECT * FROM "isrc" WHERE ("isrc"."id" IN ($1,$2,$3,$4,$5,$6,$7,$8,$9,$10));
```

The latter query will fail if the number of ISRCs exceeds a 16 bit integer (which pg uses to count query parameters), resulting in an error like:

```
failed to eager load Isrc: extended protocol limited to 65535 parameters
```

It would be fantastic if SQL boiler could intelligently adapt the generated query if the # parameters is too large.

sadayuki-matsuno:
What I want to fix in this PR is the following error. ( fix #850 )

```
 failed to assign all query results to xxx slice: failed to eager load xxx: failed to eager load xxx: pq: got 158369 parameters but PostgreSQL only supports 65535 parameters
```

This is an error that occurs when eager load is performed, because the table data is large and a large number of in clauses are specified for the table to be loaded.

This error is related to the amount of data and is very difficult to control in advance, which annoys me a lot.

I found [this article](https://stackoverflow.com/questions/52712022/what-is-the-most-performant-way-to-rewrite-a-large-in-clause) as a solution to this problem, and actually implemented it, and it seems to have solved the problem.

Specifically, the process of replacing the IN clause used in eager load with ANY.

However, since this is probably a postgres-specific problem and solution, we have separated the drivers for writing eager load templates and implemented them separately.

The reason I created the `psql_any.go.tpl` file is because the import in "github.com/lib/pq" is divided into necessary and unnecessary files, and I could not control the import pkg.


aarondl:
I think probably the main issue here is lack of `inner join` support inside the eager loading system. In these cases I think it's fine to resort to raw queries rather than add all this code to circumvent the problem. Keep in mind you can still struct bind to your models using proper calls to `queries.Bind`. You only need custom structs if you're going to be bringing back something other than a single model worth of data (like so: `select model.*`)

It's also true that this query will lead to performance problems and so a raw query is even more appropriate in this case.

I'm very sorry for not accepting many of your pull requests @sadayuki-matsuno. They are of high quality and are solving problems. I really appreciate your efforts. Having said that I am trying to be very strict regarding what comes in these days in terms of additional features (because it already has too many and is very complicated), as well as fixes to the eager loading system (because it needs a total rewrite).

PR #920, opened 2021-03-03
fix typo in README.md

gb9888:
there was a mistake in README.md that used pilot instead of jet.


aarondl:
Because this was opened against master not dev I've made the change myself to dev. Thanks for the PR and good catch!

PR #931, opened 2021-03-27
Add Expression query mod

varyoo:
Allow to query custom columns with arguments like this:

```Go
models.Articles(qm.Expression("MATCH (title) AGAINST (?) AS score", "+MySQL")).
  Bind(ctx, db, &fullTextSearchResults)
```

:blush:

### DIFF
--- a/queries/_fixtures/34.sql
+++ b/queries/_fixtures/34.sql
@@ -1,1 +1,1 @@
-SELECT *, MATCH (title) AGAINST (?) AS score FROM "articles";
+SELECT MATCH (title) AGAINST (?) AS score FROM "articles";
--- a/queries/_fixtures/34.sql
+++ b/queries/_fixtures/34.sql
@@ -1,0 +1,1 @@
+SELECT *, MATCH (title) AGAINST (?) AS score FROM "articles";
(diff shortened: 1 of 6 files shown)


aarondl:
Is this shared with any other database engine? Or is it just mysql?

varyoo:
`MATCH AGAINST` is purely MySQL, why?

vhakulinen:
Related: postgresql has syntax for doing a distinct select on a column (e.g. `SELECT DISTINCT ON (col1) col1, col2 ...`): https://www.postgresql.org/docs/12/sql-select.html#SQL-DISTINCT. If the placement of `Expression` query mod would be dependent on the order it was defined in the `NewQuery` args, this could also allow psql's `DISTINCT ON`.

E.g.
```go
models.Aricles(qm.Expression("DISTINCT ON (foo)"), qm.Select("foo", "bar"))
```
would become
```sql
SELECT DISTINCT ON (foo), foo, bar FROM articles```

aarondl:
@varyoo I typically like to refrain from adding highly specific query mods (especially database specific).

@vhakulinen the query mod ordering isn't tracked anywhere unfortunately :(

aarondl:
One thing to note is I'm very much considering opening up the query struct so that we can have custom query mods in the wild. I think that would be very nice.

PR #939, opened 2021-04-17
add uniq_keys info

wave2588:


aarondl:
Hi there Wave. I'm not sure what this PR is because you've left no description. It's also PR'd against the wrong branch. It's also got a decidedly odd branch name. Going to close for now.

If you would like to have this merged, please either reference a discussion we've had in the past or give some reasoning for why this change exists and is necessary.

PR #955, opened 2021-05-20
Add always-wrap-errors flag

jhnj:
Currently all methods on models except for `Model.One(...)` and `FindModel(...)` return wrapped errors with stacktraces. If no rows are returned `Model.One(...)` and `FindModel(...)` unwrap the returned error and return `sql.ErrNoRows` without  a stacktrace.

This PR adds an option, `always-wrap-errors` that disables this unwrapping. Comparing wrapped errors is supported in the standard library with `errors.Is(err, sql.ErrNoRows)`, and e.g. in our codebase we have a linter that doesn't allow direct equality checks when comparing errors (`err == sql.ErrNoRows`).

This PR doesn't include any additional tests, there doesn't seem to be tests for individual flags (correct me if I'm wrong). The test-suite passes locally with the flag enabled


aarondl:
See CONTRIBUTING.md, this needs to change branches.

jhnj:
My bad, will do

PR #961, opened 2021-06-07
Add always-wrap-errors flag

jhnj:
Currently all methods on models except for `Model.One(...)` and `FindModel(...)` always return wrapped errors with stacktraces. If no rows are returned `Model.One(...)` and `FindModel(...)` unwrap the returned error and return `sql.ErrNoRows` without  a stacktrace.

This PR adds an option, `always-wrap-errors` that disables this unwrapping. Comparing wrapped errors is supported in the standard library with `errors.Is(err, sql.ErrNoRows)`, and e.g. in our codebase we have a linter that doesn't allow direct equality checks when comparing errors (`err == sql.ErrNoRows`).

This PR doesn't include any additional tests, there doesn't seem to be tests for individual flags (correct me if I'm wrong). The test-suite passes locally with the flag enabled

Previously opened this PR against the master branch #955 - new try against dev here

### DIFF
--- a/boilingcore/boilingcore.go
+++ b/boilingcore/boilingcore.go
@@ -140,6 +140,7 @@ func (s *State) Run() error {
 		NoRowsAffected:    s.Config.NoRowsAffected,
 		NoDriverTemplates: s.Config.NoDriverTemplates,
 		NoBackReferencing: s.Config.NoBackReferencing,
+		AlwaysWrapErrors:  s.Config.AlwaysWrapErrors,
 		StructTagCasing:   s.Config.StructTagCasing,
 		TagIgnore:         make(map[string]struct{}),
 		Tags:              s.Config.Tags,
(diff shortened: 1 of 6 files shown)


aarondl:
Trying to figure out why this is necessary. Does `errors.Is` fail? It's fine you don't have to use equality tests but `errors.Is(err, sql.ErrNoRows)` does not require wrapping for it to work correctly:
https://play.golang.org/p/0dn-ZGReTpv

jhnj:
@aarondl `errors.Is` works fine for us. I'll try to elaborate on what the issue we're trying to solve with this is:

We want all of our errors to have stacktraces. Right now, for all of our sqlboiler `SELECT` calls, we have to `return errors.WithStack(err)` because if the returned error from sqlboiler is `sql.ErrNoRows` it doesn't have a stacktrace. There's a lot of situations where we assume that the call will always return a row and we thus don't want to handle `sql.ErrNoRows` separately. 

The reason we always want stacktraces is that we had some situations where we needed to add additional logging in order to figure out where some `sql.ErrNoRows` actually originated from.

With this option we wouldn't need to wrap the errors from sqlboiler in stacktraces manually.

aarondl:
@jhnj sorry need a rebase here :(

jhnj:
@aarondl rebased. Thanks for taking a look at this :pray:

aarondl:
Merged into dev. Thanks for this.

PR #962, opened 2021-06-10
Add configuration for timestamp and soft delete columns

stephenafamo:
## Use Case

I am inheriting a codebase which has the timestamp columns as `createdAt` and `updatedAt`.

I thought it would make sense to allow the timestamp columns to be configurable, while making this change, I thought it made sense to extend it to the `deleted_at` column for soft deletes.


aarondl:
I didn't really want to accept this because sqlboiler has many issues when you don't use it's naming scheme but presumably you use this patch to completely enable a use case so accepted. Thanks for the PR!

templates/18_delete.go.tpl
```diff
@@ -69,7 +69,7 @@ func (o *{{$alias.UpSingular}}) Delete({{if .NoContext}}exec boil.Executor{{else
 	} else {
 		currTime := time.Now().In(boil.GetLocation())
 		o.DeletedAt = null.TimeFrom(currTime)
-		wl := []string{"deleted_at"}
+		wl := []string{(or $.AutoColumns.Deleted "deleted_at")}
```
ryosukee, line 72: @stephenafamo @aarondl 

I'm sorry if i'm wrong. 
I think `{{` and `}}` should be used for `Actions` of template like below example. Is the formatting of this line correct? 
<details>
<summary> my sample code </summary>

```sh
$ go version
go version go1.16.4 darwin/amd64
$ go run hoge.go
input
wl := []string{(or $.AutoColumns.Deleted "deleted_at")}
output
wl := []string{(or $.AutoColumns.Deleted "deleted_at")}
output with empty
wl := []string{(or $.AutoColumns.Deleted "deleted_at")}

input
wl := []string{"{{ or $.AutoColumns.Deleted "deleted_at" }}"}
output
wl := []string{"deletedAt"}
output with empty
wl := []string{"deleted_at"}
```

```go
package main

import (
	"bytes"
	"fmt"
	"html/template"
)

type AutoColumns struct {
	Deleted string
}

type Conf struct {
	AutoColumns AutoColumns
}

func main() {
	original := `wl := []string{(or $.AutoColumns.Deleted "deleted_at")}`
	format(original)

	changed := `wl := []string{"{{ or $.AutoColumns.Deleted "deleted_at" }}"}`
	format(changed)
}

func format(templ string) {
	fmt.Println("input")
	fmt.Println(templ)

	tpl, err := template.New("").Parse(templ)
	if err != nil {
		fmt.Println("new err")
		fmt.Println(err)
	}

	autoCol := AutoColumns{Deleted: "deletedAt"}
	autoColEmpty := AutoColumns{}
	conf := Conf{AutoColumns: autoCol}
	confEmpty := Conf{AutoColumns: autoColEmpty}

	buf := &bytes.Buffer{}
	if err := tpl.Execute(buf, conf); err != nil {
		fmt.Println("exec err")
		fmt.Println(err)
	}
	fmt.Println("output")
	fmt.Println(string(buf.Bytes()))

	buf = &bytes.Buffer{}
	if err := tpl.Execute(buf, confEmpty); err != nil {
		fmt.Println("exec err")
		fmt.Println(err)
	}
	fmt.Println("output with empty")
	fmt.Println(string(buf.Bytes()) + "\n")
}
```

</details>


Actually, I'm just failing sqlboiler command at this point as showing below.
The version is `SQLBoiler v4.7.1`.
```
Error: unable to generate output: failed to format template

 1037           args = queries.ValuesFromMapping(reflect.Indirect(reflect.ValueOf(o)), brandPrimaryKeyMapping)
 1038           sql = "DELETE FROM `brands` WHERE `id`=?"
 1039   } else {
 1040           currTime := time.Now().In(boil.GetLocation())
 1041           o.DeletedAt = null.TimeFrom(currTime)
>>>>            wl := []string{(or $.AutoColumns.Deleted "deleted_at")}
 1043           sql = fmt.Sprintf("UPDATE `brands` SET %s WHERE `id`=?",
 1044                   strmangle.SetParamNames("`", "`", 0, wl),
 1045           )
 1046           valueMapping, err := queries.BindMapping(brandType, brandMapping, append(wl, brandPrimaryKeyColumns...))
 1047           if err != nil {

: 1042:22: illegal character U+0024 '$' (and 10 more errors)
ERROR: 1
make: *** [sqlboiler] Error 1
```

Is this template work correctly?

stephenafamo: No it is not. Working on a fix now.

ryosukee: I got it. Thank you for your quick reply!

PR #963, opened 2021-06-11
fix(orderBy): Update orderByQueryMod to support parameters

### LINKED ISSUE #546: Escaping parameters in an order-by clause

alexdor:
This commit adds optional parameters to the orderby query, in order to
expose the work done by @emwalker to the consumers of the library

Fixes #546

### DIFF
--- a/queries/qm/query_mods.go
+++ b/queries/qm/query_mods.go
@@ -458,17 +458,19 @@ func GroupBy(clause string) QueryMod {
 
 type orderByQueryMod struct {
 	clause string
+	args   []interface{}
 }
 
 // Apply implements QueryMod.Apply.
 func (qm orderByQueryMod) Apply(q *queries.Query) {
-	queries.AppendOrderBy(q, qm.clause)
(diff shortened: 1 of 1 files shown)


aarondl:
Thanks for this!

PR #970, opened 2021-06-30
Update 16_update.go.tpl

chochihim:
I often find myself needs to update a record using column(s) other than primary key. This enables use case like this:

```go
type UserWallet struct {
  ID      string        `boil:"id" json:"id" toml:"id" yaml:"id"`
  Type    string        `boil:"type" json:"type" toml:"type" yaml:"type"`
  UserID  string        `boil:"user_id" json:"user_id" toml:"user_id" yaml:"user_id"`
  Balance types.Decimal `boil:"balance" json:"balance" toml:"balance" yaml:"balance"`

  R *userWalletR `boil:"-" json:"-" toml:"-" yaml:"-"`
  L userWalletL  `boil:"-" json:"-" toml:"-" yaml:"-"`
}


// construct the struct manually without `find` first to save one db operation
userWallet := models.UserWallet{
  UserID: userID,
  Balance: balance,
}

userWallet.UpdateBy(ctx, db, boil.Whitelist("amount"), []string{"user_id"})
```

If this is accepted,
1. should I also add `UpdateByG`, `UpdateByP` etc variants?
2. should I also add `UpdateAllBy` function?

Also I have another idea to make this use case more complete but the work required is beyond the scope of this PR.

In above example, fields other than `userWallet.UserID` and `userWallet.Balance` are invalid.
So after the update operation, the struct is kind of useless. If we utilize the `UPDATE ... RETURNING ...` clause in `psql`,
we can bind the result and the struct is fully available.
The function signature can be something like `UpdateBy(..., columns boil.Columns, updateBy []string, returningColumns boil.Columns) (int64, error)`.

What do you think?

### DIFF
--- a/templates/16_update.go.tpl
+++ b/templates/16_update.go.tpl
@@ -45,6 +45,13 @@ func (o *{{$alias.UpSingular}}) UpdateGP({{if not .NoContext}}ctx context.Contex
 // See boil.Columns.UpdateColumnSet documentation to understand column list inference for updates.
 // Update does not automatically update the record in case of default values. Use .Reload() to refresh the records.
 func (o *{{$alias.UpSingular}}) Update({{if .NoContext}}exec boil.Executor{{else}}ctx context.Context, exec boil.ContextExecutor{{end}}, columns boil.Columns) {{if .NoRowsAffected}}error{{else}}(int64, error){{end -}} {
+	return o.UpdateBy({{if not .NoContext}}ctx, {{end -}} exec, columns, nil)
+}
+
+// UpdateBy does the same as Update except it allows updated by columns other than primary keys.
+// If updateBy is nil, the behaviour will be the same as Update.
+// See Update for more documentation.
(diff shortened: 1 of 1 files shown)


stephenafamo:
I believe you can already achieve the same effect by doing something like this:

```go
_, err := models.UserWallets(
    models.UserWalletWhere.UserID.EQ(userID),
).UpdateAll(ctx, db, models.M{"balance": balance})
```

Can you clarify if this does not do what you want?

chochihim:
@stephenafamo Yes, both ways do the same, but by using the struct, the `balance` field is type-checked. Using `models.M` does not.

stephenafamo:
I get your point, but honestly, the reason I don't like the `UpdateBy` idea is that it feels unexpected.  
If I'm working with a single `UserWallet` model, I wouldn't expect that one of its methods could potentially update multiple rows.

Instead, maybe we can modify the `UpdateAll` methods of each model to use a more type-safe argument instead of `models.M{}`.

chochihim:
@stephenafamo You're right. I miss that it can update multiple rows. In my use case the `user_id` is unique key on the table. So what I want should be something like `UpdateByUniqueKey` instead of `UpdateBy`.

I think if I need to implement such a function, sqlboiler needs to make use of unique key in model generation, which it does not now.

I have also thought about modifying `UpdateAll` but a lot of work is required. Together with the `RETURNING` clause I suggested, I think I better open a discussion first. I will close this PR now.

chochihim:
The discussion: https://github.com/volatiletech/sqlboiler/discussions/971

PR #973, opened 2021-07-17
Load column comment for mysql

Wuvist:
Currently mysql driver is not loading column comment from db.

This PR should load them and be generated in templates.


aarondl:
Thanks for this. The original implementer ignored all other drivers outside psql :)

PR #983, opened 2021-08-20
Fix forgetting to call rows.Close

kawabatas:
i found rows.Close is not called in ForeignKeyInfo function.


aarondl:
This isn't actually necessary since iteration should never stop in the middle:
https://pkg.go.dev/database/sql#Rows.Close

PR #984, opened 2021-08-23
Support sqlcommenter comments

gaiottino:
https://google.github.io/sqlcommenter/ makes it possible to add tracing comments that can be interpreted by e.g. GCPs query analyser.

https://github.com/volatiletech/sqlboiler/pull/872 added support for comments but they are prepended on a separate line and sqlcommenter requires the form `/* … */` to be added on the same line as the query which is why I’ve added this PR and the `qm.AppendComment` function.

For example:
```
q := shop.Customers(
  qm.AppendComment("shop.getAllCustomers"),
  qm.Select("id", "name"),
  qm.From("customers"),
)
```
will be:
```
SELECT id, name FROM customers; /* shop.getAllCustomers /*
```

### DIFF
--- a/queries/query.go
+++ b/queries/query.go
@@ -30,22 +30,23 @@ type Query struct {
 	load     []string
 	loadMods map[string]Applicator
 
-	delete     bool
-	update     map[string]interface{}
-	withs      []argClause
-	selectCols []string
-	count      bool
-	from       []string
-	joins      []join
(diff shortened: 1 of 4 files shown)


aarondl:
You show this:  `qm.AppendComment("shop.getAllCustomers")`

but I can't find that function in the PR?

PR #993, opened 2021-09-06
add option --add-strict-upsert for psql

sadayuki-matsuno:
Added --add-strict-upsert option to psql.
With this option, an upsert method is created for each constraint, and upsert can be executed without passing the conflictColumns variable.

If you specify `--add-strict-upsert`, upsertDoNothing and upsert for each primaryKey and uniqueKey will be created instead of upsert as before, only for psql.

Here are the reasons why I created this feature

1. to be able to notice errors in conflictColumns, since they are not known at compile time and the error is not noticed until execution.
2. to make it easier to understand which columns can be used for Upsert.

- before

```
func (o *XXX) Upsert(ctx context.Context, exec boil.ContextExecutor, updateOnConflict bool, conflictColumns []string, updateColumns, insertColumns boil.Columns) error
```

- after

```
func (o *XXX) UpsertDoNothing(ctx context.Context, exec boil.ContextExecutor, insertColumns boil.Columns) error
func (o *XXX) UpsertByXxxPkey(ctx context.Context, exec boil.ContextExecutor, updateColumns, insertColumns boil.Columns) error
func (o *XXX) UpsertByXxxUkey(ctx context.Context, exec boil.ContextExecutor, updateColumns, insertColumns boil.Columns) error
```


aarondl:
Perhaps we could achieve type safety in the same way we're suggesting here? https://github.com/volatiletech/sqlboiler/discussions/971#discussioncomment-1386709

This seems like a lot :D

Also noticing many `fix` commits over time. Is this complete now?

sadayuki-matsuno:
Conflict fixed.
This PR is complete.

aarondl:
@sadayuki-matsuno what do you think of the ideas that I pointed out instead of this PR?

This adds so much "stuff". Wouldn't it be nicer to achieve type safety through less prolific chunk of code? This also breaks compatibility with existing driver versions. I don't know that we can merge this as is.

sadayuki-matsuno:
I have not been able to grasp what exactly your idea is.
I think the compatibility part definitely needs more consideration, but I like this PR because it is a simple implementation for users.

stephenafamo:

Although some discussion is needed so it may be better to first open a discussion *before* a new PR.

PR #998, opened 2021-09-24
Adding Debug Formatter function and Query duration to debug info

5anthosh:
Adding debug formatter function makes sqlboiler more extensible and lets users to more customizable debug logs

Query duration will be useful during performance improvement

### DIFF
--- a/queries/query.go
+++ b/queries/query.go
@@ -3,8 +3,8 @@ package queries
 import (
 	"context"
 	"database/sql"
-	"fmt"
 	"regexp"
+	"time"
 
 	"github.com/volatiletech/sqlboiler/v4/boil"
 	"github.com/volatiletech/sqlboiler/v4/drivers"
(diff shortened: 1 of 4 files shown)


queries/query.go
```diff
@@ -117,64 +117,88 @@ func RawG(query string, args ...interface{}) *Query {
 // Exec executes a query that does not need a row returned
 func (q *Query) Exec(exec boil.Executor) (sql.Result, error) {
 	qs, args := BuildQuery(q)
+	before := time.Now()
```
burkostya, line 120: Do we need to know the "now" for not debug mode?

5anthosh: No,  should I add it under Debug check?

burkostya: I think so. Execs and queries are on hot path so any new allocation could impact performance.

aarondl:
There are some changes to the version number and the changelog that should not be in this PR.

The comment above also needs addressing.

wsantos:
I'd love to see this merged, @5anthosh any change to update this PR ?

stephenafamo:

Also kindly address the comments.

PR #1000, opened 2021-09-27
Move templates, restore boilingcore.New compatibility

### LINKED ISSUE #999: New use of go:embed prevents use of boilingcore as library
TL;DR: The latest release removed bindata in favor of `go:embed` (213f39685322dbd79e4f188eeedeba0d614ae72f), but in doing so prevents use of the generation code as a library.

-- 

I run sqlboiler's generation code directly so I can avoid having platform-specific scripts or external dependencies: https://github.com/hortbot/hortbot/blob/master/internal/cmd/genmodels/main.go

The change made to `boilingcore` is not backwards/SIV compatible, as it changed the signature of `New`; that's not good itself, but could be worked around. Unfortunately, the FS that is needed to make things work is defined in `main.go`, which isn't accessible (as it's both the main package and unexported). These two together basically mean that nobody can use SQLBoiler's generation code as a library anymore in their own code.

I'm thinking the pattern chosen was only chosen because `templates` and `templates_test` are root directories in the repo. The function itself is only used twice in the project, once in the `main.go` with the embedded data, and once in the tests with `os.DirFS` to access a parent directory.

Can these two folders be placed in `boilingcore` instead, then just have a single embbedded FS that's used, restoring `New` to its old parameter?

I'm more than happy to submit a PR to do this.

zikaeroh:
Fixes #999.

`boilingcore` is the only place these templates are used; just declare `templatesBuiltin` there and use the same `embed.FS`. This restores the old `boilingcore.New` signature (fixing compiler issues).


zikaeroh:
Hm, no idea what would have failed here. Will debug. :smile:

zikaeroh:
The tests are failing due to a bug in the new loading code; the pattern `*/*.tpl` is not enough to get nested templates, so:

- When running `sqlboiler`, the `singleton/*.tpl` templates are not loaded.
- When running `boilbench`'s tests, `os.DirFS` points to the wrong folder, and the wrong templates are loaded (but "pass" because they don't generate anything, which `go test` is happy to accept as a passing package).

zikaeroh:
I've fixed the bug by making the code in `initTemplates` more similar to the code that was there for gobindata.

Regardless of #999, this is a bugfix that is likely needed, but it's also definitely needed to make this PR work (as it exposes the bug in the spuriously-passing test).

zikaeroh:
Having switched locally to my fork for my SQLBoiler-using project, I'm happy to say that this works.

aarondl:
Closing in favor of what's been released. It would have been a lot easier to just merge this, but I really don't like the idea of templates going into boilingcore, it's just not right on so many levels. Thanks for providing all the fixes regardless Zik. You're an sqlboiler hero :)

PR #1006, opened 2021-10-01
fix bug with soft deletes and custom auto-columns

stephenafamo:
#962 seems to have broken soft deletes.

1. The generation fails because of a syntax error
2. Even if it succeeded, soft deletes would be disabled if the user changed the `deleted_at` column name since `table.CanSoftDelete()` did not take the customization into account.

I've fixed these in this PR, tests are all passing and from my own personal testing, it seems to be working as expected.

### DIFF
--- a/drivers/table.go
+++ b/drivers/table.go
@@ -65,9 +65,13 @@ func (t Table) CanLastInsertID() bool {
 	return true
 }
 
-func (t Table) CanSoftDelete() bool {
+func (t Table) CanSoftDelete(deleteColumn string) bool {
+	if deleteColumn == "" {
+		deleteColumn = "deleted_at"
+	}
+
 	for _, column := range t.Columns {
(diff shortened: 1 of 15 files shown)


ping-localhost:
Any idea when this will be merged? I'm trying out `sqlboiler` to see if it's a suitable replacement for the ORM I use now and I do need the soft-delete functionality as I heavily rely on it :)

stephenafamo:
@ping-localhost you can maybe use 4.6.0 in the meantime. Sorry I broke this 😅

ping-localhost:
> @ping-localhost you can maybe use 4.6.0 in the meantime. Sorry I broke this 😅

That's is what I'm doing for now, so I can keep experimenting. No worries and thank you for the quick reply!

aarondl:
Merged. Thanks for the fix.

PR #1014, opened 2021-10-18
Collect cascade option on FKs

ccakes:
Initially Postgres-only, this collects any actions configured on foreign keys during database introspection so that data can be used in the template generation later.

I'm running SQLBoiler with some pretty custom templates, this information is useful for being able to know in Go what will happen in the database when keys are updated or deleted. Unsure if you have a policy of only accepting features that are engine-agnostic but thought I'd throw this up for feedback to potentially save continuing to maintain a fork

### DIFF
--- a/drivers/keys.go
+++ b/drivers/keys.go
@@ -2,6 +2,16 @@ package drivers
 
 import "fmt"
 
+type ForeignKeyAction string
+
+const (
+	ForeignKeyNoAction   ForeignKeyAction = "a"
+	ForeignKeyRestrict                    = "r"
+	ForeignKeyCascade                     = "c"
+	ForeignKeySetNull                     = "n"
+	ForeignKeySetDefault                  = "d"
+)
+
 // PrimaryKey represents a primary key constraint in a database
 type PrimaryKey struct {
 	Name    string   `json:"name"`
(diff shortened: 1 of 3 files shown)


aarondl:
Thanks for submitting the PR and checking. We don't generally accept things that would only help forks or don't have great usecases or reasoning others can lean on. We also try to keep to the lowest common denominator in terms of feature set across all 3 supported drivers.

What's the use case for your custom templates to need these fields? Is there something you'd be interested in bringing into sqlboiler's default template set that you're accomplishing with these?

ccakes:
> We don't generally accept things that would only help forks or don't have great usecases or reasoning others can lean on. We also try to keep to the lowest common denominator in terms of feature set across all 3 supported drivers.

That's perfectly fair 👍 I'll outline my use case below and if you think it's interesting and something you'd merge I'm happy to spend some time making it work on MySQL and MS-SQL as well. If not - absolutely fine, feel free to close this MR 🙂 

Admittedly this is quite specific, but the `ON ... CASCADE` is part of ANSI SQL so can be implemented for all the supported DB engines. If many users are heavily customising templates it might be worth merging (a more complete impl) as an extra knob to use in the template.

### Use case
I have some entities where I want to record changes to them over time and generate a custom diff format that I can use to revert those changes. Think audit log but where the audit contains data to revert the change.

The database schema is normalised so we have plenty of relationships between elements, some of which have cascading side effects. If a change to my tracked entity results in a cascading change I want to capture that and generate a revert for it as well.

The achieve this, I have an instance of a "collector" that I store in the context passed to `boil` and generate code using the templates to capture the data I'm interested in on changes, for types with cascades they're followed here to capture the full extent of the change.

stephenafamo:
This is a nice PR, but before we can merge it would need to at least support MySQL and MS-SQL

stephenafamo:

ccakes:
> Kindy reopen this PR to master, it was mistakenly closed when the dev branch was deleted.

I don't have the bandwidth at the moment to spend the time to make this work for MySQL or MSSQL so I'll hold off re-opening for now, and maybe come back to it later 👍

PR #1020, opened 2021-11-04
add output of sql when error

hxchjm:
more details are necessary when error is happend，it's convenient for user to confirm where the problem is.


aarondl:
`boil.Debug()` was created to solve this problem. It is dangerous to give the sql and the arguments in an error as it presents opportunities to leak PII and other information into the logs. Won't merge this sorry :(

PR #1026, opened 2021-11-15
Fix assignment in template code

Darkclainer:
In if branch new variable `$enumValue` was initialized and destroyed,
effectually making branch useless.


optiman:
Can you also add this fix on line 113? Just to keep all related fixes in this PR.
https://github.com/volatiletech/sqlboiler/blob/9ec65427f70d62796381a15a1b36eb3443d17396/templates/main/singleton/boil_types.go.tpl#L113
It's for `--add-enum-types` case.

aarondl:
Because this missed the 2nd case mentioned by optiman, I made the change myself. Thanks for the PR. Releasing 4.8.2 to rectify this.

aarondl:
https://github.com/volatiletech/sqlboiler/releases/tag/v4.8.3
Should be fixed here.

PR #1032, opened 2021-11-18
Enum types for model fields

optiman:
This PR adds enum types to model fields (if `add-enum-types` is true).
- For non-nullable columns types added in `v4.8.0` are used.
- For nullable columns new nullable types are generated (basically `null.String`, but typed)
In case of name collisions prefix for nullable enum types can be changed with `--enum-null-prefix` (default "Null").

Drivers for PostgreSQL and MySQL are updated to support this feature.

Unfortunately, `Constructor` interface (that is implemented by drivers) needed to be changed because MySQL enums are by convention named as `TableNameColumnName`, and `TranslateColumnType` method previously had no knowledge of what table is being processed.
```diff
-    TranslateColumnType(Column) Column
+    TranslateColumnType(col Column, tableName string) Column
```

Ready to discuss if this can be implemented without incompatible changes (though they are incompatible only for third-party drivers).
I'm ready to update CockroachDB driver to support this change. This leaves only SQLite driver.


aarondl:
The backwards incompatible change is bad :( This means people's custom drivers may break. I don't know that this is acceptable as is until v5 - but I also don't know how many custom drivers are out there in the wild / if this feature is being leveraged.

Even so in order for this to land though we'd need commitments to fix all drivers referenced in the readme including sqlite3.

Also I'm curious about the benefit that this PR brings. Is it that nullable enum types work better with Where helpers now?

optiman:
I reverted the interface back, no BC break anymore.
But to do so, I needed to add a new helper interface for MySQL or any other database that needs the table name to generate go column names - `TableColumnTypeTranslator`. Implemented this interface in internal MySQL driver, all other drivers should work as before.

> Also I'm curious about the benefit that this PR brings. Is it that nullable enum types work better with Where helpers now?

I'm sorry, isn't the whole purpose of this project to make a type safe ORM?

The README says:
- `The models package is type safe`
- `Our types closely correlate to your database column types`
- `Auto-completion provides work-flow efficiency gains`

Since strict enum types were already added in `v4.8.0`, why not use them in generated models? 
It's a convenience feature - you can Ctrl-Click model fields in IDE and see the real types instead of generic `null.String`. Also the IDE shows type hints for `Where` helpers. It doesn't require any more memory allocation. What's the downside?

aarondl:
Haha. Sorry @optiman you misunderstand me. Obviously I don't need my own README quoted at me to remember what the library's all about lmao.

In my recent days as a maintainer of this library I've grown very apathetic about the project. I entertain changes as a matter of guilt not pleasure, it's duty and one that's hard to get excited about since I haven't been using sqlboiler much myself recently. Often I wait a month between looking at incoming PRs and bugs because frankly I don't want to deal with it, but I don't want to burden anybody else by giving it away nor do I want to hurt existing users by completely archiving it, there's many people who enjoy using this in their day to day.

When I saw your change here I'm wondering: "What does this actually do?" - meaning if I turned on this flag what's the end result for a user?

>    For non-nullable columns types added in v4.8.0 are used.
>    For nullable columns new nullable types are generated (basically null.String, but typed)
>    In case of name collisions prefix for nullable enum types can be changed with --enum-null-prefix (default "Null").

In this description I'm not sure -where- they were used, it doesn't really specify and the added context of the enum types is lost on me because I don't really remember that change. There's lots of places this could be used. Maybe it's only in where helpers, maybe it's in the structs, maybe it's in the generated enum vars, I don't actually know :)

In my inbox I have 30 sqlboiler related emails piled up and it's pretty easy to get lazy in evaluating each PR. I didn't clone this, run this, or give it much thought other than I saw the red flag of breaking changes that need fixing before I'm even willing to look at the code change. So at the same time I asked: "What does this actually do?" and instead of reply by saying: "It replaces generated struct fields with nullable types where previously it would have been X which isn't as nice because etc. etc. here's a nice example etc. etc" you seem to have gotten frustrated and a touch passive aggressive.

For what it's worth I apologize for being a fairly detached maintainer. But that said, I expect that PRs prove their own worth and describe in no uncertain terms what the change is about and why sqlboiler users / you as an sqlboiler user would want it.

This ask is spelled out here: https://github.com/volatiletech/sqlboiler/blob/master/CONTRIBUTING.md (as is the branch you're supposed to open PRs against which is incorrect in this PR). Our views may differ on what constitutes enough of a description but in this case it doesn't seem like it was enough.

optiman:
I didn't mean to be passive aggressive, sorry about that. 
I just thought it was pretty self-explanatory. Sqlboiler generates model types, they have fields, which now will use enum types, that were added previously. That's it.
The burden of supporting an open-source project is real - that's understandable, no one should be expected to work for free. And I want to thank you for your work on this awesome project.

That said, should I open a new PR against the `dev` branch or is this feature just not as useful as I thought?

importers/imports.go
```diff
@@ -285,6 +285,27 @@ func NewDefaultImports() Collection {
 	return col
 }
 
+// NewNullableEnumImports returns imports collection for nullable enum types.
+func NewNullableEnumImports() Collection {
```
stephenafamo, line 289: Is there a reason the name of this function starts with **New**? Perhaps we can just call it `NullableEnumImports`

drivers/sqlboiler-mysql/driver/mysql_test.go
```diff
@@ -18,13 +18,15 @@ import (
...
+	"github.com/stretchr/testify/require"
+
 	"github.com/volatiletech/sqlboiler/v4/drivers"
 )
 
 var (
 	flagOverwriteGolden = flag.Bool("overwrite-golden", false, "Overwrite the golden file with the current execution results")
 
-	envHostname = drivers.DefaultEnv("DRIVER_HOSTNAME", "localhost")
+	envHostname = drivers.DefaultEnv("DRIVER_HOSTNAME", "127.0.0.1")
```
stephenafamo, line 29: I assume this was changed by accident. Can we revert?

drivers/sqlboiler-mysql/driver/mysql_test.go
```diff
@@ -18,13 +18,15 @@ import (
 	"os/exec"
 	"testing"
 
+	"github.com/stretchr/testify/require"
+
```
stephenafamo, line 22: I think this blank space is redundant.

stephenafamo (changes requested):
This looks really nice. Just some minor changes and it should be ready to be merged.

importers/imports.go
optiman: No reason, just an old habit from pre-Go days. Fixed.

drivers/sqlboiler-mysql/driver/mysql_test.go
optiman: Reverted.

drivers/sqlboiler-mysql/driver/mysql_test.go
optiman: Yes, my IDE just likes to group packages from the main module separately. Fixed.

optiman:
@stephenafamo 
Done.

stephenafamo:
The tests are failing... can you please check and fix?

optiman:
Forgot to revert the MockDriver. Can you re-run the test?

https://github.com/optiman/sqlboiler/actions/runs/1715141576

stephenafamo:
Still failing... looks to be related to a recent change about null columns and default values. Perhaps merge from dev again and fix.

optiman:
Merged latest changes from dev. Seems fine now.

https://github.com/optiman/sqlboiler/actions/runs/1715463478

stephenafamo (approved):

PR #1033, opened 2021-11-19
Update volatiletech/null dependency to v9.0.0

ian-fox:
This PR updates the dependency on volatiletech/null to the newly released [v9.0.0](https://github.com/volatiletech/null/commit/ad763b656f3239c1d8127db501bd4b148de59712).

This might technically be a breaking change that requires waiting until the next major version, as anybody serializing a sqlboiler-generated type to JSON will see a different serialization than before if their type contains a `null.Bytes` field.

### DIFF
--- a/README.md
+++ b/README.md
@@ -302,7 +302,7 @@ into your `go.mod` file at the correct version.
 # Do not forget the trailing /v4 and /v8 in the following commands
 go get github.com/volatiletech/sqlboiler/v4
 # Assuming you're going to use the null package for its additional null types
-go get github.com/volatiletech/null/v8
+go get github.com/volatiletech/null/v9
 ```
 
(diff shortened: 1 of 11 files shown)


aarondl:
This is a breaking change in more than one way unfortunately. Because of the way this was implemented:

```
null.Bool{Valid: true, Bool: true}
```

This no longer creates a valid bool, but was a key part of the way the API ended up being used because these fields were exported. Unfortunately we'll have to hold off. I will also say that I'm keenly interested in a Generics based null package instead of continuing with v9. So likely sqlboiler will drop support for `null` as it stands and use a fresh new package in a breaking v5.

Really wish we could add this though. Appreciate the PR even if we couldn't add it at this time.

PR #1035, opened 2021-11-29
Support * wildcard as table name in black/white listing of columns

Yoshiji:
This PR adds support for `*` as a wildcard for the table name in the white/black listing of columns.

This allows one to use `blacklist = ["*.tenant_id"]` to prevent all `tenant_id` columns (from any table) related code to be generated, rather than explicitly listing all tables having this column.

---

Unrelated note: Thank you for this amazing lib!

### DIFF
--- a/README.md
+++ b/README.md
@@ -359,10 +359,11 @@ Example of whitelist/blacklist:
 
 ```toml
 [psql]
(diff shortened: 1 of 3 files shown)


aarondl:
This seems fine. Thanks for this :)

PR #1038, opened 2021-12-04
Support max 512 columns/fields instead of 256

xc:
256 in a struct(not including embed struct) is bit small for us. We have 260 fields for now and may add more.

Also max columns better to be documented(in readme?).

Thanks.
XC

### DIFF
--- a/queries/reflect.go
+++ b/queries/reflect.go
@@ -31,7 +31,7 @@ const (
 	loadMethodPrefix       = "Load"
 	relationshipStructName = "R"
 	loaderStructName       = "L"
-	sentinel               = uint64(255)
+	sentinel               = uint64(511)
 )
 
 // BindP executes the query and inserts the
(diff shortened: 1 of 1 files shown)


aarondl:
Unfortunately I think this is a backwards incompatible change. The way the pointer mappings work is that a single field is represented by a uint64 divided into 8 uint8.

Each byte represents a level of recursion into a struct, ie:
```go
type User struct {
   OneLevel struct {
       TwoLevel struct {
       } `boil:"two_level"`
   } `boil:"one_level"`
}
```

This PR does two things: Increases the column max columns, but unfortunately it also trades for max recursive depth as we're using 9 bits per column now meaning we can only fit 7 levels of depth.

As is I don't think I can accept this.

xc:
Thanks for the answer.

Backward compatible wise it's true. For us and for most of users I think the more important is the field number instead of struct depth, but it's indeed about backward compatibility.

Maybe it's better to use 2 dimension slice eg. [10][512]int(one for struct depth, one for that depth's fields) instead of uint64? then the 'mapping' variable will not be uint64, but [2]int(eg.{0,20}). Not sure if it will affect performance.  I'm not sure I'm able to write a PR but I can have a try.

aarondl:
It was definitely made this way for performance concerns so that's top of mind.

xc:
Give me some time to think if there is other possible to increase column number. Will update this.

PR #1055, opened 2022-01-14
Prevent codegen panic when excluding columns refering foreign keys

zapo:
This updates the filterForeignKeys function to remove foreign keys related to columns that are excluded using blacklist/whitelist options respecting the new wildcard option.
This prevents codegen to panic when setting up foreign key constraints for unknown columns and properly exclude the relationship.

### DIFF
--- a/drivers/interface.go
+++ b/drivers/interface.go
@@ -131,12 +131,24 @@ func Tables(c Constructor, schema string, whitelist, blacklist []string) ([]Tabl
 	return tables, nil
 }
 
+func knownColumn(table string, column string, whitelist, blacklist []string) bool {
+	return (len(whitelist) == 0 ||
+		strmangle.SetInclude(table, whitelist) ||
+		strmangle.SetInclude(table+"."+column, whitelist) ||
+		strmangle.SetInclude("*."+column, whitelist)) &&
+
+		(len(blacklist) == 0 || (!strmangle.SetInclude(table, blacklist) &&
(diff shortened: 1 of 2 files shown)


stephenafamo:
Can you clarify the use case here?

Is there a reason why you want to exclude a column referred to by a foreign key constraint?

zapo:
Yes for sure @stephenafamo. We're implementing table level multi tenancy using postgres RLS & triggers to scope database access / mutations for a specific tenant (enforced at the database level rather than in the application for now). To do so we're adding a tenant_id column on all tables but we don't really need sqlboiler to be aware of those foreign keys to our tenants table.

The main issue is with join tables which are no longer being detected as such because of the additional column, and prevent the related helpers from being properly generated.

stephenafamo:
So, to understand, you have a `tenant_id` column on all the tables, but you have all of them blacklisted from Sqlboiler.

And now you also want all the relation helpers to be ignored right?

zapo:
> And now you also want all the relation helpers to be ignored right?

Yes, that is what this PR aims to do since generating relation helpers for ignored columns currently panic on the GetColumn calls at:
https://github.com/volatiletech/sqlboiler/blob/475493476401b3fd8fd0801eece058e514d284d7/drivers/interface.go#L172-L174

Also even outside of my usecase, I feel that this would be the expected behavior from ignoring a column participating in a relationship.

stephenafamo:
Alright. Makes sense. Merging...

PR #1078, opened 2022-01-28
Simultaneous generation of the same nullable and non-nullable enum type (fix)

optiman:
**Problem:**
When using PostgreSQL/CockroachDB if the same enum type is used for both nullable and non-nullable columns (across all tables), only the Go type of the first column is being generated at the moment (because of the `onceHas` check in the template).
https://github.com/volatiletech/sqlboiler/blob/7e7fd0cb90d18291c74879109350d3448c97e87c/templates/main/singleton/boil_types.go.tpl#L70
Also if the non-nullable enum type was generated first, the detection of nullable enum column will cause the rendering of `NullableEnumImports`, thus corrupting the generated `boil_types.go` file, because these imports aren't going to be used (nullable Go type that uses those imports not generated because of the `onceHas` check).

**Solution**:
In addition to the `$once` map in `boil_types.go.tpl` add `$onceNull` map to save the names of generated nullable enum types and check them separately.

### DIFF
--- a/templates/main/singleton/boil_types.go.tpl
+++ b/templates/main/singleton/boil_types.go.tpl
@@ -61,73 +61,89 @@ MySQL output looks like:    TableNameColNameEnumValue = "enumvalue"
 It only titlecases the EnumValue portion if it's snake-cased.
 */}}
 {{$once := onceNew}}
+{{$onceNull := onceNew}}
 {{- range $table := .Tables -}}
 	{{- range $col := $table.Columns | filterColumnsByEnum -}}
 		{{- $name := parseEnumName $col.DBType -}}
 		{{- $vals := parseEnumVals $col.DBType -}}
 		{{- $isNamed := ne (len $name) 0}}
 		{{- $enumName := "" -}}
-		{{- if not (and $isNamed (onceHas $once $name)) -}}
-			{{- if $isNamed -}}
-				{{$_ := oncePut $once $name}}
-			{{- end -}}
+		{{- if not (and
+			$isNamed
+			(and
+				(onceHas $once $name)
+				(onceHas $onceNull $name)
+			)
+		) -}}
 			{{- if and (gt (len $vals) 0) (isEnumNormal $vals)}}
 				{{- if $isNamed -}}
 					{{ $enumName = titleCase $name}}
 				{{- else -}}
 					{{ $enumName = printf "%s%s" (titleCase $table.Name) (titleCase $col.Name)}}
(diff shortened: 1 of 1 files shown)


stephenafamo:
It seems that this PR fixes 2 issues and I can confirm one of them but I don't quite get what the other one is.

As you said, when the non-nullable column comes first, it breaks the type generation.  
However, if the nullable column was defined first, the generation seems to work fine.   

Am I missing something?

stephenafamo:
I have encountered another bug with your previous PR.

When `add-enum-types` is true, but there are no nullable enum types, `boil_types.go` contains a number of imports that are not used.

optiman:
1) Non-nullable Go type is just a string type alias with the name of DB type.
It has only two methods: `IsValid()` and `String()` and requires only `errors` package import, which is imported either way by default.
2) Nullable Go type is a struct type with methods like `MarshalJSON`, `UnmarshalJSON`, `Scan` that require additional imports, that are provided by function `NullableEnumImports()`:
https://github.com/volatiletech/sqlboiler/blob/7f42809cb14eabe6d721ffe190a53d0e185c1a28/importers/imports.go#L289-L307

These imports are rendered in `boil_types.go` if:
- `add-enum-types = true`
https://github.com/volatiletech/sqlboiler/blob/7f42809cb14eabe6d721ffe190a53d0e185c1a28/boilingcore/boilingcore.go#L94-L96
- There is at least one nullable column with enum type across all tables
https://github.com/volatiletech/sqlboiler/blob/7f42809cb14eabe6d721ffe190a53d0e185c1a28/boilingcore/boilingcore.go#L436-L440

So the imports for nullable types are already merged into `s.Config.Imports` before the `onceHas` check in `boil_types.go.tpl` prevents nullable Go type itself being rendered. 
This results in unused imports in final `boil_types.go`, which is considered an error by Go compiler.

If nullable Go type is generated first, these imports are being used correctly, but we need to do an additional check to not duplicate the base string alias type (with `IsValid()` and `String()` methods).

I know it's not the most elegant solution, but it's the most straightforward without doing a complete overhaul of current templates, which can cause more unexpected bugs.

stephenafamo:
Yes that is correct, I was testing the wrong branch.

Minor change.... instead of using `onceHas $once $name`, it would be better to do `$once.Has $name`. Same for `oncePut` and `$once.Put`.

I have done this locally, but I won't be able to push to your branch

optiman:
@stephenafamo
Done.

PR #1086, opened 2022-02-15
Add sprig FuncMap to templates

### LINKED ISSUE #659: Feature Request: Implement Sprig Template Library
https://masterminds.github.io/sprig/

I'm working on a custom template requires me to +1 to the index integer of a range of table columns, but I see no easy way to do this.

Over in Helm2 (for Kubernetes), they use the Sprig library to enrich the vanilla Go Template system.

Please let me know your thoughts.

ccakes:
Resolves #659 

This adds the well-known [Sprig](https://pkg.go.dev/github.com/Masterminds/sprig/v3) template functions to the sqlboiler template renderer.

In addition, and in a separate commit as it might be a little more controversial, this also adds a config option that can only be set by someone embedding sqlboiler as a library to allow the caller to add their own `template.FuncMap` to make extending simpler without needing to fork.

Feedback welcome. If the `UserFuncs` concept is acceptable to the maintainers let me know and I'll add some test coverage and documentation before we merge.

### DIFF
--- a/boilingcore/templates.go
+++ b/boilingcore/templates.go
@@ -12,6 +12,7 @@ import (
 	"strings"
 	"text/template"
 
+	"github.com/Masterminds/sprig/v3"
 	"github.com/friendsofgo/errors"
 	"github.com/volatiletech/sqlboiler/v4/drivers"
 	"github.com/volatiletech/strmangle"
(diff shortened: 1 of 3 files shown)


stephenafamo:
These make sense. However, I think it makes more sense to move the `UserFuncs` to a different PR.

ccakes:
No worries, I can do that 👍

ccakes:
I've split this up so we only have Sprig here, the `UserFuncs` are now implemented in #1088

boilingcore/templates.go
```diff
@@ -257,10 +263,10 @@ var templateStringMappers = map[string]func(string) string{
 
 var goVarnameReplacer = strings.NewReplacer("[", "_", "]", "_", ".", "_")
 
-// templateFunctions is a map of all the functions that get passed into the
+// boilerTemplateFunctions is a map of all the functions that get passed into the
 // templates. If you wish to pass a new function into your own template,
 // add a function pointer here.
```
stephenafamo, line 268: A little nitpicky, but I think it makes sense to update this comment too since it is no longer **all the functions**

stephenafamo (changes requested):

stephenafamo:
Copied the changes to #1097 and added a few minor changes on top.

I'll close this and merge that.

PR #1087, opened 2022-02-15
Minor query builder improvements

ccakes:
This PR implements the following:

- Added `queries.BuildSubquery()` which does the same thing as `BuildQuery()` except doesn't append a semicolon so query builder results can be embedded in other queries
- Added `qm.WithSubquery() which allows generating a CTE using a query object built by sqlboiler

The `WithSubquery()` modifier is interesting imo because if there's interest, there are a number of query mods that could be added to work with subqueries to allow building pretty complex, type-checked queries without horrible code.

### DIFF
--- a/queries/qm/query_mods.go
+++ b/queries/qm/query_mods.go
@@ -194,18 +194,30 @@ func Distinct(clause string) QueryMod {
 }
 
 type withQueryMod struct {
+	alias  string
 	clause string
 	args   []interface{}
 }
 
 // Apply implements QueryMod.Apply.
 func (qm withQueryMod) Apply(q *queries.Query) {
-	queries.AppendWith(q, qm.clause, qm.args...)
+	queries.AppendWith(q, qm.alias, qm.clause, qm.args...)
 }
 
 // With allows you to pass in a Common Table Expression clause (and args)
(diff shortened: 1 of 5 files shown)


stephenafamo:
This is quite interesting and it makes complete sense. If everything checks out and it is completely backward compatible, it would make sense to merge now (or wait till v5 if not).

Just so you know, we are thinking of making the query builder more driver-specific, so that each driver could potentially support more specific things. This will likely happen in v5.

The base one will still exist for others to extend if they want so of course, this will still be useful.

ccakes:
This *should* be fully backwards compatible. If it's not, I'll make changes so that it is 👍 

> Just so you know, we are thinking of making the query builder more driver-specific, so that each driver could potentially support more specific things. This will likely happen in v5.

Interesting.. that makes sense. Looking forward to seeing what's coming!

stephenafamo:
I just noticed that this will not be backward compatible since the function signature of `AppendWith` changed.

In its current state, we will only be able to merge it into v5.

ccakes:
At some point over the next few days I'll see if I can come up with a change here to resolve this, will comment back either way

sn0rk64:
Come on merge it please!

PR #1088, opened 2022-02-15
Custom template functions for library users

ccakes:
sqlboiler users who call it as a library rather than an executable can now add a `template.FuncMap` to their config objects which will be available in the templates under UserFuncs.

Allows advanced users to extend their templates further without needing to fork sqlboiler itself.

### DIFF
--- a/boilingcore/boilingcore.go
+++ b/boilingcore/boilingcore.go
@@ -156,6 +156,7 @@ func (s *State) Run() error {
 		LQ:                strmangle.QuoteCharacter(s.Dialect.LQ),
 		RQ:                strmangle.QuoteCharacter(s.Dialect.RQ),
 		OutputDirDepth:    s.Config.OutputDirDepth(),
+		UserFuncs:         s.Config.TemplateUserFuncs,
 
 		DBTypes:     make(once),
 		StringFuncs: templateStringMappers,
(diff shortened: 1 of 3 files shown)


boilingcore/boilingcore.go
```diff
@@ -156,6 +156,7 @@ func (s *State) Run() error {
 		LQ:                strmangle.QuoteCharacter(s.Dialect.LQ),
 		RQ:                strmangle.QuoteCharacter(s.Dialect.RQ),
 		OutputDirDepth:    s.Config.OutputDirDepth(),
+		UserFuncs:         s.Config.TemplateUserFuncs,
```
stephenafamo, line 159: I think we should call this `CustomFuncs` or `AdditionalFuncs` or something else.

Because aside from users, sqlboiler can also be used as a base for other executables that want to generate code from a DB schema. e.g. [boilingfactory](https://github.com/stephenafamo/boilingfactory)

stephenafamo (changes requested):

stephenafamo:
It has also occurred to me that adding the functions to the template data directly would not work.

Only predefined templates functions or struct methods can be called. A `template.FuncMap` does not work because it is a map.

stephenafamo:
Copied your commits to #1096 and made some changes on top. I'll close this in favor of that.

Thanks for your work.

PR #1091, opened 2022-02-16
feat: added stringer implementation to where helpers

ecasilla:
This change helps when your doing raw queries but still want to use the boiler generated code for compile time check on you join clauses, where filters, ordering etc..

### DIFF
--- a/templates/main/00_struct.go.tpl
+++ b/templates/main/00_struct.go.tpl
@@ -57,6 +57,7 @@ var {{$alias.UpSingular}}TableColumns = struct {
 	{{- if (oncePut $.DBTypes .Type)}}
 		{{$name := printf "whereHelper%s" (goVarname .Type)}}
 type {{$name}} struct { field string }
+func (w {{$name}}) String() string { return w.field }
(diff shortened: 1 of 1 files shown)


stephenafamo:
This doesn't seem very helpful. The `String()` method just returns `field` which would just be the column name and we already generate types for that (e.g `models.PilotColumns` and `models.PilotTableColumns`).

ecasilla:
But those fields do not include the schema which is the only difference. In postgres if your schema is not public you should include the schema as a prefix to your table to ensure there's no hiccups so when doing queries I could use the TableColumns but that would mean doing something like `fmt.Sprintf("customschema.%s", models.PilotTableColumns.Column)` which gets old rather quickly. I already have to do this for table names

stephenafamo:
Then perhaps we should find a way to have constants that include the schema name. Because adding it to the where helpers do not feel like the right place.

stephenafamo:
I'll close this PR. I understand the use case, but I don't quite agree with this implementation.  
If you have a different approach in mind I'll love to review it.

PR #1102, opened 2022-03-10
Make the join query unique

sadayuki-matsuno:
Due to db specifications, only one join with the same alias can be specified for the same table.

Therefore, when a query with the same join is specified, we have added the ability to pass through one of the joins.

By adding this, we are happy to be able to specify a join without db error even if the query is dynamically created in multiple places.

### DIFF
--- a/queries/query.go
+++ b/queries/query.go
@@ -303,21 +303,49 @@ func SetFrom(q *Query, from ...string) {
 
 // AppendInnerJoin on the query.
 func AppendInnerJoin(q *Query, clause string, args ...interface{}) {
+	if len(args) == 0 {
+		for _, j := range q.joins {
+			if j.clause == clause && j.kind == JoinInner {
+				return
+			}
+		}
+	}
 	q.joins = append(q.joins, join{clause: clause, kind: JoinInner, args: args})
 }
 
 // AppendLeftOuterJoin on the query.
 func AppendLeftOuterJoin(q *Query, clause string, args ...interface{}) {
+	if len(args) == 0 {
+		for _, j := range q.joins {
+			if j.clause == clause && j.kind == JoinOuterLeft {
+				return
+			}
+		}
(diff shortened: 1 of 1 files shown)


stephenafamo:
But an alias can be added when adding the join clause right?  
For example:

```go
qm.InnerJoin("pilots p1 on p1.id = jets.pilot_id"),
qm.InnerJoin("pilots p2 on p2.id = jets.pilot_id"),
```

I'm a little unsure why this is needed. Can you explain what the use case is?

sadayuki-matsuno:
> But an alias can be added when adding the join clause right?

Yes

> I'm a little unsure why this is needed. Can you explain what the use case is?

When I am building a web system, I sometimes want to dynamically change the query of a db based on requests from the frontend.

For example, suppose we have the following table

```
CREATE TABLE orgs(
        name character varying(50) NOT NULL
);

CREATE TABLE users(
        org_id bigint NOT NULL REFERENCES orgs(id),
        name character varying(100) NOT NULL
);
```

When retrieving a list of users, if we want to filter by a particular orgs, we need to join users and orgs.
Since the architecture is implemented in such a way that this join is applied in multiple places, it is possible to write a duplicate join in rare cases, resulting in a DB error.

We would like to be able to go through duplicate joins as described in this PR so that we can write joins without thinking about where we are executing the join and without generating a db error.

stephenafamo:
I'm not *totally* on board with this, because for each join added, we add an extra loop through all the current joins. 

Also, why are you only comparing when `len(args) == 0`?

sadayuki-matsuno:
Since args are  interface type, I found it difficult to compare all of this.

stephenafamo:
After some thought, I would rather not merge this. The additional loop, for every join, for every user of SQLBoiler does not seem worth it for this use case.

I'm sorry.

Issue #1101, opened 2022-03-03
`qm.WithDeleted()` doesn't work as expected


ktakenaka:
If you're having a generation problem please answer these questions before submitting your issue. Thanks!

### What version of SQLBoiler are you using (`sqlboiler --version`)?
SQLBoiler v4.8.6

### What is your database and version (eg. Postgresql 10)
MySQL 5.7

### If this happened at runtime what code produced the issue? (if not applicable leave blank)
```golang
post, err := models.Posts().One(ctx, db)
post.Comments(qm.WithDeleted()).All(ctx, db) // This doesn't omit "deleted_at is not null"
```

**↓This is the generated code**
<details>
<summary>generated code</summary>

```golang
// Comments retrieves all the comment's Comments with an executor.
func (o *Post) Comments(mods ...qm.QueryMod) commentQuery {
	var queryMods []qm.QueryMod
	if len(mods) != 0 {
		queryMods = append(queryMods, mods...)
	}

	queryMods = append(queryMods,
		qm.Where("`comments`.`post_id`=?", o.ID),
		qmhelper.WhereIsNull("`comments`.`deleted_at`"), // <-- 1st
	)

	query := Comments(queryMods...)
	queries.SetFrom(query.Query, "`comments`")

	if len(queries.GetSelect(query.Query)) == 0 {
		queries.SetSelect(query.Query, []string{"`comments`.*"})
	}

	return query
}

// Comments retrieves all the records using an executor.
func Comments(mods ...qm.QueryMod) commentQuery {
	mods = append(mods, qm.From("`comments`"), qmhelper.WhereIsNull("`comments`.`deleted_at`")) // <-- 2nd
	return commentQuery{NewQuery(mods...)}
}
```

</details>

### Please provide a relevant database schema so we can replicate your issue (Provided you are comfortable sharing this)
```sql
CREATE TABLE posts (
  id INTEGER PRIMARY KEY
);

CREATE TABLE comments (
  id INTEGER PRIMARY KEY,
  post_id INTEGER NOT NULL,
  deleted_at TIMESTAMP NULL,
  CONSTRAINT `fk_posts` FOREIGN KEY(post_id) REFERENCES posts(id)
);
```

### Further information. What did you do, what did you expect?
- My investigation
  - `qmhelper.WhereIsNull("comments.deleted_at")` is added twice in `post.Comments()` (*I added comments above generated cd)
    - 1st: In `(o *Post) Comments`
[...]

PR #1103, opened 2022-03-13
Avoid duplicated `deleted_at IS NULL` clause in relationship_to_many

ktakenaka:
# Background of this PR
`qm.WithDeleted()` doesn't work as expected for `to_many` relationship
https://github.com/volatiletech/sqlboiler/issues/1101

# What I did in this PR
Remove `qmhelper.WhereIsNull` from `06_relationship_to_many` template.
Even if we remove it, `qmhelper.WhereIsNull` is added in [here](https://github.com/volatiletech/sqlboiler/blob/60b2e37ce968fa3f72fdc28833fd7b7b6af18546/templates/main/06_relationship_to_many.go.tpl#L32)

---
Confirmed on local
```golang
func main() {
	post := models.Post{ID: 1}
	fmt.Println("post.Comments()")
	post.Comments().All(ctx, db)

	fmt.Println("post.Comments(qm.WithDeleted())")
	post.Comments(qm.WithDeleted()).All(ctx, db)
}
```
```
$ go run main.go
post.Comments()
SELECT `comments`.* FROM `comments` WHERE (`comments`.`post_id`=?) AND (`comments`.`deleted_at` is null);
[1]
post.Comments(qm.WithDeleted())
SELECT `comments`.* FROM `comments` WHERE (`comments`.`post_id`=?);
[1]
```

### DIFF
--- a/templates/main/06_relationship_to_many.go.tpl
+++ b/templates/main/06_relationship_to_many.go.tpl
@@ -23,9 +23,6 @@ func (o *{{$ltable.UpSingular}}) {{$relAlias.Local}}(mods ...qm.QueryMod) {{$fta
 		{{else -}}
 	queryMods = append(queryMods,
 		qm.Where("{{$schemaForeignTable}}.{{$rel.ForeignColumn | $.Quotes}}=?", o.{{$ltable.Column $rel.Column}}),
-		{{if and $.AddSoftDeletes $canSoftDelete -}}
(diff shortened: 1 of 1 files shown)


stephenafamo:
Aside from the `to_many` relationships, this seems to also affect `to_one` and `one_to_one` queries. Thanks for pointing this out, I'll fix those too

PR #1118, opened 2022-04-13
parsePoint(s) regexp.MustCompile is now called only once

ivokanchev:
Scanning a GeoPoint used to transitively call `regexp.MustCompile`. This leads to unnecessary memory allocations and therefore substantial GC runs which leads to high CPU usage. 

This PR extracts regular expression compile step to be performed only once. Performance is improved as the benchmarks show.

Screenshot 2022-04-13 at 11 56 52


stephenafamo:
This is very welcome. Thanks for catching it.

Issue #994, opened 2021-09-10
pgeo.Path does not allow paths with less than 2 points.


despondency:
The problem is the following:

We are using the geometric type Path from PostgreSQL, to persist a bunch of points on a path. However if we have Point(A) == Point(B) the path between them would be a single Point(A==B).

We insert the data in the DB where we have a path with one point, however when we try to get it, we get the wrong path exception. Why is this assumption made on ORM level? Why does sqlboiler allows us to persist one point on a path, but never to get it? 

Much thanks for the assistance.

The code path that contains the problem is this:

pgeo.Path 

```
(*p).Points, err = parsePoints(val)
	if err != nil {
	return err
}

if len((*p).Points) < 2 {
     return errors.New("wrong path")
}
```

aarondl:
I actually have no idea why this is like this as I had nothing to do with the implementation of it and would rely on an expert in pgeo to tell us why that might be the case.

saulortega:
I wrote that code, and I assumed that a path should have at least two points, but you are right, it can have one single point.

Fixed in https://github.com/volatiletech/sqlboiler/pull/1123

stephenafamo:
Released in `v4.11.0`

stephenafamo:
Thanks for the contribution @saulortega

PR #1123, opened 2022-04-19
Fix one single point in paths and polygons

saulortega:
By checking https://github.com/volatiletech/sqlboiler/issues/994 I confirm that Postgres allow paths and polygons consistent of one single point. So the previous assumption that it should be at least two points was wrong.

Also two regexp are now called once.


stephenafamo:
Thank you. Great work 👍🏾

PR #1143, opened 2022-06-03
Fetch column and table info in parallel

pavel-krush:
Hello.

I have a large set of tables for sqlboiler(>50 and counting).
It takes more than 1 minute to generate code for all tables. Most of the time is spent getting tables and columns info.
After reviewing the sources, I realised that fetching this data can be done in parallel safely.

This PR reduces time needs from 1 min to 5 sec with default setting of 10 threads

PS: need help with overriding default amount of threads via config or command line


stephenafamo:
This is very nice!!!!  
The settings are passed through environment variables to the driver as `drivers.Config{}` which is based on `map[string]interface{}`

Aside from some preset values [here](https://github.com/volatiletech/sqlboiler/blob/89d4a67ae24ee3857a73cfbe987e92837d390b73/main.go#L202), the rest are passed transparently for all configuration parameters that have the prefix of the driver name.

pavel-krush:
sorry, i was a little busy. i added ability to define concurrency parameter in drivers configuration file and set it's default value to 10

stephenafamo:
From my first look, it doesn't look as though the parameters are actually configurable, because reading the user's configuration happens in `main.go` and is then passed down to the driver and I don't see any changes to `main.go`.

Also, I don't understand why there is a need to configure default concurrency.   
Wouldn't the default be whatever we set it to and then the user configures whatever they also want?

pavel-krush:
@stephenafamo as you mentioned before main.go reads config file as map[string]interface{} and passes it to the driver. drivers can then parse any parameters as they want. so we do not need change main.go, only drivers parsers.
I tested sqlboiler on 55 tables with a poor internet connection with the following configs:

1. manually set low amount of threads
```
[psql]
dbname    = "xxx"
host      = "10.238.44.38"
user      = "xxx"
pass      = "xxx"
sslmode   = "disable"
concurrency = 1
```
generation time: 1m 27 s

2. set high amount of threads
```
[psql]
dbname    = "xxx"
host      = "10.238.44.38"
user      = "xxx"
pass      = "xxx"
sslmode   = "disable"
concurrency = 20
```
generation time: 8 s

3. default amount of threads
```
[psql]
dbname    = "xxx"
host      = "10.238.44.38"
user      = "xxx"
pass      = "xxx"
sslmode   = "disable"
```
generation time: 12s

stephenafamo:
You're right. Sorry 🤦🏾‍♂️

Can you also explain why it is possible to configure the default? Isn't it enough to set the desired concurrency?

pavel-krush:
The default value is just a constant. It's not possible to configure it. it's possible to override the default value by the value given by the user in configuration file.

pavel-krush:
@stephenafamo fixed, also fixed tests

stephenafamo:
It just occurred to me that by changing the exported `Tables()` function signature, the change isn't backwards compatible, and may break any 3rd party driver (e.g. the cockroachDB driver) that depends on it.

Instead we should create a new function that takes a compulsory concurrency argument and make the current function call the new function with the default concurrency

pavel-krush:
Yes, backward compatibility is really important here for third party drivers. So i made a new function TablesConcurrently and redirected all out drivers to that function. Also I set 1 thread for drivers that use Tables() in case if they are not able to do that job in parallel. They can support concurrency setting and then start sing TablesConcurrently()

stephenafamo:
Nice work 👍🏾

drivers/interface.go
```diff
@@ -182,44 +232,71 @@ func views(c ViewConstructor, schema string, whitelist, blacklist []string) ([]T
...
-			Name:   name,
-		}
+	ret := make([]Table, len(names))
+
+	limiter := newConcurrencyLimiter(concurrency)
+	wg := sync.WaitGroup{}
+	errs := make(chan error, len(names))
+	for i, name := range names {
+		limiter.get()
+		wg.Add(1)
```
stephenafamo, line 242: Very minor change, but I feel like maybe `limiter.get()` should come after `wg.Add(1)`.

While very unlikely, it feels like it is possible that `wg.Wait()` unblocks sometime between when one of the workers finish and another begins.

pavel-krush: This cannot happen here, because limiter.get is a blocking operation when there are no tickets available and wg.Wait() is called after all Add().
But you are right, wg.Add() should be called first just to be consistent with the cases where such situations are possible.

drivers/interface.go
```diff
@@ -102,77 +107,122 @@ type TableColumnTypeTranslator interface {
...
-		filterPrimaryKey(&t, whitelist, blacklist)
-		filterForeignKeys(&t, whitelist, blacklist)
+	ret := make([]Table, len(names))
+
+	limiter := newConcurrencyLimiter(concurrency)
+	wg := sync.WaitGroup{}
+	errs := make(chan error, len(names))
+	for i, name := range names {
+		limiter.get()
+		wg.Add(1)
```
stephenafamo, line 152: Another instance here 😅

PR #1145, opened 2022-06-06
Generate IN/NIN methods for enum type query helpers [#1142]

### LINKED ISSUE #1142: No `IN` and `NIN` helpers for enum types

SQLBoiler v4.11.0

MySQL 5.7

```
sqlboiler --wipe --add-enum-types mysql
```

Was expecting to get the `IN` and `NIN` helper methods for where clauses, similar to primitive types. For some reason, these are missing for enum types.

Could these be enabled for enums as well, or is there a reason against it? We have lots of enums in the DB that represent different states, so we often select over a range of applicable states and not having these helper function is a bit of a _dealbreaker_ for using `--add-enum-types`.

optiman:
Resolves #1142.
Enum regexp was copied from `strmangle` package.

Tested only with CockroachDB, but should work the same for MySQL and PostgreSQL as all other enum-related stuff.

### DIFF
--- a/drivers/table.go
+++ b/drivers/table.go
@@ -92,7 +92,7 @@ func (t Table) CanSoftDelete(deleteColumn string) bool {
 func TablesHaveNullableEnums(tables []Table) bool {
 	for _, table := range tables {
 		for _, col := range table.Columns {
-			if col.Nullable && rgxEnum.MatchString(col.DBType) {
+			if col.Nullable && IsEnumDBType(col.DBType) {
 				return true
 			}
 		}
--- a/drivers/table.go
+++ b/drivers/table.go
@@ -2,7 +2,6 @@ package drivers
 
 import (
 	"fmt"
-	"strings"
 )
 
 // Table metadata from the database schema.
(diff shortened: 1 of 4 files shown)


drivers/table.go
```diff
@@ -93,9 +92,7 @@ func (t Table) CanSoftDelete(deleteColumn string) bool {
 func TablesHaveNullableEnums(tables []Table) bool {
 	for _, table := range tables {
 		for _, col := range table.Columns {
-			if col.Nullable &&
-				(strings.HasPrefix(col.DBType, "enum.") || // postgresql
-					strings.HasPrefix(col.DBType, "enum(")) { // mysql
+			if col.Nullable && rgxEnum.MatchString(col.DBType) {
```
stephenafamo, line 95: You can use your freshly added `IsEnumDBType` here right? That way it is easier to refactor in the future if necessary?

stephenafamo (changes requested):

drivers/table.go
optiman: Yes, it can be used. Done.

stephenafamo (approved):

Issue #1147, opened 2022-06-09
Inconsistent Results from Bind


reiman2222:
sqlboil cli version: `SQLBoiler v4.8.6`

Go version: `go version go1.17.10 darwin/amd64`

go.mod: 
```
go 1.17

require (
	github.com/DATA-DOG/go-sqlmock v1.5.0
	github.com/volatiletech/sqlboiler/v4 v4.11.0
)
```

Test to reproduce issue: 
```
// was generated by sqlboiler
var dialect = drivers.Dialect{
	LQ: 0x22,
	RQ: 0x22,

	UseIndexPlaceholders:    true,
	UseLastInsertID:         false,
	UseSchema:               false,
	UseDefaultKeyword:       true,
	UseAutoColumns:          false,
	UseTopClause:            false,
	UseOutputClause:         false,
	UseCaseWhenExistsClause: false,
}

// NewQuery initializes a new Query using the passed in QueryMods
// was generated by sqlboiler
func NewQuery(mods ...qm.QueryMod) *queries.Query {
	q := &queries.Query{}
	queries.SetDialect(q, &dialect)
	qm.Apply(q, mods...)

	return q
}

type Apple struct {
	UUID string `boil:"uuid" json:"uuid" toml:"uuid" yaml:"uuid"`
}

type Banana struct {
	UUID      string `boil:"uuid" json:"uuid" toml:"uuid" yaml:"uuid"`
	AppleUUID string `boil:"apple_uuid" json:"apple_uuid" toml:"apple_uuid" yaml:"apple_uuid"`
}

type Coconut struct {
	Apple  `boil:"apple,bind"`
	Banana `boil:"b,bind"`
}

func Test(t *testing.T) {
	db, mock, _ := sqlmock.New()
	defer db.Close()

	mockRow := mock.NewRows([]string{"uuid", "uuid", "apple_uuid"}).AddRow(
		"apple",
		"banana",
		"apple",
	)

	mock.ExpectQuery(".*").WillReturnRows(mockRow)

	query := []qm.QueryMod{
		qm.Select("apple.*", "b.*"),
		qm.From("apple"),
		qm.InnerJoin("banana b on b.apple_uuid = apple.uuid"),
	}

	var data []Coconut
	err := NewQuery(query...).Bind(context.Background(), db, &data)
	assert.Nil(t, err)
	assert.Equal(t, 1, len(data))

	fmt.Printf("Bind Result: %+v\n", data)

[...]

PR #1148, opened 2022-06-09
Clairfy what Bind does when boil tags are duplicated

reiman2222:
See: https://github.com/volatiletech/sqlboiler/issues/1147

### DIFF
--- a/queries/reflect.go
+++ b/queries/reflect.go
@@ -77,8 +77,9 @@ func (q *Query) BindG(ctx context.Context, obj interface{}) error {
 //   - If the ",bind" option is specified on a struct field and that field
 //     is a struct itself, it will be recursed into to look for fields for
 //     binding.
-//   - If one or more boil struct tags are duplicated the behaviour of Bind
-//     is undefined.
+//   - If one or more boil struct tags are duplicated and there are multiple
+//     matching columns for those tags the behaviour of Bind will undefined
+//     for those fields with duplicated struct tags.
 //
 // Example usage:
 //
--- a/queries/reflect.go
+++ b/queries/reflect.go
@@ -78,7 +78,7 @@ func (q *Query) BindG(ctx context.Context, obj interface{}) error {
 //     is a struct itself, it will be recursed into to look for fields for
 //     binding.
 //   - If one or more boil struct tags are duplicated and there are multiple
-//     matching columns for those tags the behaviour of Bind will undefined
+//     matching columns for those tags the behaviour of Bind will be undefined
 //     for those fields with duplicated struct tags.
 //
 // Example usage:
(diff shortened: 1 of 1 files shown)


stephenafamo:
I don't think the statement is actually correct.

In the example in your issue, the problem is that you asked boil to bind `apple.uuid` but there is no columned named as such.

reiman2222:
> I don't think the statement is actually correct.
> 
> In the example in your issue, the problem is that you asked boil to bind `apple.uuid` but there is no columned named as such.

@stephenafamo This is case regardless of the prefix but let me provided an updated example:

``` go
// was generated by sqlboiler
var dialect = drivers.Dialect{
	LQ: 0x22,
	RQ: 0x22,

	UseIndexPlaceholders:    true,
	UseLastInsertID:         false,
	UseSchema:               false,
	UseDefaultKeyword:       true,
	UseAutoColumns:          false,
	UseTopClause:            false,
	UseOutputClause:         false,
	UseCaseWhenExistsClause: false,
}

// NewQuery initializes a new Query using the passed in QueryMods
// was generated by sqlboiler
func NewQuery(mods ...qm.QueryMod) *queries.Query {
	q := &queries.Query{}
	queries.SetDialect(q, &dialect)
	qm.Apply(q, mods...)

	return q
}

type Apple struct {
	UUID string `boil:"uuid" json:"uuid" toml:"uuid" yaml:"uuid"`
}

type Banana struct {
	UUID      string `boil:"uuid" json:"uuid" toml:"uuid" yaml:"uuid"`
	AppleUUID string `boil:"apple_uuid" json:"apple_uuid" toml:"apple_uuid" yaml:"apple_uuid"`
}

type Coconut struct {
	Apple  `boil:",bind"`
	Banana `boil:",bind"`
}

func Test(t *testing.T) {
	db, mock, _ := sqlmock.New()
	defer db.Close()

	mockRow := mock.NewRows([]string{"uuid", "uuid", "apple_uuid"}).AddRow(
		"apple",
		"banana",
		"apple",
	)

	mock.ExpectQuery(".*").WillReturnRows(mockRow)

	query := []qm.QueryMod{
		qm.Select("apple.*", "b.*"),
		qm.From("apple"),
		qm.InnerJoin("banana b on b.apple_uuid = apple.uuid"),
	}

	var data []Coconut
	err := NewQuery(query...).Bind(context.Background(), db, &data)
	assert.Nil(t, err)
	assert.Equal(t, 1, len(data))

	fmt.Printf("Bind Result: %+v\n", data)

	assert.Equal(t, "apple", data[0].Apple.UUID)
	assert.Equal(t, "banana", data[0].Banana.UUID)
	assert.Equal(t, "apple", data[0].Banana.AppleUUID)
}
```

In this example I've updated the `Coconut` struct to have no prefixes. If you execute this example text you will find that it sporadically flips between passing and failing.

Having this behavior detailed in the docs would have saved me a lot of time in debugging. Im hoping this will also help others should they erroneously construct the same kind of join.

stephenafamo:
In your new example, you have duplicated column names in your SQL result, and duplicated matching values in the struct.  
Perhaps in your PR, you could better clarify that this is also because there are multiple matching result columns.

The whole thing is kinda weird, just the way the SQL results are returned should make it sort of clear that there would be a problem.

PR #1151, opened 2022-06-17
Improve psql performance by isolating uniqueness query.

peterldowns:
Hi, we here at Pipe have been huge fans of sqlboiler for years now and I am excited to be able to (hopefully) contribute something back. We've noticed that generating models has grown extremely slow for us -- it takes about a minute to generate around 190 different `<model>.go` files. We re-generate models quite frequently while developing, which means that this has been a particular pain point for us. I finally got around to doing some performance investigation, and found that we could speed things up so that it would only take about 5 seconds!

This PR speeds the psql/postgres driver up by moving the uniqueness query (used to determine if a given column is unique) from happening once-per-table-per-run to happening once-per-run, and caching the results on the driver.

Unlike previously, the code will query for the uniqueness status of _all_ columns across _all_ tables across _all_ schemas, not just the ones that are being generated. I don't expect this to cause any issues but if this is a barrier to merging it, it shouldn't be too hard to change it to only query against specific schemas/tables.

Based on testing locally, I believe this is the most significant possible speedup for the postgres driver. Some other small improvements that I tried that were effective, but not so dramatically:
* Running the `Columns()` method in parallel for multiple tables at once (observed change: 60 seconds -> 40 seconds)
  * Also increasing the max open / max idle connections on the connection used by the driver
* Fixing `o(n^2)` behavior in column/table reference comparisons (observed change: 40 seconds -> 35 seconds)
  * There are a lot of places where lists are iterated over and over again, which could ~simply be replaced by a map
* Other sql query changes to avoid re-querying column information multiple times (not tested, but given how much this query change improved things, I would expect there to be other wins available)

### DIFF
--- a/drivers/sqlboiler-psql/driver/psql.go
+++ b/drivers/sqlboiler-psql/driver/psql.go
@@ -44,6 +44,14 @@ type PostgresDriver struct {
 	version        int
 	addEnumTypes   bool
 	enumNullPrefix string
+
+	uniqueColumns map[columnIdentifier]struct{}
+}
+
+type columnIdentifier struct {
+	Schema string
+	Table  string
+	Column string
 }
 
 // Templates that should be added/overridden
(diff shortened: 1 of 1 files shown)


stephenafamo:
This is very nice. Thank you! All the tests are passing, so I'll go ahead and merge this.

Once #1143 is also completed and merged, then there would be a significant speed boost overall.

Issue #54, opened 2016-09-30
Generate types for views?


SlinSo:
Will you add a feature to generate types for views in the future?
Currently it is not possible, because Views don't have primary keys.

aarondl:
So, I don't have an answer that's positive or negative yet. I'd like to discuss what it would mean.

Out of all of the things the models package currently provides you would be left with the read portions only.
* Query mod queries with `.One()` `.All()` `Count()` and `Exists()` (which you still have anyway through `models.NewQuery()`)
* Struct gen
* Hooks

The following things wouldn't be generated due to needing a primary key
* UpdateAll
* Update
* Insert
* Upsert
* Delete
* FindX
* XExists
* Reload
* ReloadAll

Is the above struct generation and hooks then worth it to implement this feature?

nullbio:
Closing due to inactivity.

expbbc:
I'd like to re-open this issue if possible. Not having support for views makes it impossible for me to take advantage of sqlboiler. The entirety of my application interacts with the database through stored procedures and views. The underlying table topologies are ultimately irrelevant.

Furthermore, there is no good reason that I can see why INSERT, UPDATE, UPSERT et al have a hard requirement on a primary key. If this is an internal requirement of sqlboiler then the architecture chosen is not able to represent actual databases and is thus faulty. These Postgres views are not read only and can be INSERTed into despite their lack of a primary key, constraint violations are dealt with via exceptions, not prematurely.

If there is any more information or justification that I can provide please let me know, but as it stands although sqlboiler seems an excellent solution, i cannot adopt it.

aarondl:
The requirement of having a primary key of some kind is mostly because a lot of functionality all relies on the fact that we can find a single given row. You can see the list above, I won't repeat it here.

In the case of Upsert/Update/Delete/Find etc. how do you delete a row that you cannot reference? How do you find the row to update for example. A primary key is the only good way to refer to a single given row, and so all these operations that work on single rows -do- have a hard requirement on having a primary key as without one there is no good mechanism to find that single row (that I know of, I'm happy to be corrected).

[...]

PR #1154, opened 2022-06-22
Support for psql materialized view

severedsea:
Related to https://github.com/volatiletech/sqlboiler/pull/1066
Addresses https://github.com/volatiletech/sqlboiler/issues/54#issuecomment-496046557 partially

## Description
Added psql support to generate models for materialized views. The implementation is built on top of the support view PR (https://github.com/volatiletech/sqlboiler/pull/1066)

## Implementation
In relation to views, have identified 3 methods that can be appended with Materialized Views support
- ViewNames
- ViewCapabilities
- ViewColumns

### ViewNames
Added a SELECT query to retrieve materialized view names from [pg_matviews](https://www.postgresql.org/docs/current/view-pg-matviews.html)

### ViewCapabilities
Since materialized views are read-only, and the only way to "update" the data is by executing [REFRESH MATERIALIZED VIEW](https://www.postgresql.org/docs/current/sql-refreshmaterializedview.html), default the capabilities to all `FALSE`

### ViewColumns
Current implementation is reusing the `Columns()` method to retrieve the metadata. This is mainly because views metadata can be retrieved just like a table. 

By applying similar principle, crafted a query to retrieve the necessary column metadata for a materialized view mainly from the catalog tables (i.e. pg_attributes, pg_class, etc), and explicit looking for [`relkind = 'm'`](https://www.postgresql.org/docs/current/catalog-pg-class.html). 

Patterned the SELECT based on the existing `Columns` query. 

Notes on testing efforts:
* Created a view from `type_monsters` in `psql/testdatabase.sql`
* Created a materialized view from the view (i.e. CREATE MATERIALIZED VIEW type_monsters_mv AS SELECT * FROM type_monsters_v) 
* Compared the results of existing `Columns` query and result of materialized view columns query for the view and materialized view respectively.
* Same [output](https://github.com/volatiletech/sqlboiler/files/8954813/columns_mv.csv)
* Executed `./boil.sh gen psql`, did a file-comparison between the generated model `TypeMonstersV` vs `TypeMonstersMV` (See diff here: https://www.diffchecker.com/iK09iRuf)


stephenafamo:
The MR looks good, but it is causing the tests to fail. Can you try to figure out why? 🙏🏾

Issue #1158, opened 2022-07-05
qm.Load() does not work when using .Bind() with an embedded struct


evanrolfe:
SqlBoiler seems incapable of using `.Load()` to eager load a relationship when using `.Bind()` along with an embedded struct i.e. `PetWithStats` in the code below.

#### Versions
`SQLBoiler v4.11.0` and `mysql:5.7`

#### Code
```go
package main

import (
	"context"
	"fmt"

	"github.com/evanrolfe/sqlboiler_bug/internal/models"
	_ "github.com/go-sql-driver/mysql"
	"github.com/jmoiron/sqlx"
	"github.com/volatiletech/sqlboiler/boil"
	"github.com/volatiletech/sqlboiler/v4/queries/qm"
)

type PetWithStats struct {
	models.Pet      `boil:",bind"`
	NumberOfFriends int64 `boil:"number_of_friends"`
	NumberOfEnemies int64 `boil:"number_of_enemies"`
}

func main() {
	db, err := sqlx.Connect("mysql", "...")
	if err != nil {
		fmt.Printf("failed connecting to the database: %w\n", err)
		return
	}
	ctx := context.Background()

	query := []qm.QueryMod{
		qm.Select("p.*"),
		qm.Select("(SELECT COUNT(*) FROM pets p2 WHERE p2.animal_id <> p.animal_id) AS number_of_friends"),
		qm.Select("(SELECT COUNT(*) FROM pets p2 WHERE p2.animal_id = p.animal_id) AS number_of_enemies"),
		qm.From(models.TableNames.Pets + " as p"),
		qm.Load(models.PetRels.Animal),
	}

	pets := []*PetWithStats{}
	err = models.NewQuery(query...).Bind(ctx, db, &pets)
	if err != nil {
		fmt.Printf("failed query: %w\n", err)
		return
	}

	for _, pet := range pets {
		fmt.Println(pet.ID, " ", pet.Name, "(", pet.R.Animal.Name, ")")
	}
}
```

#### Database Schema
```sql 
CREATE TABLE animals (
     id MEDIUMINT NOT NULL AUTO_INCREMENT,
     name CHAR(30) NOT NULL,
     PRIMARY KEY (id)
);

CREATE TABLE pets (
     id MEDIUMINT NOT NULL AUTO_INCREMENT,
     name CHAR(30) NOT NULL,
     animal_id MEDIUMINT NOT NULL,
     PRIMARY KEY (id),
[...]

PR #1160, opened 2022-07-09
Loading model relationships when binding to a struct with embedded model

optiman:
This PR addresses the issue described in #1158.
In case when `Bind()` is used with a struct that contains embedded model field, model's `Load*` methods will panic because types do not match.
Only in this case the reflection is used to find the embedded field(s) of the same type as model and save it's pointer(s) to the model `object`/`slice` already allocated in `Load*` methods.

It's not the most elegant solution, but it works without breaking backward compatibility and doesn't require to rewrite the query builder.

### DIFF
--- a/queries/eager_load.go
+++ b/queries/eager_load.go
@@ -309,3 +309,101 @@ var (
 	applicatorSentinel    Applicator
 	applicatorSentinelVal = reflect.ValueOf(&applicatorSentinel).Elem()
 )
+
+// SetFromEmbeddedStruct sets `to` value from embedded struct
+// of the `from` struct or slice of structs.
+// Expects `to` and `from` to be a pair of pre-allocated **struct or *[]*struct.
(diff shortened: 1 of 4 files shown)


stephenafamo:
At first look, this seems to solve the issue. I'll find some time later to review again before merging.

Thank you so much 💯

stephenafamo:
@evanrolfe It's quite an extensive PR, so I wouldn't want to just merge it without a closer look. The other people who would be able to merge it have even less time to review it at the moment 😅

In the meantime perhaps you can use a replace directive to use this fork. It'll also act as a real world test of wether it works as expected.

evanrolfe:
@stephenafamo of course, no problem. I've checked out this branch locally and can confirm that this does fix the problem from https://github.com/volatiletech/sqlboiler/issues/1158.

PR #1165, opened 2022-07-26
Allocate struct pointers fields with "bind" tag

### LINKED ISSUE #1164: Call of `reflect.Value.Field` on zero `Value`
```
SQLBoiler v3.7.1
```

Postgres 10.6

```go
var u struct {
	User *models.User `boil:",bind"`
}
err = models.Users().Bind(nil, db, &u)
if err != nil {
	panic(err)
}
```

```sql
CREATE TABLE "user" (
  id SERIAL PRIMARY KEY
);
```

I expected the `u` variable to be filled but instead I get this panic:
```
panic: reflect: call of reflect.Value.Field on zero Value

goroutine 1 [running]:
reflect.Value.Field({0x0?, 0x0?, 0xc000143bc0?}, 0x44b9f2?)
        /usr/lib/go/src/reflect/value.go:1220 +0xef
github.com/volatiletech/sqlboiler/queries.ptrFromMapping({0x7063e0?, 0xc0000b0de0?, 0x7d7938?}, 0xff0000, 0x1)
        /home/elia/go/pkg/mod/github.com/volatiletech/sqlboiler@v3.7.1+incompatible/queries/reflect.go:378 +0x129
github.com/volatiletech/sqlboiler/queries.PtrsFromMapping({0x7063e0?, 0xc0000b0de0?, 0xc0000a0480?}, {0xc0000a3270, 0x1, 0x0?})
        /home/elia/go/pkg/mod/github.com/volatiletech/sqlboiler@v3.7.1+incompatible/queries/reflect.go:344 +0x94
github.com/volatiletech/sqlboiler/queries.bind(0x6e5e40?, {0x6e5e40?, 0xc0000b0de0}, {0x7d9360, 0x7063e0}, {0x6ecca0?, 0xc000098e90?}, 0x0)
        /home/elia/go/pkg/mod/github.com/volatiletech/sqlboiler@v3.7.1+incompatible/queries/reflect.go:280 +0x62b
github.com/volatiletech/sqlboiler/queries.(*Query).Bind(0xc0000fb600, {0x0, 0x0}, {0x7d7d98?, 0xc00009fee0}, {0x6e5e40, 0xc0000b0de0})
        /home/elia/go/pkg/mod/github.com/volatiletech/sqlboiler@v3.7.1+incompatible/queries/reflect.go:147 +0x147
main.main()
        /home/elia/dump/sqlboilertest/main.go:18 +0xfa
exit status 2
```

The panic is not caused by the `if err != nil { panic(err) }` but by the call to `.Bind` itself.

When I change the type of the field from `*models.User` to `models.User` the issue disappears. But I'd like to use a pointer for performance reasons.

eliaperantoni:
Hello,
this is an attempt to fix #1164.

If I were to do something like:
```go
var slice []*struct{
    User *models.User `boil:",bind"`
}
models.Users().BindG(nil, &slice)
```

It wouldn't have worked prior to this PR because `bind` would've created a zeroed instance of the struct and then attempted to get a `reflect.Value` by dereferencing the `User` field. But in a zeroed struct the pointer is nil, and so you get this panic:

```
panic: reflect: call of reflect.Value.Field on zero Value
```

A workaround would be to embed `models.User` without a pointer. But this is not always feasible or performant, depending on how the resulting slice is going to be used afterward.

This PR makes it so that `bind` also allocates any pointer field provided that the target is `[]*Struct` and the `bind` tag is set.

---

Implementation wise, I simply enumerate the struct fields after the call to `reflect.New` and allocate any pointer with the `bind` tag.

When the `bind` target is `[]Struct` (instead of `[]*Struct`) the behavior remains unchanged as the structs are copied-by-value anyways and the [documentation](https://pkg.go.dev/github.com/volatiletech/sqlboiler/v4/queries#Bind) already warns about not using structs with references when binding like that.

### DIFF
--- a/queries/reflect.go
+++ b/queries/reflect.go
@@ -251,7 +251,7 @@ Rows:
 		case kindSliceStruct:
 			pointers = PtrsFromMapping(oneStruct, mapping)
 		case kindPtrSliceStruct:
-			newStruct = reflect.New(structType)
+			newStruct = makeStructPtr(structType)
 			pointers = PtrsFromMapping(reflect.Indirect(newStruct), mapping)
 		}
 		if err != nil {
@@ -277,6 +277,27 @@ Rows:
 	}
 
 	return nil
+}
+
+// makeStructPtr takes a struct type and returns a pointer to a new instance of it. This is used by bind to allocate new
+// slice elements when the bound-to variable is []*Struct
+func makeStructPtr(typ reflect.Type) reflect.Value {
+	// Allocate struct
+	val := reflect.New(typ)
+
+	// For all the fields
+	for i := 0; i < typ.NumField(); i++ {
+		field := typ.Field(i)
+		_, recurse := getBoilTag(field)
+
+		// If ",bind" was in the tag and the field is a pointer
+		if recurse && field.Type.Kind() == reflect.Ptr {
+			// Then allocate the field
+			val.Elem().Field(i).Set(reflect.New(field.Type.Elem()))
+		}
+	}
+
+	return val
 }
 
 // BindMapping creates a mapping that helps look up the pointer for the


eliaperantoni:
@stephenafamo

PR #1221, opened 2022-11-26
Prepend schema to generated TableNames values

hopesea:
The absence of the schema name prefix can lead to ambiguity and unexpected results if relying on default schema or search path behaviour of the target db.


stephenafamo:
This will be a breaking change, so it will definitely not be merged until at least `v5`.

Also, I'm not entirely sure about the benefits of this. There are many cases where a user will prefer just the table name without the schema prefix.

PR #1258, opened 2023-03-15
Fix qm.WithDeleted helper with a custom soft delete column

### LINKED ISSUE #1256: Using `auto-columns.deleted` config and `qm.WithDeleted()`
SQLBoiler v4.13.0

cockroach v22.2.6

```
sqlboiler --wipe --no-hooks --no-rows-affected --no-tests --add-soft-deletes -p crdb -o crdb crdb
```

```sql
CREATE TABLE IF NOT EXISTS kitties (
    id
        CHAR(26) NOT NULL,
    delete_time
        TIMESTAMPZ NULL
);
```

**sqlboiler.yaml**
```
crdb:
  user: user
  pass: password
  host: localhost
  port: 26257
  dbname: 'app'
  sslmode: disable
auto-columns:
  deleted: 'delete_time'
```

Hello,
the problem that I have is that when I try to generate everything and I use it, the query modifier `qm.WithDeleted()` doesn't remove the part that is `AND ("kitties"."delete_time" is null)` in the SQL query. I checked it with the `boil.DebugMode = true`.

If I use the default `auto-columns.deleted` setting and modify my schema from `delete_time` to `deleted_at`, it works as expected.

I reivewed the code inside the repo and I saw that is because the `deleted_at` is hardcoded in the code instead of read it from the config (this commit is actually the HEAD of master).
https://github.com/volatiletech/sqlboiler/blob/66cb4680047d7801156a080ba69a3f14ef5986fc/queries/query.go#L411 

If the bug exists, any workaround in the meantime?

Thanks

lopezator:
As the `deletedAtRgx` was hardcoded, there wasn't way to use the `qm.WithDeleted` helper when you have a custom soft delete column set in `auto-columns.deleted` config.

This PR fixes that, overwriting the variable when the the custom column is set and the soft deletion capatibilies are activated for that table.

Closes #1256

### DIFF
--- a/templates/main/singleton/boil_queries.go.tpl
+++ b/templates/main/singleton/boil_queries.go.tpl
@@ -12,6 +12,17 @@ var dialect = drivers.Dialect{
 	UseCaseWhenExistsClause: {{.Dialect.UseCaseWhenExistsClause}},
 }
 
+{{- if not .AutoColumns.Deleted }}
+    // This is a dummy variable to prevent unused regexp import error
+    var _ = &regexp.Regexp{}
+{{- end }}
+
+{{- if and (.AutoColumns.Deleted) (ne $.AutoColumns.Deleted "deleted_at") }}
+    func init() {
+        queries.SetRemoveSoftDeleteRgx(regexp.MustCompile("{{$.AutoColumns.Deleted}}[\"'`]? is null"))
+    }
+{{- end }}
+
 // NewQuery initializes a new Query using the passed in QueryMods
 func NewQuery(mods ...qm.QueryMod) *queries.Query {
 	q := &queries.Query{}
--- a/importers/imports.go
+++ b/importers/imports.go
@@ -212,6 +212,9 @@ func NewDefaultImports() Collection {
 
 	col.Singleton = Map{
 		"boil_queries": {
+			Standard: List{
+				`"regexp"`,
+			},
 			ThirdParty: List{
 				`"github.com/volatiletech/sqlboiler/v4/drivers"`,
 				`"github.com/volatiletech/sqlboiler/v4/queries"`,
--- a/queries/query.go
+++ b/queries/query.go
@@ -408,7 +408,13 @@ func RemoveSoftDeleteWhere(q *Query) {
 	q.removeSoftDelete = true
 }
 
-var deletedAtRgx = regexp.MustCompile("deleted_at[\"'`]? is null")
+var removeSoftDeleteRgx = regexp.MustCompile("deleted_at[\"'`]? is null")
+
+// SetRemoveSoftDeleteRgx sets the removeSoftDeleteRgx variable to support
+// qm.Where method to work with a custom soft delete auto-column name.
+func SetRemoveSoftDeleteRgx(rgx *regexp.Regexp) {
+	removeSoftDeleteRgx = rgx
+}
 
 // removeSoftDeleteWhere attempts to remove the soft delete where clause
 // added automatically by sqlboiler.
@@ -419,7 +425,7 @@ func (q *Query) removeSoftDeleteWhere() {
 
 	for i := len(q.where) - 1; i >= 0; i-- {
 		w := q.where[i]
-		if w.kind != whereKindNormal || !deletedAtRgx.MatchString(w.clause) {
+		if w.kind != whereKindNormal || !removeSoftDeleteRgx.MatchString(w.clause) {
 			continue
 		}
 


templates/main/13_all.go.tpl
```diff
@@ -13,6 +13,9 @@ func {{$alias.UpPlural}}(mods ...qm.QueryMod) {{$alias.DownSingular}}Query {
     if len(queries.GetSelect(q)) == 0 {
         queries.SetSelect(q, []string{"{{$schemaTable}}.*"})
     }
+    {{if and .AddSoftDeletes $canSoftDelete .AutoColumns.Deleted -}}
+        queries.SetRemoveSoftDeleteRgx(removeSoftDeleteRgx)
+    {{end}}
```
stephenafamo, line 18: This should be in a singleton template. It only needs to be called once. here it will be called for every table.

templates/main/singleton/boil_queries.go.tpl
```diff
@@ -12,6 +12,13 @@ var dialect = drivers.Dialect{
 	UseCaseWhenExistsClause: {{.Dialect.UseCaseWhenExistsClause}},
 }
 
+{{if not .AutoColumns.Deleted}}
+    // This is a dummy variable to prevent unused regexp import error
+    var _ = &regexp.Regexp{}
+{{else}}
+    var removeSoftDeleteRgx = regexp.MustCompile("{{$.AutoColumns.Deleted}}[\"'`]? is null")
+{{- end}}
```
stephenafamo, line 20: This would be better alongside other global variables which are a little after the model definition.

stephenafamo (changes requested):
Thanks for this. Looks like a good implementation

templates/main/singleton/boil_queries.go.tpl
lopezator: What do you exactly mean by `model`? The table's corresponding struct? Do you mean to have the variable replicated inside each table's generated file? Like here?

https://github.com/volatiletech/sqlboiler/blob/ca0d71ed51bab25289fb52dd217df8a4ffd859dc/templates/main/01_types.go.tpl#L45-L61

Would it make sense to set this globally in an `init()` function inside some singleton template? Create a new one for this use cases?

stephenafamo: > Would it make sense to set this globally in an `init()` function inside some singleton template? Create a new one for this use cases?

I mixed some stuff up. I think that would be the best solution for the current architecture, but this will break down if there are multiple generated models with different `deleted` columns, since they will all change the global `rgx` and whichever is last will win.

In a hypothetical `v5` the query mod to remove `deleted_at` should move into generated code, but that would need some changes to the query object.

templates/main/13_all.go.tpl
lopezator: It makes sense, changed!

lopezator:
@stephenafamo Thanks for the quick review! I simplified the things a bit, let me know what you think :slightly_smiling_face:

templates/main/singleton/boil_queries.go.tpl
lopezator: >  In a hypothetical v5 the query mod to remove deleted_at should move into generated code

I agree! Should I open a new issue labeled for v5 and so we can go over it later?

stephenafamo: Yes, that would be nice. Also if making any changes, send the PR to the `v5` branch.

lopezator: Gotcha! :+1:

templates/main/singleton/boil_queries.go.tpl
```diff
@@ -12,10 +12,18 @@ var dialect = drivers.Dialect{
...
+    var _ = &regexp.Regexp{}
+{{- end}}
+
 // NewQuery initializes a new Query using the passed in QueryMods
 func NewQuery(mods ...qm.QueryMod) *queries.Query {
 	q := &queries.Query{}
 	queries.SetDialect(q, &dialect)
+	{{- if and (.AutoColumns.Deleted) (ne $.AutoColumns.Deleted "deleted_at") }}
+		queries.SetRemoveSoftDeleteRgx(regexp.MustCompile("{{$.AutoColumns.Deleted}}[\"'`]? is null"))
+	{{- end }}
```
stephenafamo, line 26: I think just using an init function would be better. This would call the `SetRemoveSoftDeleteRgx` every time a new query is created.

lopezator: Right! Changed :)

lopezator:
Changes made, PTAL @stephenafamo :slightly_smiling_face:

stephenafamo (approved):

lopezator:
Thank your for all the support and responsiveness @stephenafamo !! Btw are you planning to release a bugfix release anytime soon? Thanks again 🙏.

Issue #499, opened 2019-03-20
Concurrent Log Writes


natdm:
### What version of SQLBoiler are you using (`sqlboiler --version`)? 3

Running sql queries concurrently log the queries and arguments in random order 
eg:
```
SELECT WHERE...
SELECT WHERE...
[1212]
[1315]
SELECT WHERE ...
[1111]
```

Proposal:
Safer concurrent logging and don't separate the query and the args

aarondl:
You can replace the logger with whatever you'd like so long as it conforms to the io.Writer interface in order to make it goroutine safe:
https://github.com/volatiletech/sqlboiler/blob/master/boil/global.go#L25

Not separating query and args could definitely be done but it will make it harder to see. Certainly you cannot put a newline between them in a log because some people may be using structured logging adapters that may not play nice with that sort of thing. Although maybe it'd be escaped. Open to suggestions here.

natdm:
I think it might be more about the fact that they make multiple writes at the same time (and can be done safely) but then they intertwine with each-other, making the logs useless to read since you don't know what is doing what with which arguments. Maybe the solution is to do something like `sarama.Logger`, which can easily be replaced with l`og.New()`, or` zap.Logger.Sugar()`, and then can be written to with formatted lines.

aarondl:
@natdm Still happy to accept a PR to merge these outputs into a single call to log.

PR #1267, opened 2023-04-10
Add Inline query + args print

cynsupercat:
Attempt at https://github.com/volatiletech/sqlboiler/issues/499

It works well with non `Null` types but unfortunately `Null` types will always output their `valid` values like so:

```
"INSERT INTO \"rockets\" (\"id\",\"name\",\"model\",\"launched\",\"launch_date\",\"data\") VALUES ('1c309a44-c3b6-4924-b5b4-57532cff1c1f',{bunpollo 13 true},{1 true},{true true},{2023-04-10 02:59:16.207365677 +0000 UTC m=+9.860239922 true},{[123 34 100 97 116 97 34 58 34 116 101 115 116 34 125] true})"
``` 

From a personal standpoint, this is an improvement to current logging so I'm happy with this, but this might be a naive approach so any suggestions welcome.

### DIFF
--- a/boil/debug.go
+++ b/boil/debug.go
@@ -58,9 +58,9 @@ func PrintQuery(writer io.Writer, query string, args ...interface{}) {
 	fmt.Fprintln(writer, substitutedQuery)
 }
 
-// substituteQueryArgs takes a SQL query string and an array of
-// // arguments. It returns a modified query string with placeholders replaced by their
-// // corresponding argument values.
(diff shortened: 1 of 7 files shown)


cynsupercat:
@stephenafamo I suppose you could do an opt in using a `func (writer, query, args)` global variable in debug.go to override during logging if it's not nil?

stephenafamo:
I suppose that could work. While I like to avoid globals, this may be the best way to avoid breaking changes.

Issue #1218, opened 2022-10-31
Add `LIKE` and `ILIKE` Operators


falsaffa:
## Overview

Add `LIKE` and `ILIKE` to the set of operators used in `[MODEL]Where.[COLUMN].[OPERATOR]`. Below is a sample of the existing operators supported including: `EQ`, `GTE`, `GT`, `IN`, `LT`, `LTE`, `NEQ`, `NIN`.

<img width="804" alt="Screenshot 2022-10-30 at 10 25 20 PM" src="https://user-images.githubusercontent.com/44938179/198924749-929b91aa-4ede-4d37-bf2b-895501a2d2a8.png">

With the `LIKE` and `ILIKE` operator, the following:

```go
qm.Where(gen_models.ProductTableColumns.Name+" ilike ?", queryTarget),
```

turns into:

```go
gen_models.ProductTableColumns.Name.ILIKE(productID.String())
```

I may be alone in this, but I use the `LIKE` and `ILIKE` operators readily enough that having shorthand would be useful.

## BTW

Absolutely love sqlboiler. I wish I knew of it early! ❤️

stephenafamo:
Officially, SQLBoiler supports 4 dialects.

* Postgres
* MySQL
* SQL Server
* SQLite

To add `LIKE` and `ILIKE`, they have to be operators in all 4 dialects. Alternatively, we could add some check to make sure the dialect supports it before generating the code for the operators, but that will not be as straightforward.

falsaffa:
If I were to open a PR using the code generation method (check the dialect before adding the enum) would that be an implementation you would be happy to review/approve?

stephenafamo:
Yes. Alternatively, you can add it under the driver templates, the same way dialect-specific code for Upserts are generated.

falsaffa:
So played with it a bit and I have a couple questions:
- `qmhelper` is it intended to be widly used outside of sqlboiler code? Considered adding `ILIKE` and `LIKE` operator, but since they aren't supported for all dialects it may seem deceptive to include?
- `LIKE` is a common operator across the dialects, so that makes sense to include in the core templates. However, `ILIKE` is unique to postgres. I thought of just including it in the core template only if the driver given is postgres, but how strongly are your thoughts in having it be a separate template file in postgres? (only postgres since it's the only driver that supports it)

stephenafamo:
- I think a separate template file is best for any driver-specific template. This is their intended purpose.
[...]

PR #1280, opened 2023-06-01
add string pattern matching to templates

jerrysalonen:
Based on issue [Add LIKE and ILIKE Operators #1218](https://github.com/volatiletech/sqlboiler/issues/1218)

There was discussion in the issue that this should not be added to `qmhelper` so I just used `qm.Where`. If there are no technical reasons to not extend `qmhelper` I think adding at least the shared operators there would fit nicely. I ended up extending the base templates with the `LIKE` and `NOT LIKE` operators because those should be shared between all the supported dialects. Since postgres is the only one that supports `ILIKE`, I added that as an override. I tried to think over how to add tests to these but did not come up with anything so far.

Tested the resulting models with postgres, could use with more dialects being manually tested

p.s. this is my first contribution to open source so all feedback is very appreciated


jerrysalonen:
@stephenafamo good point! fixed

Issue #1163, opened 2022-07-23
Feature: Define relationships in config


nhuysamen-godaddy:
This is probably a non-starter, but it would be great (albeit probably error-prone), to be able to specify relationships in config rather than relying on foreign keys. This would allow you to use sqlboiler with platform like [PlanetScale](http://planetscale.com/) that does not support foreign keys.

stephenafamo:
This sounds interesting. If you can create a PR for this feature, I'll be happy to review it.

tanaka-takuto:
I was just about to use SQLBoiler with PlanetScale!
I had no choice but to try to use it without foreign key constraints (no relations).
This is a long awaited feature!

Rosswell:
imo this would be a nice feature in dbs with foreign keys as well, as afaict the reverse relationship is auto-derived, and cannot be renamed through a toml config, because the foreign key only exists on one of the tables

stephenafamo:
@Rosswell I believe it is possible to alias the reverse relationship using the `foreign` property.

Regardless, I agree that the ability to manually define relationships would be quite useful. If only to create some "convenient" relationships that you don't necessarily need to be enforced by the DB.

chrisngyn:
@stephenafamo Could you review this PR above?

adsteel:
Would love to see this as well. Not being able to add non-FK columns to a join table is a limitation this would help to remove.

stephenafamo:
Implemented in #1282

PR #1282, opened 2023-06-06
Allow define foreign key relationships in config file

chrisngyn:
## What?
I've added support for define relationships in config file. This would allow you to use sqlboiler with platform that does not support foreign keys.
For more background, see #1163 

## Proposal configs
Here is an example config will be added to config file.
```toml
#...
[foreign_keys.jet_pilots_fkey]
table = "jets"
column = "pilot_id"
foreign_table = "pilots"
foreign_column = "id"

[foreign_keys.pilot_language_pilots_fkey]
table = "pilot_languages"
column = "pilot_id"
foreign_table = "pilots"
foreign_column = "id"

[foreign_keys.pilot_language_languages_fkey]
table = "pilot_languages"
column = "language_id"
foreign_table = "languages"
foreign_column = "id"
#...
```

### DIFF
--- a/boilingcore/config.go
+++ b/boilingcore/config.go
@@ -48,10 +48,11 @@ type Config struct {
 	DefaultTemplates    fs.FS            `toml:"-" json:"-"`
 	CustomTemplateFuncs template.FuncMap `toml:"-" json:"-"`
 
-	Aliases      Aliases       `toml:"aliases,omitempty" json:"aliases,omitempty"`
-	TypeReplaces []TypeReplace `toml:"type_replaces,omitempty" json:"type_replaces,omitempty"`
-	AutoColumns  AutoColumns   `toml:"auto_columns,omitempty" json:"auto_columns,omitempty"`
-	Inflections  Inflections   `toml:"inflections,omitempty" json:"inflections,omitempty"`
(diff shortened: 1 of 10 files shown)


stephenafamo:
Sorry, this is a fairly big change and I haven't had time to review it properly. I will try to find some time over the next few weeks

chrisngyn:
I know, the main reason make this PR so big because I've changed `drivers.Config` from `map[string]interface{}` to `struct`. I think that will make the `Config` be more explicit.
And the interface function will be more shorter.
```go
func TablesConcurrently(c Constructor, config Config) ([]Table, error)
```
instead of
```go
func TablesConcurrently(c Constructor, schema string, whitelist, blacklist []string, concurrency int, foreignKeys []drivers.ForeignKey) ([]Table, error)
```
If I'm missing anything or you need me update this PR, feel free to tell me.

stephenafamo:
> I've changed `drivers.Config` from `map[string]interface{}` to `struct`

The thing is since `drivers.Config` is exported, this would become a "breaking change" and should only then be included in `v5`.

I would suggest you undo that for now. While it looks clumsy, it is the price of maintaining backward compatibility.

chrisngyn:
> I would suggest you undo that for now
I've already undone that

PR #1287, opened 2023-06-30
Fix parsing pgeo point in scientific notation format

zhongduo:
Databases, at least Postgres, can save points using scientific notation for small numbers. Using a completed regex for floating point is nontrivial, also not useful as ParseFloat should be smart enough to validate the number format. This commit relaxes the regex and rely on ParseFloat.


zhongduo:
@stephenafamo would you mind reviewing this PR? Thanks.

PR #1291, opened 2023-07-15
Downgrade decimal version before decompose interface

### LINKED ISSUE #803: Unable to convert type: types.Decimal
v4 (4.2.0)

MySQL - 8.0.19

Inserting a record, table has a field using type: `decimal(5,4)`

### What is the output of the command above with the `-d` flag added to it?
unable to insert into (table_name): cannot convert type: types.Decimal

```
CREATE TABLE `test_tables` (
`decimal_field` decimal(5,4) NOT NULL DEFAULT 0.00 COMMENT 'This field has the issue'
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_general_ci COMMENT='Test model';
```

I thought this issue would be fixed by now. 

I saw something about hardcoding the ericlagergren decimal package, but I want to make sure I'm not missing something...

zhongduo:
Any version after https://github.com/ericlagergren/decimal/pull/124 causes problems when saving Decimal to sql drivers like postgres. While the best solution may be to update all the sql drivers, it has been 3 years since then, so it is better to make sure the latest version does not have that commit. This version is the latest before it.

close #803

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -126,8 +126,8 @@ github.com/envoyproxy/go-control-plane v0.9.9-0.20210512163311-63b5d3c536b0/go.m
 github.com/envoyproxy/go-control-plane v0.9.10-0.20210907150352-cf90f659a021/go.mod h1:AFq3mo9L8Lqqiid3OhADV3RfLJnjiw63cSpi+fDTRC0=
 github.com/envoyproxy/go-control-plane v0.10.2-0.20220325020618-49ff273808a1/go.mod h1:KJwIaB5Mv44NWtYuAOFCVOjcI94vtpEz2JU/D2v6IjE=
 github.com/envoyproxy/protoc-gen-validate v0.1.0/go.mod h1:iSmxcyjqTsJpI2R4NaDN7+kN2VEUnK/pcBlmesArF7c=
-github.com/ericlagergren/decimal v0.0.0-20211103172832-aca2edc11f73 h1:odNUt+pGupjtZyfaNIGLT/PUxT7r3fZ0Kf+QH9reIoM=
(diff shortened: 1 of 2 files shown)


go.sum
```diff
@@ -454,8 +454,6 @@ github.com/volatiletech/null/v8 v8.1.2/go.mod h1:98DbwNoKEpRrYtGjWFctievIfm4n4Mx
 github.com/volatiletech/randomize v0.0.1 h1:eE5yajattWqTB2/eN8df4dw+8jwAzBtbdo5sbWC4nMk=
 github.com/volatiletech/randomize v0.0.1/go.mod h1:GN3U0QYqfZ9FOJ67bzax1cqZ5q2xuj2mXrXBjWaRTlY=
 github.com/volatiletech/strmangle v0.0.1/go.mod h1:F6RA6IkB5vq0yTG4GQ0UsbbRcl3ni9P76i+JrTBKFFg=
-github.com/volatiletech/strmangle v0.0.4 h1:CxrEPhobZL/PCZOTDSH1aq7s4Kv76hQpRoTVVlUOim4=
```
zhongduo, line 457: this is deleted by `go mod tidy`

zhongduo:
Other related issues from this: 
https://github.com/ericlagergren/decimal/issues/175
https://github.com/ericlagergren/decimal/issues/154

zhongduo:
@stephenafamo while I am can modify my local go.mod for my project, I think it makes sense to make sure the latest version of sqlboiler will just work. Today if someone use the latest version, they will always hit this problem.

zhongduo:
@stephenafamo thanks for merging this. I would also suggest to cut a new version to stop the bleeding for new users. But obviously it depends on the release schedule.

Issue #1298, opened 2023-08-29
[Feature Request] Ignore Composite Foreign Key Constraint


paulo-raca:
### What version of SQLBoiler are you using (`sqlboiler --version`)?
SQLBoiler 4.14.2

### What is your database and version (eg. Postgresql 10)
PostgreSQL 14.4

### If this happened at generation time what was the full SQLBoiler command you used to generate your models? (if not applicable leave blank)
`sqlboiler psql -p dbmodels`

### Please provide a relevant database schema so we can replicate your issue (Provided you are comfortable sharing this)

```sql
DROP TABLE IF EXISTS node;

CREATE TABLE node (
    id int PRIMARY KEY,
    parent_id int,
    root_id int NOT NULL
);

-- Sanity check: On root nodes only, parent_id=NULL and root_id=id
ALTER TABLE node
ADD CONSTRAINT check_root_is_self_if_no_parent CHECK ((root_id = id) = (parent_id IS NULL));

-- Sanity check: node.root_id is the same as nodex.parent.root_id
ALTER TABLE node
ADD CONSTRAINT unique_id_rootid UNIQUE (id, root_id);

ALTER TABLE node
ADD CONSTRAINT fk_id_rootid FOREIGN KEY (parent_id, root_id) REFERENCES node(id, root_id);
```

### Further information. What did you do, what did you expect?

For context, my database contains tree structures. Within each node I have links to the parent and root nodes (Yes, this is redundant, but having a direct link to the root makes some queries much simpler)

To ensure my DB remains consistency, I've added the `fk_id_rootid` composite constraint. which ensures all nodes in the tree have the same value of `root_id`.

This works as expected, here is a tiny test suite:
```sql
-- Success Test: Creates 1 -> 2 -> 3
INSERT INTO node(id, parent_id, root_id) values (1, NULL, 1);
INSERT INTO node(id, parent_id, root_id) values (2, 1, 1);
INSERT INTO node(id, parent_id, root_id) values (3, 2, 1);

-- Failure: violates check constraint "check_root_is_self_if_no_parent"
-- When adding a root node (parent_id=NULL), we must have (root_id = id)
INSERT INTO node(id, parent_id, root_id) values (4, NULL, 1);
-- Failure: violates foreign key constraint "fk_id_rootid"
-- When adding a child node, the root_id must match those of the parent
INSERT INTO node(id, parent_id, root_id) values (4, 3, 2);
```

[...]

PR #1299, opened 2023-08-30
Ignore Composite Foreign Keys.

paulo-raca:
Boiler doesn't support composite FKs -- And this is OK.

However if there is one on the database, it should ignore it and generate valid code.

This addresses #1298


paulo-raca:
I've updated it with tests for Postgres+MySQL+SQLite3.

I had troubled installing mssql-tools tools on my machine, is it ok to skip it?

adsteel:
@stephenafamo This is pretty exciting and currently blocking us. What's the chance we could get a release cut that includes this fix? Happy to help with that if there's anything I/we can do, too.

stephenafamo:
The latest release should already include it. Are you still having issues?

adsteel:
We are on 4.15, but I think this merged a few days after that release shipped.

stephenafamo:
That's true, I got things mixed up, I haven't done a release in a while. 

I'll tag a new version.

adsteel:
Anything I can do to help get a new release out?

stephenafamo:
Released now. Apologies for the delay @adsteel

adsteel:
@stephenafamo no problem at all. This is exciting, thank you!

Issue #1296, opened 2023-08-23
Upsert not returning all model attributes on conflict


adsa95:
If you're having a generation problem please answer these questions before submitting your issue. Thanks!

### What version of SQLBoiler are you using (`sqlboiler --version`)?
SQLBoiler v4.15.0

### What is your database and version (eg. Postgresql 10)
Postgresql 12.5

### If this happened at runtime what code produced the issue? (if not applicable leave blank)
```go
u := &model.User{
	ID:        "newClientGeneratedId", // generated in go, not using database generated id
	Email:     "user@example.org",
	DeletedAt: null.Time{},
}

err = u.Upsert(
	db,
	true,
	[]string{model.UserColumns.Email},
	boil.Whitelist(model.UserColumns.DeletedAt),
	boil.Infer(),
)
if err != nil {
	panic(err)
}

log.Println(u.ID)
```

### Please provide a relevant database schema so we can replicate your issue (Provided you are comfortable sharing this)
```sql
CREATE TABLE USERS (
	id varchar not null primary key, 
	email varchar not null unique, 
	deleted_at timestamp with time zone
);
```

### Further information. What did you do, what did you expect?
I'm trying to use upsert  to undelete (set deleted_at=null) or add a new row, but the model I'm using to upsert the row is not refreshed with all details from the existing row in case of a conflict, the main issue being that the primary key (ID) is not updated.

With the following data in the database:
| id | email | deleted_at |
| --- | --- | --- |
| abc123 | user@example.org | 2023-08-23 11:18:00 |

I expected the upsert to update the existing database row, which it did, and then update the `u` struct with all existing data so that the log call at the end would print `abc123`. However, the ID of the struct is not updated and the log call instead prints `newClientGeneratedId`. Since the ID is not updated and the `u` struct still has an ID that does not exist in the database, using this ID in following code would probably cause some major issues.

I've looked through some of the sqlboiler code and think this is because the return values in the generated query is the same for an insert as an upsert, which is columns-with-default minus some columns depending on the type of `Columns` used. **Note that we're generating the IDs in go before upserting and does not use any database generated IDs**, hence the ID field is not in columns-with-defaults. However I would argue that with an upsert you'd expect the whole model to get updated with values from the database if they exist, so maybe all columns should be returned in this case? When I manually added the ID column to the return set it was processed and set as I expected.

stephenafamo:
Can you turn on `boil.DebugMode` and share the queries that were generated and ran?

[...]

PR #1306, opened 2023-09-13
Return all columns not in both insert and update columns when doing upsert

adsa95:
I would argue that a model should be completely up to date with the database after an upsert, however as demonstrated in #1296 this is currently not always the case. This PR is a proposed fix for this. By returning all columns not present in both the insert and update column sets we make sure that the model is always up to date with the database after an upsert, no matter if an insert or update was performed.

As `updateColumns` and `insertColumns` are both `Columns` structs that might have different types (infer, whitelist etc) I did not find it suitable to add a method on the `Columns` struct to calculate this and do it different depending on type. Instead opting to just always calculate the return columns the same way.

Although the biggest issue in my example in #1296 was that the ID (primary key) was not updated when generated in go code and not in the database, I still think all properties on the model should be updated to match the database when an upsert is performed. Hence not just adding a check to make sure primary keys are in the return set.

Happy to hear feedback on this!

**Note: This PR also relies on the `SetIntersect()` method being added to the strmangle package. See https://github.com/volatiletech/strmangle/pull/11**

### DIFF
--- a/drivers/sqlboiler-mssql/driver/override/main/17_upsert.go.tpl
+++ b/drivers/sqlboiler-mssql/driver/override/main/17_upsert.go.tpl
@@ -71,7 +71,7 @@ func (o *{{$alias.UpSingular}}) Upsert({{if .NoContext}}exec boil.Executor{{else
 	var err error
 
 	if !cached {
-		insert, ret := insertColumns.InsertColumnSet(
+		insert, _ := insertColumns.InsertColumnSet(
 			{{$alias.DownSingular}}AllColumns,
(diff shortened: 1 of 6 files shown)


stephenafamo:
The PR on `strmangle` has been merged, so you should update `sqlboiler` to depend on `v0.0.6` of strmangle so that the user's version gets updated once they update `sqlboiler` and regenerate the models with it.

Once that is done, I'll be happy to merge this. Thank you.

adsa95:
This should now be updated! Thanks for looking at this PR :)

PR #1311, opened 2023-09-28
Update README.md about materialized view support for postgresql

pankona:
README.md saids materialized view is not supported but it psql is supported
Refs: https://github.com/volatiletech/sqlboiler/pull/1154


pankona:
@stephenafamo Thank you! 😋

PR #1317, opened 2023-10-17
Ability to extend upsert expression with options

atzedus:
This PR adds ability to provide some options to extend UPSERT expression.
For example, following sql needed:

```SQL
CREATE TABLE test (name text);
CREATE UNIQUE INDEX ON test(lower(name)) WHERE name IS NOT NULL;

INSERT INTO test (name) ON CONFLICT (lower(name)) WHERE name IS NOT NULL DO NOTHING;
```

Currently it not possible with "Upsert" method.
This PR adds ability to provide options to Upsert method, for example:

```go
m := &models.Test{ name: "test" }

m.Upsert(ctx, db, false, nil, boil.Infer(), boil.Infer(), models.UpsertConflictObject("(lower(name)) WHERE name IS NOT NULL"))
```

Also, it adds UpsertUpdateSet(string) option to customize DO UPDATE SET expression to any custom value.


stephenafamo:
This looks quite nice and is fully backwards-compatible. Thank you.

PR #1322, opened 2023-11-04
fix: not using quotes on column names

Flo4604:
Fix potential issue where column name like "createdAt" would not be using quotes


stephenafamo:
This PR seems to have caused a bug in #1341. Can you please take a look @Flo4604

stephenafamo:
@eirikbell if you can verify that reverting this commit fixes it, let me know.

serhii-tokranov:
If I understand correctly, "add quotas logic" (or whatever dialect-specific symbol) is implemented at the beginning of this method. Meanwhile, in this PR, extra quotas are added whether needed there or not.

eirikbell:
I can confirm that reverting the commit https://github.com/volatiletech/sqlboiler/pull/1322/commits/4cd18bddc416b9662781df49792d64ae500e29e0 will fix the issue

PR #1329, opened 2023-11-22
no-relationship-funcs flag added

atzedus:
This PR adds new model generation functionality with "no-relationship-funcs" flag.
This flag disable generation helper methods for model relationships loaders.
In some situations its better do disable this functionality, rather then getting N+1 problem on code reviews.

### DIFF
--- a/drivers/sqlboiler-psql/driver/psql.go
+++ b/drivers/sqlboiler-psql/driver/psql.go
@@ -11,6 +11,7 @@ import (
 	"io/fs"
 	"os"
 	"strings"
+	"sync"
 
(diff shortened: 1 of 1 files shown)


stephenafamo:
This is a big change and I am reluctant to merge it because we will then have to maintain this flag indefinitely.

atzedus:
Well. This is a completely reasonable decision. The popularity of this feature is unknown. However, I would like to suggest leaving the template reorganization in the tests from this PR. This will make it possible to easily customize them to achieve the same effect by simply deleting the *_relationship_* files. If you accept this, thrn I will make changes soon.

PR #1347, opened 2024-01-26
Add Go module files

quells-at-work:
Adds `go.mod` and `go.sum` to pin dependencies to versions which still support Go 1.18.


quells-at-work:
Sorry - I was trying to merge this into our fork.

PR #1348, opened 2024-02-01
Added hooks to templates for relationships.

### LINKED ISSUE #445: Proposal: doBeforeUpdateHooks in SetRelation function
I have a validation function I run as a `BeforeUpdateHook` on one model. It runs some checks an associated model (in the context of the local model), and it would improve code flow if I could rely on 

`Local.SetForeign(_, _, foreignObject)`

rather than setting the foreign ID and calling `local.Update`.

I'd love to take a stab at this if you don't have any objections to it.

mmitten:
My attempt to resolve #445 

Added into the templates for relationships as I did not see any other place that made sense for how the hooks otherwise behaved. Happy to follow up if a something was missed.

### DIFF
--- a/templates/main/10_relationship_to_one_setops.go.tpl
+++ b/templates/main/10_relationship_to_one_setops.go.tpl
@@ -65,6 +65,12 @@ func (o *{{$ltable.UpSingular}}) Set{{$rel.Foreign}}({{if $.NoContext}}exec boil
 		}
 	}
 
+	{{if not $.NoHooks -}}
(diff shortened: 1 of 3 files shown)


stephenafamo:
While I see this as a good change, I wonder if this may cause issues for those who are used to the current behaviour 🤔

mmitten:
Now that you mention it, it could be a breaking change, as adding another hook trigger where it was previously not expected could lead to some unpleasant side effects.

I am not opposed to tabling this for a major version revision, as the use case I wrote it for has since been resolved with something else.

stephenafamo:
I'll label this as breaking and may add it to an eventual `v5` release.

PR #1357, opened 2024-02-16
Lowercase column names before applying `TitleCase`

esdrasbeleza:
I have a database which I cannot change and all the fields are in `SCREAMING_SNAKE_CASE`.

When I generate my models, `USER_ID` becomes `USER_ID` in my Go structs, which is not what I want and doesn't follow the expected Go conventions: I expect my field to be generated as `UserID`.

This small change will make my model be properly generated.

### DIFF
--- a/boilingcore/aliases.go
+++ b/boilingcore/aliases.go
@@ -2,6 +2,7 @@ package boilingcore
 
 import (
 	"fmt"
+	"strings"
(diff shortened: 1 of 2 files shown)


stephenafamo:
This would be a breaking change for many users, so I will close this.

For your case, I suggest renaming them using aliases.

Issue #1359, opened 2024-03-01
Increase Depth Limit in ptrFromMapping Function for Deeper Structure Access in reflect.go


doscode-kr:
### What version of SQLBoiler are you using (`sqlboiler --version`)?
- SQLBoiler v4.16.2

### What is your database and version (eg. Postgresql 10)
- MariaDB 10.6.16

### If this happened at generation time what was the full SQLBoiler command you used to generate your models? (if not applicable leave blank)
- ( )

### If this happened at runtime what code produced the issue? (if not applicable leave blank)
```go
...
	records, err := models.AnyTableWithMoreThan255Columns(
		Select(),
	).All(ctx, db)
...
```

### What is the output of the command above with the `-d` flag added to it? (Provided you are comfortable sharing this, it contains a blueprint of your schema)
- ( )

### Please provide a relevant database schema so we can replicate your issue (Provided you are comfortable sharing this)
```sql

CREATE TABLE `AnyTableWithMoreThan255Columns` (
  `id` bigint(20) NOT NULL AUTO_INCREMENT,
  `created_at` datetime NOT NULL DEFAULT current_timestamp(),
  `updated_at` datetime DEFAULT NULL,
  `deleted_at` datetime DEFAULT NULL,
  `column1` text NOT NULL DEFAULT '',
  `column2` text NOT NULL DEFAULT '',
  ....
  `column255` text NOT NULL DEFAULT '',
  ...
  `column500` text NOT NULL DEFAULT '',
```

### Further information. What did you do, what did you expect?

I have designed a special-purpose table to store hundreds of text columns. 
However, when creating the table and querying the data, I encountered the following error message:

```
[models/research_data_vault.go:2388] models: failed to assign all query results to ResearchDataVault slice: 
[sqlboiler/v4@v4.16.2/queries/reflect.go:262] failed to bind pointers to obj:
sql: Scan error on column index 255, name "r7_use_bispho": unsupported Scan, storing driver.Value type []uint8 into type *models.ResearchDataVault
```

The key takeaway from this error message is the number `255`. The table I created during testing had about 300 columns.

in reflect.go

```go
...
    //reflect.go:231
    mapping, err := getMappingCache(structType).mapping(cols)
...
    //reflect.go:471
    func (b *mappingCache) mapping(cols []string) ([]uint64, error) {
...
    //reflect.go:491
    mapping, err := BindMapping(b.typ, b.structMap, cols)
...
}
```

[...]

PR #1360, opened 2024-03-01
EDIT - Increase Depth Limit in ptrFromMapping

doscode-kr:
- Increase Depth Limit in ptrFromMapping Function for Deeper Structure Access in reflect.go
- This fix addresses errors encoutered with tables exceeding 255 columns.
- this fixes issue #1359


doscode-kr:
I found that this is same issue as below :
https://github.com/volatiletech/sqlboiler/pull/1038

PR #1375, opened 2024-04-05
feature: set custom table name for insertion

c9s:


c9s:
@stephenafamo please take a look,, thanks!

stephenafamo:
What is the use case for this?

c9s:
> What is the use case for this?

for database sharding, for large datasets we need to move these data in different tables with the same schema

c9s:
this is the related ticket (discussion)
https://github.com/volatiletech/sqlboiler/discussions/1378

templates/main/00_struct.go.tpl
```diff
@@ -71,6 +74,8 @@ var {{$alias.UpSingular}}TableColumns = struct {
 	{{end -}}
 }
 
+const {{$alias.UpSingular}}TableName = "{{.Table.Name}}"
```
stephenafamo, line 77: This is unnecessary, there is a struct with table names in the singleton. Let's reuse that.

stephenafamo (changes requested):
Still not sure about this use case for most users, but I have a few proposed changes to the PR.

stephenafamo:
> for database sharding, for large datasets we need to move these data in different tables with the same schema

Instead of making this change to SQLBoiler, did you consider having extra templates to add an extra method to the models?  
<https://github.com/volatiletech/sqlboiler?tab=readme-ov-file#templates>

c9s:
> > for database sharding, for large datasets we need to move these data in different tables with the same schema
> 
> Instead of making this change to SQLBoiler, did you consider having extra templates to add an extra method to the models? https://github.com/volatiletech/sqlboiler?tab=readme-ov-file#templates


Thank you for the suggestion. I have considered the template approach, but it seems overly complex for this particular need. Given SQLBoiler's status as a modern ORM tool, I believe incorporating native support for database sharding would significantly enhance its utility for projects requiring robust scalability. This feature would not only simplify development but also align SQLBoiler with the evolving demands of large-scale applications.

c9s:
> > for database sharding, for large datasets we need to move these data in different tables with the same schema
> 
> Instead of making this change to SQLBoiler, did you consider having extra templates to add an extra method to the models? https://github.com/volatiletech/sqlboiler?tab=readme-ov-file#templates

Hi @stephenafamo 

Thank you for the suggestion. I have considered the template approach, but it seems overly complex for this particular need. Given SQLBoiler's status as a modern ORM tool, I believe incorporating native support for database sharding would significantly enhance its utility for projects requiring robust scalability. This feature would not only simplify development but also align SQLBoiler with the evolving demands of large-scale applications.

templates/main/00_struct.go.tpl
c9s: this is fixed

c9s:
Or maybe you prefer other API design? maybe we can discuss?

stephenafamo:
The API looks good.

For now, since this is not a very common usecase, I would suggest putting it behind a generation flag.

If you can do that, I'll be happy to merge.

c9s:
> The API looks good.
> 
> For now, since this is not a very common usecase, I would suggest putting it behind a generation flag.
> 
> If you can do that, I'll be happy to merge.

Sure! I am willing to do that

Thanks!

PR #1380, opened 2024-04-19
Implement (Un-)marshalText for Decimal and NullDecimal

MJacred:
UnmarshalText is required for parsing a text/csv file. Without, the code uses decimal directly and ends in a nil pointer exception because decimal.Big is nil on `scan`


MJacred:
@stephenafamo, please have a look. Thanks!

PR #1385, opened 2024-05-22
templates: conditionally set created automatic column

glerchundi:
This change takes into account if the the created automatic column was set by the user in the update operation. If set, it will set the one provided by the user. If not set, it will set to the default "created_at" value.


glerchundi:
Thanks @stephenafamo! 🙏!

Issue #1377, opened 2024-04-06
Fix comment position of first column of table


i16fujimoto:
### What version of SQLBoiler are you using (`sqlboiler --version`)?
SQLBoiler v4.16.0

### Problem
Comment position in first column of table does not break line

### Example auto generation code for sturuct
```go
type Item struct { // 商品番号
	ID int `boil:"id" json:"id" toml:"id" yaml:"id"`
	// メーカー
	Maker string `boil:"maker" json:"maker" toml:"maker" yaml:"maker"`
	// 作成日時
	CreatedAt time.Time `boil:"created_at" json:"created_at" toml:"created_at" yaml:"created_at"`
	// 更新日時
	UpdatedAt time.Time `boil:"updated_at" json:"updated_at" toml:"updated_at" yaml:"updated_at"`
	// 削除日時
	DeletedAt null.Time `boil:"deleted_at" json:"deleted_at,omitempty" toml:"deleted_at" yaml:"deleted_at,omitempty"`

	R *itemR `boil:"-" json:"-" toml:"-" yaml:"-"`
	L itemL  `boil:"-" json:"-" toml:"-" yaml:"-"`
}
```

### Request
I want to fix following file, but I don't have permission to push to this repository, so I want to get it.
https://github.com/volatiletech/sqlboiler/blob/15d15d1038f359cdd3a795a472e02ab4cc550f58/templates/main/00_struct.go.tpl#L9-L10

winterjung:
I hope for this too. You can fork the repository, then commit your fix to the `master` branch of the forked repository (i.e., `i16fujimoto/sqlboiler`) and open a pull request into `volatiletech/sqlboiler`.

PR #1390, opened 2024-06-08
fix comment position in first column of table

hizzuu:
My attempt to resolve https://github.com/volatiletech/sqlboiler/issues/1377

The comment position in the first column of the table has been corrected to a new line.

### DIFF
--- a/templates/main/00_struct.go.tpl
+++ b/templates/main/00_struct.go.tpl
@@ -3,10 +3,13 @@
 
 // {{$alias.UpSingular}} is an object representing the database table.
 type {{$alias.UpSingular}} struct {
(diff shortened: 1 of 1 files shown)


hizzuu:
@stephenafamo

stephenafamo:
Have you tried this yourself?  @hizzuu 

I haven't but from what I understand of Go templating syntax, I don't think your change would fix it.

hizzuu:
@stephenafamo 
I have tested it myself and confirmed that line breaks are applied correctly.
I would appreciate another review.

[87a0d4d](https://github.com/volatiletech/sqlboiler/pull/1390/commits/87a0d4d9637603efde507e62e9b1580176201a17)

PR #1395, opened 2024-06-25
feat: add version checking flags

090809:
In some cases, using sqlboiler in old projects can be painful, 'cause version in project and on machine not the same.
These flags will let developers know as soon as possible that their project is slightly outdated


stephenafamo:
This is good.  So good I want this to be the default behaviour.

I think it should check and warn, should always happen, but the `strict` mode should be the flag.

090809:
@stephenafamo updated =)

PR #1398, opened 2024-07-08
Count from subquery if query uses HAVING or GROUP BY

renom:
This PR enables using statement `SELECT COUNT(*) FROM (SELECT ...);` if there's HAVING or GROUP BY in the query.

### DIFF
--- a/queries/query_builders.go
+++ b/queries/query_builders.go
@@ -53,6 +53,14 @@ func buildSelectQuery(q *Query) (*bytes.Buffer, []interface{}) {
 	writeComment(q, buf)
 	writeCTEs(q, buf, &args)
 
+	hasHaving := len(q.having) != 0
+	hasGroupBy := len(q.groupBy) != 0
+	hasSimpleCount := q.count && !hasHaving && !hasGroupBy
+	hasComplexCount := q.count && (hasHaving || hasGroupBy)
+	if hasComplexCount {
+		buf.WriteString("SELECT COUNT(*) FROM (")
+	}
+
 	buf.WriteString("SELECT ")
 
 	if q.dialect.UseTopClause {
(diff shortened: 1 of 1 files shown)


stephenafamo:
Why is this necessary? Are the queries more performant?

renom:
It's necessary because aggregate functions are run for each group separately.
So if there's `GROUP BY`, `COUNT(*)` will count a number of grouped rows in each group (i.e. `COUNT(*)` will be calculated for each returned row).
But we need to return the count of returned rows.

renom:
https://www.w3schools.com/sql/trysql.asp?filename=trysql_editor
Try here the following queries:
```sql
SELECT Country, COUNT(*) FROM Customers GROUP BY Country;
```
vs
```sql
 SELECT COUNT(*) FROM (SELECT Country FROM Customers GROUP BY Country);
```
The first query returns `COUNT(*)` for each country, the second query returns `COUNT(*)` of countries.
The first query returns 21 rows, the second query return one row with value `21`.

eklatzer:
@stephenafamo @renom Hey, I am currently experiencing the following issue and I think the reason is this change: `pq: subquery in FROM must have an alias`

Here an example:
```sql
CREATE TABLE items (
    id serial PRIMARY KEY,
    category varchar(255) NOT NULL,
    is_special boolean NOT NULL
);
```

Example usage:
```go
count, err := models.Items(
	models.ItemWhere.IsSpecial.EQ(true),
	// group because of possible joins to other tables
	qm.GroupBy(models.ItemColumns.ID),
).Count(boil.WithDebug(ctx, true), s.DB)
require.NoError(t, err)
```

SQL:
```sql
SELECT COUNT(*) FROM (SELECT * FROM "items" WHERE ("items"."is_special" = $1) GROUP BY id);
[true]
```

You can try it in https://onecompiler.com/postgresql/435bk9nra if you remove the `as foo`

PR #1403, opened 2024-07-27
Migrate lib/pq to jackc/pgx/v5

pan93412:
- Tests are updated to reflect the latest PostgreSQL (15) default value.
- Migrate `lib/pq` to `pgx/v5`'s compatibility layer (to prevent rewriting too many things)
- Update Go lockfiles to drop out `lib/pq`'s dependencies (only the indirect one)


pan93412:
> Error: ../../../go/pkg/mod/github.com/jackc/pgx/v5@v5.6.0/pgconn/pgconn.go:159:57: undefined: errors.Join

Seems `pgx/v5` has been adopted to a newer Go version.

090809:
@stephenafamo can you check this?

pan93412:
> @stephenafamo can you check this?

I don't think it can be merged soon. This project needs to upgrade the Go version to use `pgx`, which is somewhat breaking.

My branch currently matches the `master` branch of sqlboiler. You can use `go get` as a workaround. 🥰

stephenafamo:
The driver is only used to read the database during generation and the user is free to use whatever `*sql.DB` they want at runtime.

Since there is no benefit to upgrading, we will just stick with `lib/pq` unless something else forces us to upgrade the Go version.

PR #1414, opened 2024-09-04
wip: added --no-schema option scaffolding, needs tests

morganhein:
This adds the `--no-schema` option, which allows for overriding the default option in the driver.

This flag only allows forcefully ignoring the schema, setting `--no-schema=false` does not force usage of the schema.

This code works, but I'd like some feedback on what else needs to be updated to make this land.

I don't feel confident deciding which tests to update for this, nor if you'd like README stuffs to be updated.

Any guidance is welcome, and i'll do what I can.

Thanks!

### DIFF
--- a/boilingcore/templates.go
+++ b/boilingcore/templates.go
@@ -39,6 +39,7 @@ type templateData struct {
 	RQ string
 
 	// Control various generation features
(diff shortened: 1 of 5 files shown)


stephenafamo:
Kindly fix the new conflicts 🙏🏾

PR #1422, opened 2024-10-24
feat(template): transfer []byte type in load function

xux1217:
when primary and foreign key is []byte, transfer []byte to string for map key.

This PR also fixes the binary foreign key error in the mysql test schema.

### DIFF
--- a/boil/load_map_key.go
+++ b/boil/load_map_key.go
@@ -1,0 +1,17 @@
+package boil
+
+import (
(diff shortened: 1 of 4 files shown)


stephenafamo:
For a tiny bit more efficiency, I'd prefer to detect the type during generation not at runtime. For example, take a look at how `isPrimitive` and `usesPrimitive` are used in the templates.

PR #1426, opened 2024-12-10
add TestBindJsonSlice

ianrose14:
Demonstrates what looks like a buffer-reuse (overwrite) bug when reading query results into `types.JSON` values.


ianrose14:
> An update to the `ubuntu-latest` GitHub runner makes the tests fail very early.
> 
> Kindly rebase again `master`

done!

Issue #1279, opened 2023-05-30
Unused import "database/sql" in view models


adatob:
Hey, I've noticed that the model generator added unused "database/sql" import for SQL VIEW models.

![image](https://github.com/volatiletech/sqlboiler/assets/115064733/8ee98045-a319-4b4a-a133-be461a6c23a8)

version of SQLBoiler: 4.14.2

database version: Postgresql 10

adatob:
The view was created as follow:
`CREATE OR REPLACE VIEW myschema.jets_pilots AS
  SELECT jets.id AS jet_id, jets.name as name, pilots.name AS pilot_name FROM myschema.jets AS jets
  LEFT JOIN myschema.pilots AS pilots ON pilots.id=jets.pilot_id;`

Tables `pilots` and `jets` was created within `myschema`.

Sqlboiler invoked using command:
`sqlboiler psql --debug --no-tests --add-enum-types --always-wrap-errors --pkgname myschema -c sqlboiler.toml -o sql/models/mydb/myschema`

PR #1438, opened 2025-02-02
Add imports processing when formating source files to remove unused imports

eklatzer:
Fixes https://github.com/volatiletech/sqlboiler/issues/1279


stephenafamo:
I'm unsure how this fixes the issue

parnic:
> I'm unsure how this fixes the issue

The report is that an unused import can be generated in the file, leading to Go compile errors. This PR appears to post-process the file after it has been generated and run it through Go's own imports processor which will, I assume, remove the unused import.

https://pkg.go.dev/golang.org/x/tools/imports#Process

eklatzer:
@parnic You are correct, thanks for clarifying.
@stephenafamo Any further questions regarding this PR?

PR #1439, opened 2025-02-03
Add relation getters on base model structs

parnic:
This prevents having to check the pointer before accessing a relation.

It was already safe to access a nil R's Get() method, but this removes some of the extra typing and nil checks as a convenience.

Previously:
```go
var office *models.Office
if employee != nil {
    office = employee.R.GetOffice()
}
```

With this helper:
```go
office := employee.GetOffice()
```

It is much more useful, however, when accessing nested relationships. Instead of

```go
var c *models.RelC
if m != nil && m.R.GetA() != nil && m.R.GetA().R.GetB() != nil {
    c = m.R.GetA().R.GetB().R.GetC()
}
```

we can do

```go
c := m.GetA().GetB().GetC()
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -415,6 +415,7 @@ not to pass them through the command line or environment variables:
 | no-auto-timestamps        | false    |
 | no-rows-affected          | false    |
 | no-driver-templates       | false    |
+| no-relation-getters       | false    |
 | tag-ignore                | []       |
 | strict-verify-mod-version | false    |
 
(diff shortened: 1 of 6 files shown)


stephenafamo:
There is a slight possiblity this will conflict with a table name.

Can you kindly add a way to disable this feature using a flag?

parnic:
> There is a slight possiblity this will conflict with a table name.
> 
> Can you kindly add a way to disable this feature using a flag?

Done.

PR #1450, opened 2025-05-19
Revert "Add imports processing when formating source files to remove unused imports"

stephenafamo:
Reverts volatiletech/sqlboiler#1438


robinlindhout:
@stephenafamo would be great if this get's a release :)

Issue #1436, opened 2025-01-20
Incorrect relationships created with conditional unique constraint


hamstah:
Using v4.18.0 (previously tested on 4.16 too)
Using the postgres driver (but not sure it's postgres specific)

given the schema generated by

```sql
CREATE TABLE A (
    id serial PRIMARY KEY,
    content text NOT NULL
);

CREATE TABLE b (
    id serial PRIMARY KEY,
    a_id INT NOT NULL REFERENCES A(id) ON
    DELETE
        CASCADE,
        content text NOT NULL
);

CREATE UNIQUE INDEX unique_a_id ON b(a_id)
WHERE
    content = 'test';
```

The generated relationship on A is

```golang
...
// aR is where relationships are stored.
type aR struct {
	B *B `boil:"B" json:"B" toml:"B" yaml:"B"`
}
...
```
ie it created a 1:1 relationship, but it should be a 1:many relationship as there could be multiple B linked to the same A, as long as at most one of them has the `test` content.

if I remove the unique constraint I get the correct relationship

```golang
...
// aR is where relationships are stored.
type aR struct {
	BS BSlice `boil:"BS" json:"BS" toml:"BS" yaml:"BS"`
}
...
```

It seems like the issue if with the relationship logic trying to handle the unique constraint and convert the relationship from 1:many to 1:1 but not accounting for the conditional part of it.

PR #1462, opened 2025-08-20
Do not treat columns with partial unique indexes as unique in psql driver

adsa95:
This PR fixes a bug in the psql driver where partial unique indexes will mark a column as unique, which is not true as it is only unique over a subset of rows. This currently leads to bugs such as #1436 where incorrect relations are generated (one-to-one instead of one-to-many) but could potentially also lead to other bugs where this difference is important.

This fix is implemented by adding a condition to the query in `*PostgresDriver.loadUniqueColumns` that excludes partial indexes by filtering on the `pg_index.indpred` column which will be null if it is not a partial index as stated in the [PostgreSQL docs:](https://www.postgresql.org/docs/current/catalog-pg-index.html)
> indpred pg_node_tree
> Expression tree (in nodeToString() representation) for partial index predicate. Null if not a partial index.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic
 Versioning](http://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Fixed
+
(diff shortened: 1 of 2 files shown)


adsa95:
@stephenafamo done! Also reverted some accidental autoformatting that was not meant to be included.

Issue #1411, opened 2024-08-21
JSON camelCase config uses go like naming for words like ID


arjendevos:
Generated:
```
AccountID        string    `boil:"account_id" json:"accountID" toml:"account_id" yaml:"account_id"`
```

But it should be:
```
AccountID        string    `boil:"account_id" json:"accountId" toml:"account_id" yaml:"account_id"`
```

Config:
```
output = "models/dm"
pkgname = "dm"
wipe     = true
no-tests = true
add-enum-types = true
add-soft-deletes = true

[struct-tag-cases]
json = "camel"
```

Is this a bug or is it configurable?

stephenafamo:
Unfortunately, this is not configurable. Open to suggestions/PRs

KeitaShimura:
I gave this a try in #1470. Would love to hear your thoughts!

PR #1470, opened 2026-02-28
fix(boilingcore): use conventional camelCase for struct tag casing

KeitaShimura:
## Summary

When `json = "camel"` (or any tag format set to `camel`) is configured, SQLBoiler
was using `strmangle.CamelCase` which follows Go's initialism convention. This
caused column names like `account_id` to be rendered as `accountID` in struct
tags, whereas the conventional camelCase for JSON APIs is `accountId`.

This PR introduces `camelCaseNoInitialisms` in `boilingcore/templates.go` and
uses it in `generateTagWithCase` for `TagCaseCamel` instead of
`strmangle.CamelCase`.

## Changes

- **`boilingcore/templates.go`**: Add `camelCaseNoInitialisms` which converts
  snake_case to camelCase by title-casing only the first letter of each word,
  without uppercasing common Go initialisms (e.g. `ID`, `URL`). Update
  `generateTagWithCase` to use this function for `TagCaseCamel`.
- **`boilingcore/templates_test.go`**: Add `TestCamelCaseNoInitialisms` and
  `TestGenerateTagWithCaseCamel` to verify the new behaviour.

## Behaviour change

| Column name      | Before (broken) | After (fixed)  |
| ---------------- | --------------- | -------------- |
| `account_id`     | `accountID`     | `accountId`    |
| `http_url`       | `httpURL`       | `httpUrl`      |
| `column_name_id` | `columnNameID`  | `columnNameId` |

> **Note**: This is a breaking change for users who were relying on the previous
> (Go-initialism) output. However, the previous output did not conform to JSON
> camelCase conventions, so this is considered a bug fix.

## Related issue

Related to #1411

### DIFF
--- a/boilingcore/templates.go
+++ b/boilingcore/templates.go
@@ -13,8 +13,8 @@ import (
 	"text/template"
 
 	"github.com/Masterminds/sprig/v3"
-	"github.com/friendsofgo/errors"
 	"github.com/aarondl/strmangle"
+	"github.com/friendsofgo/errors"
 
 	"github.com/aarondl/sqlboiler/v4/drivers"
 )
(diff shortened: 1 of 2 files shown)


stephenafamo:
This will be a breaking change which is unacceptable

Issue #1409, opened 2024-08-09
Support relative database path for sqlite3 driver


muesli:
### What version of SQLBoiler are you using (`sqlboiler --version`)?
`SQLBoiler v4.16.2`

### What is your database and version (eg. Postgresql 10)
`sqlite3`

### If this happened at generation time what was the full SQLBoiler command you used to generate your models? (if not applicable leave blank)
`sqlboiler sqlite3`

### Further information. What did you do, what did you expect?

I realize the sqlite3 driver's README states to use absolute paths, but this is rather inconvenient to run on multiple systems / for different users. Being able to specify a relative path would be greatly appreciated.

Specifying a relative path to a database file will result in the generated tests failing (`no such table`). If that's the expected outcome and intended behavior, rejecting relative paths would improve the experience for first time users of sqlboiler.

muesli:
Related, but not exactly the same issue: #1407.

stephenafamo:
I'll be happy to review a PR

KeitaShimura:
I gave this a try in #1471. Would love to hear your thoughts!

fzlzjerry:
I can reproduce this on current master with `dbname = "database.sqlite"` and `output = "models"`: generation succeeds, but the generated test setup resolves the database from `models/`, clones an empty database, and every model test fails with `no such table`.

The earlier driver-only approach does not address that generated-test path. I’ll resolve relative SQLite database paths against the directory of `viper.ConfigFileUsed()` in the SQLite test template, preserve absolute paths, and add an end-to-end regression without restructuring the existing driver test.

PR #1471, opened 2026-02-28
fix(sqlite3): support relative path for dbname

KeitaShimura:
## Summary

- Users can now specify a relative path (e.g. `./mydb.sqlite`) for `dbname` in `sqlboiler.toml`
- In `Assemble`, the `dbname` is resolved to an absolute path via `filepath.Abs()` before building the connection string

## Problem

When a relative path was given as `dbname`, SQLite resolved it against the working directory at the time of execution. Since the working directory when running sqlboiler differs from the one used when running generated tests, the same relative path pointed to different locations, causing `no such table` errors.

## Changes

**`drivers/sqlboiler-sqlite3/driver/sqlite3.go`**

- Resolve `dbname` to an absolute path with `filepath.Abs()` if it is not already absolute

**`drivers/sqlboiler-sqlite3/driver/sqlite3_test.go`**

- Extract DB creation logic into a `createSQLiteDB` helper
- Add test cases for each path format:

| Test case                 | Path format              | Example                          | `filepath.IsAbs()` | Resolved by `filepath.Abs()` |
| ------------------------- | ------------------------ | -------------------------------- | :----------------: | :--------------------------: |
| `absolute_path`           | Absolute path            | `/var/folders/.../mydb.sql`      |   `true` (no-op)   |              -               |
| `relative_path`           | Relative path with `../` | `../../var/folders/.../mydb.sql` |      `false`       |             yes              |
| `relative_path_dot_slash` | Relative path with `./`  | `./mydb.sql`                     |      `false`       |             yes              |

## Related issue

Related to #1409

### DIFF
--- a/drivers/sqlboiler-sqlite3/driver/sqlite3_test.go
+++ b/drivers/sqlboiler-sqlite3/driver/sqlite3_test.go
@@ -12,14 +12,27 @@ import (
 	"testing"
 	"time"
 
-	"github.com/stretchr/testify/require"
 	"github.com/aarondl/sqlboiler/v4/drivers"
+	"github.com/stretchr/testify/require"
 	_ "modernc.org/sqlite"
 )
 
 var (
 	flagOverwriteGolden = flag.Bool("overwrite-golden", false, "Overwrite the golden file with the current execution results")
 )
+
+func createSQLiteDB(t *testing.T, path string, sql []byte) {
+	t.Helper()
+	out := &bytes.Buffer{}
+	cmd := exec.Command("sqlite3", path)
+	cmd.Stdout = out
+	cmd.Stderr = out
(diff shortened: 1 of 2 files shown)


drivers/sqlboiler-sqlite3/driver/sqlite3_test.go
stephenafamo, line 1: I don't understand why the tests needed to be modified.

stephenafamo (changes requested):

drivers/sqlboiler-sqlite3/driver/sqlite3_test.go
KeitaShimura: @stephenafamo 
I have no plans to make any further fixes. Given the time that has passed, please go ahead and close the PR if it’s no longer needed.

PR #1480, opened 2026-09-15
perf(queries): cache bind pointer fields per struct type

nohhyun:
bind() allocated every []*Struct element through makeStructPtr(), which walked each field of the struct type and parsed its boil tag looking for the "bind" pointer fields that need allocating. The struct type is fixed for the lifetime of a bind, so this recomputed the same answer on every row: a 1M-row query over a 20-column model ran 20M redundant tag parses.

The per-type mappingCache already caches column mappings, so the field scan moves there as bindPtrFields, built once in newMappingCache and consumed by the new newStructPtr method. It is fixed at construction time, so it is read without holding the cache mutex.

Binding 1000 rows, benchstat over 6 runs:

    BindSmallPtrSlice     312.3µ -> 176.5µ   -43.50%
    BindWidePtrSlice     1344.3µ -> 687.8µ   -48.83%

The []Struct paths and the allocation counts are unchanged, since the work removed was pure CPU.

Adds TestBind_InnerJoinPtrFields, which covers the allocation of ",bind" pointer fields. No existing test bound to a struct with one, despite it being the documented example on Bind.


nohhyun:
It is a minor performance improvement but I believe the improvement is monotonic. 
Please let me know. 

I also wrote a small benchmark test - which I could add if folks thinks that is better.

